PR #38, opened 2020-06-23
feature: ethers-core wasm32 arch compatibility

prestwich:
Adds wasm compatibility for the `ethers-core` crate.

## Motivation

Core functionality (abi encoding etc) should be usable in more environments (e.g a browser).

## Solution
1. Update the following constant in `crate::utils`:
    - `pub const WEI_IN_ETHER: U256 = U256([0x0, 0x0, 0x0, 0x0de0b6b3a7640000]);`
2. Remove `ganache` and `solc` utils when compiling for wasm32

building this way should work now: 
`$ cargo build --target wasm32-unknown-unknown`


prestwich:
Note that we're a long way off from wasm compatibility for any other part of the crate

Issue #36, opened 2020-06-22
Batched Transactions


gakonst:
**Is your feature request related to a problem? Please describe.**

Currently, submitting multiple transactions can be done either by:
1. joining multiple futures together: problem -> not atomic
2. using a smart contract which would have internal business logic for each transaction type: problem -> inflexible

It'd be nice if we could batch together a bunch of transactions atomically in a flexible way.

**Describe the solution you'd like**

We should provide a nice API to the [MultiCall](https://github.com/makerdao/multicall/blob/master/src/Multicall.sol) contract, similar to [Multicall.js](https://github.com/makerdao/multicall.js). Liquity has another [JS impl](https://github.com/liquity/liquity/blob/master/packages/providers/src/BatchedProvider.ts#L1-L221) for something similar.

PR #43, opened 2020-07-01
Implement Multicall functionality for batched calls

roynalnaruto:
## Motivation

This PR targets #36 

The motivation is to be able to read multiple values from the blockchain state (`call`s may be made to different smart contracts). This reduces the number of round trips and hence saves time.

## Solution

The solution has been inspired by the one prescribed in the #36 issue. We aggregate the calldata in a `Multicall` struct instance and then call Maker's Multicall contract, which is deployed on [multiple Ethereum networks](https://github.com/makerdao/multicall.js/blob/master/src/addresses.json).

Using the `Multicall` API would look something, like [implemented in the test](https://github.com/roynalnaruto/ethers-rs-1/blob/master/ethers-contract/tests/contract.rs#L180-L271):
```rust
// initiate the Multicall instance
let multicall = Multicall::new(client, Some(multicall_contract.address())).await?;
// OR
// In this case, ethers-rs will fetch the Multicall contract address from the client's network ID
// let multicall = Multicall::new(client, None).await?;
//
// OR
// let multicall = Multicall::new(client.clone(), None)
//     .await?
//     .add_call(first_call)
//     .add_call(second_call);

// add calls to multicall
let multicall = multicall.add_call(call_1);
let multicall = multicall.add_call(call_2);

// In solidity
// (u256 blockNumber, []bytes memory returnData)
let return_data: (String, (String, Address), Address, Address) =
    multicall.call().await?;
```

## Pending tasks
- [x] Document the `Multicall` module, as it is also exported
- [x] I was not really sure what `Err` type I should return. I have simply used the `ContractError::ConstructorError` type. This should of course be changed, but not sure if I should add another error type to the `ContractError` enums? Or have a separate enum for `Multicall`?
- [x] Make a small fix to the `Abigen` logic
At present, for a smart contract function with a single argument, the generated binding is:
```rust
pub fn aggregate(
    &self,
    calls: Vec<(Address, Vec<u8>)>,
) -> ContractCall<P, S, (U256, Vec<Vec<u8>>)> {
    self.0
        .method_hash([37, 45, 186, 66], (calls, )) // <--- note this
        .expect("method not found (this should never happen)")
}
```
This does make sense for multi-argument function, where a tuple is required. But in the case of a single argument, I think we should prefer something like this:
```rust
pub fn aggregate(
    &self,
    calls: Vec<(Address, Vec<u8>)>,
) -> ContractCall<P, S, (U256, Vec<Vec<u8>>)> {
    self.0
        .method_hash([37, 45, 186, 66], calls) // <--- and this
        .expect("method not found (this should never happen)")
}
```
Otherwise, the arguments supplied to the function call cannot be encoded and the `ethabi::Function::encode_input` throws an `InvalidData` error.

- [x] Handle edge-cases on the MAX number of calls allowed to be aggregated
`Ethers-rs` supports at the most 16 elements in a tuple for tokenization/detokenization. We must take note of that and either:
* Increase this limit in the `ethers-core/src/tokens.rs`
* Limit the number of calls in the `Multicall.calls` vector (this has to be done either way)

So that we don't risk a `Detokenization` error.

- [x] Re-evaluate `ethers-rs`'s tokenization/detokenization logic
This sub-task is mainly concerned with the traits `Tokenizable`, `TokenizableItem` and the way `Detokenize` is implemented for types that do implement `Tokenizable`. Please note the [pre-processing](https://github.com/roynalnaruto/ethers-rs-1/blob/master/ethers-contract/src/call.rs#L108-L113) and [post-processing](https://github.com/roynalnaruto/ethers-rs-1/blob/master/ethers-contract/src/multicall.rs#L187-L192) done to the tokens. This is done because:
* `ethabi` tokenizes a tuple return value as a vector of those individual tokens
So, for a function returning `(string, address)`, `ethabi` would tokenize it as:
```rust
let tokens: Vec<Token> = vec![Token::String(str_val), Token::Address(addr_val)]
```
But `ethers-rs` denotes them as a `Token::Tuple`, hence:
```rust
let tokens: Vec<Token> = vec![Token::Tuple(vec![Tuple::String(str_val), Tuple::Address(addr_val)])]
```
We can adapt to their tokenization approach (it already was like that before, when nested tuples wouldn't be possible to detokenize). But we will have to then either give up on the nested tokenization and also on tokenization of a vector of tuples. If not, we would have to customize `ethabi` to suit our needs, or continue with this `pre` and `post-processing` approach I have taken (To be fair, I think this wouldn't be THAT bad).

- [x] Make another small fix to the `Abigen` logic
At the moment, a contract function that is not `view` returns a `H256` for the transaction hash, as we expect a user to `send` a transaction to that function. But even a `call` could be made to such a function, and it could return some data. In order to address this, as suggested by @gakonst himself, we would need a struct similar to the [NameOrAddress](https://github.com/gakonst/ethers-rs/blob/master/ethers-core/src/types/chainstate/ens.rs), and calling it `HashOrData<T>`. This should be used as the `D: Detokenize` type for such a function, as it could return a `H256` on `ContractCall.send()` or `T` on `ContractCall.call()`.


ethers-contract/src/multicall.rs
```diff
@@ -0,0 +1,204 @@
...
+use ethers_signers::{Client, Signer};
+use hex::FromHex;
+use lazy_static::lazy_static;
+
+use std::{collections::HashMap, sync::Arc};
+
+use crate::call::{ContractCall, ContractError};
+use crate::multicall_contract::MulticallContract;
+
+lazy_static! {
```
gakonst, line 17: I prefer using `OnceCell` instead, syntax is fairly similar:https://docs.rs/once_cell/1.4.0/once_cell/sync/struct.Lazy.html

ethers-contract/src/multicall.rs
```diff
@@ -0,0 +1,204 @@
+extern crate hex;
```
gakonst, line 1:
```suggestion
```

No need for this since Rust 2018. Also, I have been using `rustc_hex` across the repo, so please use that instead of `hex`

ethers-contract/src/multicall.rs
gakonst, line 17: I prefer using `OnceCell` instead, syntax is fairly similar:https://docs.rs/once_cell/1.4.0/once_cell/sync/struct.Lazy.html

ethers-contract/src/multicall.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+#[derive(Clone)]
+struct Call {
+    target: Address,
+    data: Vec<u8>,
+    function: Function,
+}
+
+pub struct Multicall<P, S> {
```
gakonst, line 48: Please add some docs + minimal examples, to make it a bit clearer how to use this struct's methods

ethers-contract/src/multicall.rs
```diff
@@ -0,0 +1,204 @@
...
+    pub fn new(
+        address: Option<Address>,
+        block: Option<BlockNumber>,
+        client: impl Into<Arc<Client<P, S>>>,
+    ) -> Self {
+        Self {
+            address,
+            calls: vec![],
+            block,
+            client: client.into(),
```
gakonst, line 65: Instead of using a client, I think it'd be better if we stored a `MulticallContract`, that would also remove the need for instantiating it inside the `call` method

ethers-contract/src/multicall.rs
```diff
@@ -0,0 +1,204 @@
...
+impl<P, S> Multicall<P, S>
+where
+    P: JsonRpcClient,
+    S: Signer,
+{
+    pub async fn call<D: Detokenize>(&self) -> Result<(U256, D), ContractError> {
+        // 1. Fetch chain id and the corresponding address of Multicall contract
+        // preference is given to Multicall contract's address if provided
+        // otherwise check the address book for the client's chain ID
+        let address = match self.address {
```
gakonst, line 144: We can move this to the constructor, and make it `async`

ethers-contract/tests/contract.rs
```diff
@@ -172,6 +175,100 @@ mod eth_tests {
...
+
+        // initiate the Multicall instance
+        let multicall = Multicall::new(Some(multicall_contract.address()), None, client4.clone());
+
+        // add calls to multicall
+        let multicall = multicall.add_call(value);
+        let multicall = multicall.add_calls(vec![value2]);
+        let multicall = multicall.add_calls(vec![last_sender, last_sender2]);
+
+        let return_data: (U256, (String, (String, Address), Address, Address)) =
```
gakonst, line 262: I think we can drop the block_number return argument, it doesn't feel that useful, and I'm a bit confused by why they added it in the original multicall contract

ethers-contract/src/multicall.rs
```diff
@@ -0,0 +1,204 @@
...
+            })
+            .collect();
+
+        // 6. Form tokens that represent tuples
+        let tokens = vec![Token::Tuple(tokens)];
+
+        // 7. Detokenize from the tokens into the provided tuple D
+        let data = D::from_tokens(tokens)?;
+
+        Ok((block_number, data))
```
gakonst, line 202: I think we can drop the block_number return argument, it doesn't feel that useful, and I'm a bit confused by why they added it in the original multicall contract

ethers-contract/src/multicall.rs
```diff
@@ -0,0 +1,204 @@
...
+
+pub struct Multicall<P, S> {
+    address: Option<Address>,
+    calls: Vec<Call>,
+    block: Option<BlockNumber>,
+    client: Arc<Client<P, S>>,
+}
+
+impl<P, S> Multicall<P, S> {
+    pub fn new(
```
gakonst, line 56: I think instead we could simplify this a bit more and make `fn new(client: ...)`. We'd have `address` and `block` default to `None`, and add builder pattern methods which would set them separately, what do you think? I think that it's more likely that when people actually use the `MultiCall` contract, they'd just use it with the defaults

ethers-contract/src/multicall.rs
roynalnaruto: I have removed this. I think I also don't need the `rustc_hex` here, simply using `Address::from_str` should do the job!

ethers-contract/src/multicall.rs
roynalnaruto: I think this should be the better approach! Thanks. One doubt regarding this:
I would need to have a method `new` as well as `build` in order to support such functionality. Because if I need to instantiate the `Contract` instance for multicall, I would need the address. And if I'm expecting the address to be possibly supplied as an argument by the user, then I should have a `build` function that finally instantiates the contract. What do you think about this?

roynalnaruto: Or I can set the field as: `contract: Option<Contract<P, S>>`. So in that case, it will be `None` if the appropriate Network ID was not found. And we then expect the user to supply contract address. The `build` would then panic if no address was supplied and `contract` is also `None`

gakonst: > So in any case, I'd need to store client?

You can access a contract's `Client` by calling the `client()` method https://docs.rs/ethers-contract/0.1.3/ethers_contract/struct.Contract.html#method.client

roynalnaruto: Yes, but if we don't take the `address` field in the constructor `Multicall::new`, then in the case where the user would be adding it using the builder pattern later, we cannot construct an instance of the Multicall contract, and so we would need to store the client, right?

roynalnaruto:
Added another utility method `eth_balance_of` to append a call for fetching an address' ETH balance. Included in the docs and tests.

PR #47, opened 2020-07-03
Update README.md

aliatiia:
error when `cargo build` unless https prefix ("relative path without base") error


gakonst:
Oops - thank you!

PR #56, opened 2020-08-14
(feat) gas oracle support

### LINKED ISSUE #54: Gas Price API clients
It'd be nice if we could plug in one of the existing gas price APIs instead of relying on `eth_gasPrice`.

A port of [Maker's python implementation](https://github.com/makerdao/pygasprice-client/blob/a96d759cf1c6f51071def2fd262742f2d6cde37a/pygasprice_client/__init__.py) could be a good candidate.

roynalnaruto:
## Motivation

Resolves #54 

## Solution
A gas oracle is a generic struct that implements the `GasOracle` trait. The method `async fetch` in `GasOracle` allows fetching the gas price data from the underlying oracle as a `Result<U256, GasOracleError>` response.

#### API
```rust
use ethers::providers::{
   gas_oracle::{EthGasStation, Etherchain, Etherscan, GasOracle},
};

async fn foo() -> Result<(), Box<dyn std::error::Error>> {
    let gas_oracle_1 = EthGasStation::new(Some("my-api-key"));
  
    let gas_oracle_2 = EthGasStation::new(None);

    let gas_oracle_3 = Etherchain::new();
  
    let data_1 = gas_oracle_1.fetch().await?;
    let data_2 = gas_oracle_2.fetch().await?;
    let data_3 = gas_oracle_3.fetch().await?;
    Ok(())
}
```

#### Dependencies
* Needed for parsing between types (`String` to `u64`/`f32`) while deserialising
```
serde-aux = "0.6.1"
```

### DIFF
--- a/ethers-providers/src/gas_oracle/etherscan.rs
+++ b/ethers-providers/src/gas_oracle/etherscan.rs
@@ -1,17 +1,17 @@
 use async_trait::async_trait;
-use reqwest::{Client, Error as ReqwestError};
+use reqwest::Client;
 use serde::Deserialize;
 use serde_aux::prelude::*;
-use thiserror::Error;
 use url::Url;
 
-use crate::gas_oracle::{GasOracleError, GasOracleFetch, GasOracleResponse};
+use crate::gas_oracle::{GasOracle, GasOracleError, GasOracleResponse};
 
 const ETHERSCAN_URL_PREFIX: &str =
     "https://api.etherscan.io/api?module=gastracker&action=gasoracle";
 
 /// A client over HTTP for the [Etherscan](https://api.etherscan.io/api?module=gastracker&action=gasoracle) gas tracker API
-/// that implements the `GasOracleFetch` trait
+/// that implements the `GasOracle` trait
+#[derive(Debug)]
 pub struct Etherscan {
     client: Client,
     url: Url,
@@ -47,18 +47,6 @@ impl From<EtherscanResponse> for GasOracleResponse {
     }
 }
 
-#[derive(Error, Debug)]
-pub enum ClientError {
-    #[error(transparent)]
-    ReqwestError(#[from] ReqwestError),
-}
-
-impl From<ClientError> for GasOracleError {
-    fn from(src: ClientError) -> GasOracleError {
-        GasOracleError::HttpClientError(Box::new(src))
-    }
-}
-
 impl Etherscan {
     pub fn new(api_key: Option<&'static str>) -> Self {
         let url = match api_key {
@@ -76,10 +64,8 @@ impl Etherscan {
 }
 
 #[async_trait]
-impl GasOracleFetch for Etherscan {
-    type Error = ClientError;
-
-    async fn fetch(&self) -> Result<GasOracleResponse, ClientError> {
+impl GasOracle for Etherscan {
+    async fn fetch(&self) -> Result<GasOracleResponse, GasOracleError> {
         let res = self
             .client
             .get(self.url.as_ref())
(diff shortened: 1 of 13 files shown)


ethers-providers/src/gas_oracle/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use ethers_core::types::U256;
+
+use async_trait::async_trait;
+use reqwest::Error as ReqwestError;
+use thiserror::Error;
+
+/// Various gas price categories. Choose one of the available
+#[derive(Debug)]
+pub enum GasCategory {
```
shamatar, line 18: Usually such enums should derive a full set `#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]`. First few are for ease of use, but Hash may be required by end users (cache make a hash map of all the gas options from time to time and then query it)

ethers-providers/src/gas_oracle/etherscan.rs
```diff
@@ -60,6 +60,12 @@ impl Etherscan {
 #[async_trait]
 impl GasOracle for Etherscan {
     async fn fetch(&self) -> Result<U256, GasOracleError> {
+        match self.gas_category {
+            GasCategory::Fast => return Err(GasOracleError::GasCategoryNotSupported),
+            GasCategory::Fastest => return Err(GasOracleError::GasCategoryNotSupported),
+            _ => {}
+        };
```
gakonst, line 67: I believe this can be rewritten with `if matches!(self.category, GasCategory::Fast || GasCategory::Fast) { return Err(...) }`

https://doc.rust-lang.org/beta/std/macro.matches.html

ethers-providers/src/gas_oracle/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+/// let eth_gas_station_oracle = EthGasStation::new(Some("my-api-key"));
+/// let etherscan_oracle = EthGasStation::new(None).category(GasCategory::SafeLow);
+///
+/// let data_1 = eth_gas_station_oracle.fetch().await?;
+/// let data_2 = etherscan_oracle.fetch().await?;
+/// # Ok(())
+/// # }
+/// ```
+#[async_trait]
+pub trait GasOracle: std::fmt::Debug {
```
shamatar, line 58: Kind of similar comment here, you may want to be conservative to require something like
```
    : Send 
    + Sync 
    + 'static 
    + std::any::Any 
    + std::fmt::Debug
```

I'm not 100% sure how not specifying Send/Sync would interact with futures that take ownership on parameters that contain something like `Box<dyn GasOracle>` or similar.

ethers-providers/src/gas_oracle/etherscan.rs
shamatar: More or less personal codestyle preference. I'd say 2-3 specific matches and `_ => {}` are more readable than macro. But `GasCategory::Fast || GasCategory::Fast` should be used anyway in such cases.

shamatar:
If you would want to have `Clone` on trait objects then usually it's something like
```
trait Trait {
   fn box_clone(&self) -> Box<dyn Trait>
}

impl Trait for T {
    fn box_clone(&self) -> Box<dyn Trait> {
        Box::from(self.clone())
    }
}

impl Clone for Box<dyn Trait> {
    fn clone(&self) -> Self {
        self.box_clone()
    }
}
```

roynalnaruto:
> If you would want to have `Clone` on trait objects then usually it's something like
> 
> ```
> trait Trait {
>    fn box_clone(&self) -> Box<dyn Trait>
> }
> 
> impl Trait for T {
>     fn box_clone(&self) -> Box<dyn Trait> {
>         Box::from(self.clone())
>     }
> }
> 
> impl Clone for Box<dyn Trait> {
>     fn clone(&self) -> Self {
>         self.box_clone()
>     }
> }
> ```

Nice, thanks for that. For now, I modified the code to not require the `Clone` altogether. And in `ethers-rs` it is prescribed to use `Arc::Clone(&client)` so I think we won't need it to implement `Clone`.

shamatar:
You have taken my suggestions too literally :) E.g.you do not need Any if you do no plan to do upcasting/downcasting. It will not hurt, but please take such opinions with a grain of salt.

gakonst (approved):
Fantastic. Thank you so much.

PR #59, opened 2020-08-20
Nonce manager

gakonst:
## Motivation

It'd be nice if we had a way to sign multiple transactions locally and have the client auto-figure what the nonce for the transactions should be.

## Solution

Override the default remote nonce logic with a locally managed one, falling back to the on-chain call if there's a nonce mismatch

- Also improves json error messages from the HTTP provider (finally)

### DIFF
--- a/ethers-signers/src/client.rs
+++ b/ethers-signers/src/client.rs
@@ -1,4 +1,4 @@
-use crate::Signer;
+use crate::{NonceManager, Signer};
 
 use ethers_core::types::{
     Address, BlockNumber, Bytes, NameOrAddress, Signature, TransactionRequest, TxHash, U256,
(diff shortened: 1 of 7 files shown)


ethers-signers/src/client.rs
```diff
@@ -74,8 +74,35 @@ pub struct Client<P, S> {
...
     pub(crate) gas_oracle: Option<Box<dyn GasOracle>>,
+    pub(crate) nonce_manager: RwLock<NonceManager>,
 }
 
+use tokio::sync::RwLock;
+
+#[derive(Clone, Debug)]
+pub(crate) struct NonceManager {
+    initialized: bool,
+    nonce: U256,
```
prestwich, line 85: use `AtomicU64` internally and convert to a U256 for the pub api? Unless you're anticipating needing nonces above `u64::MAX`

shamatar: Nonce may be limited to 2^63 - 1 soon, so atomic is enough

PR #61, opened 2020-08-28
fix(core): serialize filters properly and always rewake

### LINKED ISSUE #51: eth_newFilter doesn't return any logs
**Version**

```
git+https://github.com/gakonst/ethers-rs#6f02bb54365f5a786db1214497b9d1ebf8304995
```

**Platform**
Linux XPS-15-9570 5.4.0-40-generic #44-Ubuntu SMP Tue Jun 23 00:01:04 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

**Description**
Using `eth_newFilter` doesn't seem to return any events. I've never used the ethers.js equivalent, so I suspect it's a problem on my end.

I tried this code:

https://github.com/Vypo/ethers-newfilter

With this contract:

```solidity
// SPDX-License-Identifier: Unlicense

pragma solidity ^0.6.8;

contract Banana {
    event EatDaBanana(uint256 indexed a0, uint256 indexed a1, uint256 indexed a2);
    
    constructor() public {
        emit EatDaBanana(1, 2, 3);
        selfdestruct(msg.sender);
    }
}
```

I expected to see this happen: Logs to be printed to the terminal.

Instead, nothing happened.

gakonst:
## Motivation

Fixes #51 

## Solution

Previously it'd serialize to `["0x123", null, null, null]` which is not spec-compliant behavior. Now we serialize according to [the spec](https://eth.wiki/json-rpc/API#eth_newfilter). In addition we always re-wake to avoid getting stuck in the reactor's multithreaded eloop

### DIFF
--- a/ethers-core/src/types/chainstate/log.rs
+++ b/ethers-core/src/types/chainstate/log.rs
@@ -3,7 +3,7 @@ use crate::{
     types::{Address, BlockNumber, Bytes, H256, U256, U64},
     utils::keccak256,
 };
-use serde::{Deserialize, Serialize, Serializer};
+use serde::{ser::SerializeStruct, Deserialize, Serialize, Serializer};
 use std::str::FromStr;
 
 /// A log produced by a transaction.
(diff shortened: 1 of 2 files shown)


ethers-core/src/types/chainstate/log.rs
```diff
@@ -125,32 +126,40 @@ impl Filter {
...
     pub fn limit(mut self, limit: usize) -> Self {
         self.limit = Some(limit);
         self
     }
+
+    fn set_or_push<T: Into<ValueOrArray<H256>>>(&mut self, index: usize, topic: T) {
+        if self.topics.len() == index {
+            self.topics.push(Some(topic.into()));
+        } else {
+            self.topics[index] = Some(topic.into());
```
prestwich, line 160: won't this fail if you `set_or_push` out of order?

e.g. 
```
let f = Filter::new();
f.set_or_push(3, some_stuff); // panics?
```

PR #62, opened 2020-09-04
demo

shamatar:
Demo of having monomorphized `Provider<T>` instead of e.g `Provider<Http>` or one for web socket.


gakonst:
Thanks, closing for now to keep the PRs clean. Will revisit later on to see if this is doable in a clean API, given the recent middleware refactor

PR #65, opened 2020-09-17
feature: Middleware Architecture

gakonst:
Introduces a new `Middleware` trait which is implemented by `Provider`. Anyone that writes a new middleware needs to implement it, like below:

```rust
struct MyStruct<M>(M);

#[derive(Error, Debug)]
pub enum MyError<M: Middleware> {
    #[error("{0}")]
    MiddlewareError(M::Error),

    // Add your middleware's specific errors here
}

#[async_trait(?Send)]
impl<M> Middleware for MyStruct<M>
where
    M: Middleware,
{
    type Error = MyError<M>;
    type Provider = M::Provider;
    type Inner = M;

    fn inner(&self) -> &M {
        &self.0
    }
 
   // Add any of the functions you want to override e.g. a logging middleware
}
```


ethers-middleware/tests/stack.rs
```diff
@@ -0,0 +1,41 @@
...
+    let provider = Provider::<Http>::try_from(ganache.endpoint()).unwrap();
+    let provider_clone = provider.clone();
+
+    // The gas price middleware MUST be below the signing middleware for things to work
+    let provider = GasOracleMiddleware::new(provider, gas_oracle);
+
+    // The signing middleware signs txs
+    let provider = Client::new(provider, signer).await.unwrap();
+
+    // The nonce manager middleware MUST be above the signign middleware so that it overrides
```
kobigurk, line 24:
```suggestion
    // The nonce manager middleware MUST be above the signing middleware so that it overrides
```

PR #67, opened 2020-09-20
[WIP] Replace libsecp k256

roynalnaruto:
## Motivation

This PR removes the dependency on `libsecp256k`, replacing it with the `k256` library.

## Pending

- [ ] Data and Tx signed in `ethers-core` should be verifiable in `web3.js` (recovered address should be the signer's address)
- [ ] Data and Tx signed in `web3.js` should be verifiable in `ethers-core`

### DIFF
--- a/ethers-core/src/types/crypto/keys.rs
+++ b/ethers-core/src/types/crypto/keys.rs
@@ -21,11 +21,10 @@ use k256::{
         SigningKey,
     },
     elliptic_curve::{error::Error as EllipticCurveError, FieldBytes},
-    EncodedPoint as K256PublicKey, SecretKey as K256SecretKey, Secp256k1,
+    EncodedPoint as K256PublicKey, Secp256k1, SecretKey as K256SecretKey,
 };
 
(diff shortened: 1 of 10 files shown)


ethers-core/src/types/crypto/keys.rs
```diff
@@ -286,20 +302,22 @@ impl<'de> Deserialize<'de> for PublicKey {
             where
                 S: SeqAccess<'de>,
             {
-                let mut bytes = [0u8; COMPRESSED_PUBLIC_KEY_SIZE];
+                let mut bytes = [0u8; FULL_PUBLIC_KEY_SIZE];
```
roynalnaruto, line 305: We don't need this, and can stick to a 33-byte representation. Because there is a way to decompress into the uncompressed format:
https://docs.rs/elliptic-curve/0.6.0/elliptic_curve/sec1/struct.EncodedPoint.html#method.decompress

gakonst:
superseded by https://github.com/gakonst/ethers-rs/pull/72

PR #70, opened 2020-09-24
Block on Ledger Calls

gakonst:
In order to re-add the Send bound on our async-traits, we need to have a `Send` signer. Ledger calls are not `Send`. We fix that by blocking on the future instead of waiting for its completion. Probably not a big deal


gakonst:
Closing and will re-open from master if my concerns around blocking are resolved

PR #72, opened 2020-09-25
feat: convert signing to k256

gakonst:
Supersedes https://github.com/gakonst/ethers-rs/pull/67

TODO: Fix ProxyDigest implementation to have byte equality with libsecp sigs

### DIFF
--- a/ethers-core/src/types/crypto/hash.rs
+++ b/ethers-core/src/types/crypto/hash.rs
@@ -1,6 +1,7 @@
 //! This is a helper module used to pass the pre-hashed message for signing to the
 //! `sign_digest` methods of K256.
 use crate::types::H256;
+use elliptic_curve::consts::U64;
 use k256::ecdsa::signature::digest::{
     generic_array::GenericArray, BlockInput, Digest, FixedOutput, Output, Reset, Update,
 };
@@ -59,12 +60,13 @@ impl<D: Digest> Reset for ProxyDigest<D> {
     }
 }
 
+// Use Sha256 with 512 bit blocks
 impl<D: Digest> BlockInput for ProxyDigest<D> {
-    type BlockSize = D::OutputSize;
+    type BlockSize = U64;
 }
 
 impl<D: Digest> FixedOutput for ProxyDigest<D> {
-    // we default to the output of the orignal digest
+    // we default to the output of the original digest
     type OutputSize = D::OutputSize;
 
     fn finalize_into(self, out: &mut GenericArray<u8, Self::OutputSize>) {
@@ -79,7 +81,7 @@ impl<D: Digest> FixedOutput for ProxyDigest<D> {
     }
 
     fn finalize_into_reset(&mut self, out: &mut GenericArray<u8, Self::OutputSize>) {
-        let s = core::mem::replace(self, Self::default());
+        let s = std::mem::take(self);
         s.finalize_into(out);
     }
 }
--- a/ethers-core/src/types/crypto/hash.rs
+++ b/ethers-core/src/types/crypto/hash.rs
@@ -1,0 +1,85 @@
+//! This is a helper module used to pass the pre-hashed message for signing to the
+//! `sign_digest` methods of K256.
+use crate::types::H256;
+use k256::ecdsa::signature::digest::{
+    generic_array::GenericArray, BlockInput, Digest, FixedOutput, Output, Reset, Update,
+};
+
+pub type Sha256Proxy = ProxyDigest<sha2::Sha256>;
+
+#[derive(Clone)]
+pub enum ProxyDigest<D: Digest> {
+    Proxy(Output<D>),
+    Digest(D),
+}
+
+impl<D: Digest + Clone> From<H256> for ProxyDigest<D>
+where
+    GenericArray<u8, <D as Digest>::OutputSize>: Copy,
+{
+    fn from(src: H256) -> Self {
+        ProxyDigest::Proxy(*GenericArray::from_slice(src.as_bytes()))
+    }
+}
+
+impl<D: Digest> Default for ProxyDigest<D> {
+    fn default() -> Self {
+        ProxyDigest::Digest(D::new())
+    }
+}
+
+impl<D: Digest> Update for ProxyDigest<D> {
+    // we update only if we are digest
+    fn update(&mut self, data: impl AsRef<[u8]>) {
+        match self {
+            ProxyDigest::Digest(ref mut d) => {
+                d.update(data);
+            }
+            ProxyDigest::Proxy(..) => {
+                unreachable!("can not update if we are proxy");
+            }
+        }
+    }
+
+    // we chain only if we are digest
+    fn chain(self, data: impl AsRef<[u8]>) -> Self {
+        match self {
+            ProxyDigest::Digest(d) => ProxyDigest::Digest(d.chain(data)),
+            ProxyDigest::Proxy(..) => {
+                unreachable!("can not update if we are proxy");
+            }
+        }
+    }
+}
+
+impl<D: Digest> Reset for ProxyDigest<D> {
+    // make new one
+    fn reset(&mut self) {
+        *self = Self::default();
+    }
+}
+
+impl<D: Digest> BlockInput for ProxyDigest<D> {
+    type BlockSize = D::OutputSize;
+}
+
+impl<D: Digest> FixedOutput for ProxyDigest<D> {
+    // we default to the output of the orignal digest
+    type OutputSize = D::OutputSize;
+
+    fn finalize_into(self, out: &mut GenericArray<u8, Self::OutputSize>) {
+        match self {
+            ProxyDigest::Digest(d) => {
+                *out = d.finalize();
+            }
+            ProxyDigest::Proxy(p) => {
+                *out = p;
+            }
+        }
+    }
+
+    fn finalize_into_reset(&mut self, out: &mut GenericArray<u8, Self::OutputSize>) {
+        let s = core::mem::replace(self, Self::default());
+        s.finalize_into(out);
+    }
+}
--- a/ethers-core/src/types/crypto/signature.rs
+++ b/ethers-core/src/types/crypto/signature.rs
@@ -101,7 +101,8 @@ impl Signature {
         };
 
         let (recoverable_sig, _recovery_id) = self.as_signature()?;
-        let verify_key = recoverable_sig.recover_verify_key_from_digest_bytes(message_hash.as_ref().into())?;
+        let verify_key =
+            recoverable_sig.recover_verify_key_from_digest_bytes(message_hash.as_ref().into())?;
 
         let uncompressed_pub_key = K256PublicKey::from(&verify_key).decompress();
         if uncompressed_pub_key.is_some().into() {
@@ -284,6 +285,10 @@ mod tests {
 
         // recover the address from the above signature
         let recovered = signature.recover("Some data").unwrap();
+        assert_eq!(
+            format!("{:?}", recovered),
+            "0x2c7536e3605d9c16a7a3d7b1898e529396a65c23"
+        );
 
         assert_eq!(recovered, address);
     }
--- a/ethers-core/src/types/crypto/signature.rs
+++ b/ethers-core/src/types/crypto/signature.rs
@@ -101,7 +101,7 @@ impl Signature {
         };
 
         let (recoverable_sig, _recovery_id) = self.as_signature()?;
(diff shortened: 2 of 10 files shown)


ethers-core/src/types/crypto/signature.rs
```diff
@@ -92,22 +99,27 @@ impl Signature {
...
     fn as_signature(&self) -> Result<(RecoverableSignature, RecoveryId), SignatureError> {
         let recovery_id = self.recovery_id()?;
         let signature = {
-            let mut sig = [0u8; 64];
-            sig[..32].copy_from_slice(self.r.as_bytes());
-            sig[32..].copy_from_slice(self.s.as_bytes());
-            RecoverableSignature::parse(&sig)
+            let gar: &GenericArray<u8, U32> = GenericArray::from_slice(self.r.as_bytes());
+            let gas: &GenericArray<u8, U32> = GenericArray::from_slice(self.s.as_bytes());
+            let sig = K256Signature::from_scalars(*gar, *gas)?;
```
tarcieri, line 121: FYI, I believe you can avoid the `GenericArray` crap here by doing:

```suggestion
            let sig = K256Signature::from_scalars(self.r.as_fixed_bytes().into(), self.r_as_fixed_bytes().into())?;
```

or if `into()` doesn't work, try `as_ref()`

ethers-signers/src/wallet.rs
```diff
@@ -77,7 +77,7 @@ impl Wallet {
     // TODO: Add support for mnemonic and encrypted JSON
 
     /// Creates a new random keypair seeded with the provided RNG
-    pub fn new<R: Rng>(rng: &mut R) -> Self {
+    pub fn new<R: Rng + CryptoRng>(rng: &mut R) -> Self {
```
tarcieri, line 80: The way we usually write these in RustCrypto is:

```suggestion
    pub fn new(mut rng: impl Rng + CryptoRng) -> Self {
```

...since at least RngCore has an impl of itself for its own reference types. I would assume `Rng` does too?

ethers-core/src/types/crypto/signature.rs
```diff
@@ -92,22 +99,27 @@ impl Signature {
             RecoveryMessage::Data(ref message) => hash_message(message),
             RecoveryMessage::Hash(hash) => hash,
         };
-        let message = Message::parse_slice(message_hash.as_bytes())?;
 
-        let (signature, recovery_id) = self.as_signature()?;
-        let public_key = Secp256k1::recover(&message, &signature, &recovery_id)?;
+        let (recoverable_sig, _recovery_id) = self.as_signature()?;
+        let verify_key = recoverable_sig.recover_verify_key(message_hash.as_bytes())?;
```
shamatar, line 104: You need to use `recover_verify_key_from_digest` for this, otherwise you will to hash(message_hash) internally

tarcieri: > Ideally there needs to be a method in k256 that makes scalar from unreduced bytes and reduces it if necessary.

Yes, something like `from_bytes_mod_order`. Ideally I'd like to see it added to the `ff` crate.

> Even while it sounds biased, it's safe for verification as no one stops a signed from being lucky and having a valid scalar that is > modulus.

The real risk is misusing ECDSA without a random oracle selecting `z`. If the attacker has their pick of `z`, they can solve for it from the public key alone e.g. given public key `Q`, and choice of nonzero integers `a` and `b`, compute `R = aG + bQ`, then `r = R.x`, `s = R.x/b`, and `z = r * a/b`.

shamatar: I don't think that making a type-level restriction to misuse is a solution as at the first place people using it should understand what they are doing, but the change above has nothing to do with generation of the ECDSA challenge scalar anyway.

tarcieri: > I don't think that making a type-level restriction to misuse is a solution as at the first place people using it should understand what they are doing

Personally I'm a big fan of using the type system to prevent misuse.

> ...the change above has nothing to do with generation of the ECDSA challenge scalar anyway.

For what it's worth, the underlying ECDSA implementation will let you generate the scalar however you wish:

https://docs.rs/ecdsa/0.8.1/ecdsa/hazmat/trait.SignPrimitive.html

...however I'm not sure why the ephemeral scalar selection is relevant here? The reason the `recover_verify_key_from_digest` takes a `Digest` instance is mostly futureproofing around potentially verifying the signature as part of the recovery process.

shamatar: Everything above is three completely different deals:
- k scalar (random nonce in EcDSA) must be just random and not reused. For this purpose determinism is good, but large field and good rng are also as good (no bias)
- Fiat-Shamir transformation challenge (I suppose it's
z in your notations) must be just well randomly generated based on the message and commitment to the nonce above and only here it makes sense to require it to be made from the output of some Digest
- scalar s in a signature and in recovery is public and I'm not 100% aware of the requirements in it's form in Ethereum. It was not restricted at the beginning, so user could place anything that fits into 32 bytes and later on (implementation specific) makes a valid scalar that passes a verification equation. Most likely the original secp256k1 did not care about the reduction there at all cause probability of having a scalar that has dual representation that fits in 32 bytes is negligible for this curve

The problem with a current signature of the recovery function in k256 is that it wants an explicit output of some hash to verify and recover the signature. This restriction is useless cause input here is controlled by the verifier and does not use any private values, so it's safe to feed any random set of bytes here to initialize a scalar

tarcieri: If there's a legitimate use case for accepting a raw digest output (as a byte array) as part of the public key recovery function, I can consider it.

In this specific example, it looks like the `RecoveryMessage::Data` variant can be addressed by making `hash_message` return a digest, however `RecoveryMessage::Hash` cannot unless that variant is modified to contain e.g. a `Keccak256` instance.

I'm not sure what contexts `RecoveryMessage::Hash` comes up in, but if it'd be really be helpful to have it take a byte array, that's doable: the verification function in the TODO I mentioned earlier can potentially directly call the underlying `VerifyPrimitive` which can take a scalar instantiated from a hypothetical `Scalar::from_bytes_mod_order` function.

shamatar: Particular question about ethers structure is up for a Georgios, but recovery function that takes digest only uses it for Scalar::from_digest that is as far as I understand is just from_bytes_reduced() underneath anyway, so it's as good to have a third function that is something like recover_from_message_digest_bytes and takes a GenericArray of a required size. 

If I'm wrong about internals of ::from_digest then it still applies, but instead another function can be used to initiate a scalar. 

I think it's good to look at how C implementation works and may be not even bother with potential reduction.

tarcieri: > I think it's good to look at how C implementation works and may be not even bother with potential reduction.

Instantiating a `Scalar` without a reduction is effectively `from_bytes_unchecked` and violates the invariants of the `Scalar` type in the `k256` crate (i.e. it is no longer a valid element of the scalar field because it overflows the order). A reduction places it back in-range.

shamatar: C library does it with reduction https://github.com/bitcoin-core/secp256k1/blob/d7838ba6a6ac77cec173080f20efcd0e311ebfaa/src/secp256k1.c#L430

tarcieri: Here's a PR to add a `recoverable::Signature::recover_verify_key_from_digest_bytes` method to `k256`:

https://github.com/RustCrypto/elliptic-curves/pull/205

shamatar: The PR doesn't look to solve what was discussed:
- no reduction of scalar
- it requires some obscure FieldBytes that external caller may not want to know about. Even the name is obscure and exposes internal machinery. Anything from &[u8] here to [u8;32] or GenericArray looks better

tarcieri: > no reduction of scalar

The scalar reduction occurs here:

https://github.com/RustCrypto/elliptic-curves/pull/205/files#diff-7ce9d0f4abf3a5b2420393fe104c4f58R175

> it requires some obscure FieldBytes that external caller may not want to know about. Even the name is obscure and exposes internal machinery. Anything from &[u8] here to [u8;32] or GenericArray looks better

`FieldBytes` is an alias of `GenericArray<u8, U32>`. If you really prefer to write that way you can.

(That said, `FieldBytes` is one of the most common types in the whole `k256` API: it represents a serialized field element, in this case an element of the scalar field)

shamatar: Don't get me wrong, I appreciate your effort in making and maintaining this code, but what I argue here is that for an outsider all this looks strange.

What external user wants is to sign a message. He does not care about hazmat functions that can leak the private key to the attacker, but in general signing any message that does not map to scalar `0` and with no nonce reuse is safe. But what he observes is a signature of a signing function
```
fn try_sign_prehashed<K: Borrow<Scalar> + Invert<Output = Scalar>>(
        &self,
        ephemeral_scalar: &K,
        hashed_msg: &Scalar,
    ) -> Result<Signature, Error> {
```
that expects a scalar. All other signing function signatures largely require some entity from which we only use a fact that it outputs `[u8; 32]`. So why not give it to user? We know that for this particular curve input of 32 bytes is safe here, so something like
```
fn try_sign_prehashed_deterministic(
        &self,
        hashed_msg: &[u8; 32],
    ) -> Result<Signature, Error> {
```
is what people would want to use. Similar logic applies for recovery. Instead of some type that is `[u8; 32]` in the essence what's exposed (and without any other options) is `FieldBytes` that requires to follow a chain of docks to find that it's just a type alias.

tarcieri: This seems like a completely different topic from public key recovery. That said:

`try_sign_prehashed` is not intended for end users. That's why it's marked hazmat. The `Signer`/`Verifier` and `DigestSigner`/`DigestVerifier` traits are the end-user facing API.

This is documented as such:
https://docs.rs/ecdsa/0.8.1/ecdsa/hazmat/index.html#-warning-hazmat

> What external user wants is to sign a message. He does not care about hazmat functions that can leak the private key to the attacker, but in general signing any message that does not map to scalar 0 and with no nonce reuse is safe.

Most users don't know what a scalar is. Many won't know what a digest function is. They want to use a signing key to obtain a message.

`SignPrimitive` is the touch point between the generic ECDSA implementation in the `ecdsa` crate and low-level ECDSA implementations in the respective elliptic curve crates (`k256`, `p256`). These implementations can optimize ECDSA in various different ways depending on what arithmetic operations are available.

That said, this API is not intended for end users, and isn't really germane to the discussion here. If you have specific concerns about the internal structure of these crates, I'd invite you to open separate, specific issues and I can link you to the rather voluminous past discussions around these issues.

shamatar: Oh, thank you, tracking issue looks to kind of address the current problem, so the existing
```
    fn try_sign_digest(&self, digest: D) -> Result<Signature, Error> {
        ecdsa_core::SigningKey::from(self.secret_scalar).try_sign_digest(digest)
    }
```
can be directly transformed to serve my
```
fn try_sign_prehashed_deterministic(
        &self,
        hashed_msg: &[u8; 32],
    ) -> Result<Signature, Error> {
```
example.

shamatar:
@gakonst Georgios, you should be able to fix the PR and use it now

gakonst:
Signature verification works now with the new method added here https://github.com/RustCrypto/elliptic-curves/pull/205. We're only missing a method to also sign a prehashed message now, to replace the `SigningKey.sign` call, and then we should be good

ethers-core/src/types/crypto/hash.rs
```diff
@@ -0,0 +1,85 @@
+//! This is a helper module used to pass the pre-hashed message for signing to the
+//! `sign_digest` methods of K256.
```
tarcieri, line 2: It might be a bit more straightforward to use `RecoverableSignPrimitive:: try_sign_recoverable_prehashed` (which is impl'd on `k256::Scalar`) for this:

https://docs.rs/ecdsa/0.8.3/ecdsa/hazmat/trait.RecoverableSignPrimitive.html

The API to compute an ephemeral scalar with RFC6979 is public:

https://docs.rs/ecdsa/0.8.3/ecdsa/rfc6979/fn.generate_k.html

You can convert the raw message digest to a scalar using `Scalar::from_bytes_reduced`

gakonst:
@tarcieri wasn't sure how to do it exactly, ended up going with @shamatar's solution which does the job! Thank you everyone for your help, this was great.

PR #73, opened 2020-09-28
Digest proxy example

shamatar:
Proxy digest to sign a hash, but use another hashing algorithm for deterministic nonce generation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,9 +87,9 @@ dependencies = [
 
 [[package]]
 name = "async-io"
-version = "1.1.3"
+version = "1.1.4"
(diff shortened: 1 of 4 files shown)


shamatar:
For some reason 
```
#[test]
    fn recover_signature_from_message()
```
is broken (and it was in the original I suppose), but I didn't dig into it.

Also, I didn't check for any test vector for nonce generation compared to libsecp256k1 (C library), but from high level those should immediately match.

gakonst:
Fixed the test, refactored and cherry-picked on https://github.com/gakonst/ethers-rs/pull/72. Thanks a lot @shamatar!!

PR #74, opened 2020-09-30
fix Sha256 proxy

shamatar:
Properly set Sha256 block size to 512 bits.


gakonst:
Cherry-picked on https://github.com/gakonst/ethers-rs/pull/72/commits/8d7a165910e24a6816e21437789e8aa8847f9060 (+ clippy lint fixes) because I couldn't push to your branch. Thank you again!

PR #83, opened 2020-10-16
addr checksum encoding

roynalnaruto:
## Motivation

This PR adds a util method to convert an `Address` to a checksum encoded `String`

### DIFF
--- a/ethers-core/src/utils/mod.rs
+++ b/ethers-core/src/utils/mod.rs
@@ -19,6 +19,7 @@ pub use rlp;
 
 use crate::types::{Address, Bytes, U256};
 use k256::{ecdsa::SigningKey, EncodedPoint as K256PublicKey};
+use rustc_hex::ToHex;
 use std::convert::TryInto;
 
 /// 1 Ether = 1e18 Wei == 0x0de0b6b3a7640000 Wei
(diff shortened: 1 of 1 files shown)


ethers-core/src/utils/mod.rs
```diff
@@ -111,6 +112,30 @@ pub fn secret_key_to_address(secret_key: &SigningKey) -> Address {
...
+pub fn to_checksum(addr: &Address, chain_id: Option<u8>) -> String {
+    let prefixed_addr = match chain_id {
+        Some(chain_id) => format!("{}0x{:x}", chain_id, addr),
+        None => format!("{:x}", addr),
+    };
+    let hash = keccak256(&prefixed_addr.as_bytes()).to_hex::<String>();
+    let hash = hash.as_bytes();
+
+    let mut encoded: String = "0x".to_owned();
+    let addr_hex = format!("{:x}", addr);
```
kobigurk, line 126: Maybe use `encode_hex` so it's less fragile to string errors?

kobigurk (commented):
This is great! Thanks!

gakonst (approved):
Added some changes on top in https://github.com/gakonst/ethers-rs/pull/85

PR #88, opened 2020-10-24
refactor: extract minimal features of Contract into BaseContract

### LINKED ISSUE #80: Refactor ethers-contract
Ethers-contract currently requires that it's connected to a `Middleware` in order to be able to broadcast the transactions. 

It'd be nicer if we were able to have a base contract struct which simply takes the ABI and produces ABI encoded data for its functions (and for filters if given a vec of addresses as the filter's address). You'd then wrap that with an address and a middleware, and you'd get the equivalent functionality of the current Contract.

It'd also be nice if we supported human readable ABIs. We could do the tokenization by looking at a function's parentheses

pawanjay176:
## Motivation

Resolves #80 

## Solution

Refactors the `Contract` struct by extracting the `abi` into a `BaseContract` struct which does only abi encoding. 
Working on adding human-readable abis.

### DIFF
--- a/ethers-contract/src/base.rs
+++ b/ethers-contract/src/base.rs
@@ -1,0 +1,74 @@
+use crate::Contract;
+
+use ethers_core::{
+    abi::{Abi, FunctionExt},
+    types::{Address, Selector},
+};
+use ethers_providers::Middleware;
+
+use std::{collections::HashMap, fmt::Debug, hash::Hash, sync::Arc};
+
+/// A reduced form of `Contract` which just takes the `abi` and produces
+/// ABI encoded data for its functions.
+#[derive(Debug, Clone)]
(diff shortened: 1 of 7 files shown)


gakonst:
@pawanjay176 added some changes on top, also working on a human readable ABIs, will have something tomorrow

pawanjay176:
Thanks @gakonst. Also started working on the human readable abis. I'm doing plain parsing something similar to this https://gist.github.com/pawanjay176/4340020c41898cfee7d0430538cee986

The biggest issue with this approach is parsing the function/event arguments since it may contain solidity tuples. I was able to extend the above gist to handle all types of arguments, but the resulting code is incredibly ugly and doesn't handle many of the edge cases.

If you could share your approach, I can try and finish up the code for it :)

PR #91, opened 2020-10-29
fix: default param name if not found following memory/calldata

roynalnaruto:
## Motivation

Param parsing failed for a param marked as `memory` but without a param name:
```
bytes memory
```

## Solution

If no param name is found, return the default param name.


gakonst:
Thanks!

Issue #102, opened 2020-12-14
Middleware::pending_transaction hangs or panics


vakond:
**Version**
873cf09
```ethers = { git = "https://github.com/gakonst/ethers-rs", features = ["full"] }```

**Platform**
```5.4.0-58-generic #64-Ubuntu SMP Wed Dec 9 08:16:25 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux```

**Description**
Call of ```pending_transaction(tx_hash)``` after ```send_transaction(tx)``` hangs or panics.

I tried this code:
```
let tx_hash = client.send_transaction(tx, None).await?;
let receipt = client.pending_transaction(tx_hash).await?;
dbg!(receipt);
```
I expected to see the contents of the receipt.

Instead, the program hangs or panics with following message:
```thread 'tokio-runtime-worker' panicked at 'WS Server panic: NoResponse'```

I've prepared two mininal projects to reproduce this behaviour (please see corresponding README files):
https://github.com/vakond/ethers-ganache-pending
https://github.com/vakond/ethers-geth-pending
The bug (if it's a bug) can also be observed with the Kovan testnet.

It seems only one variant panics: Ganache+Ws. Ganache+Http just hangs. Other variants: {Geth, Kovan}+{Ws, Http} all hang.

vakond:
I still don't know if it's a bug, but there is a workaround.

My colleaque pointed out that one should refer to the exact version of tokio in project that uses your crate:
```tokio = { version = "0.2.21", features = ["macros"] }```
instead of, say
```tokio = { version = "~0.2", features = ["full"] }```

After that change all works like a charm. Thank you for your very good library.

gakonst:
@vakond thank you for the detailed bug report. Wish everyone did it this way. I opened https://github.com/gakonst/ethers-rs/pull/103 as a fix. I've verified it fixes the issue in ganache, could you confirm if it fixes it with Geth/Kovan?

vakond:
@gakonst thank you for your excellent library. I gladly confirm the Ganache test has passed. But unfortunately the Geth test still hangs, as well as Kovan. In case you want to try Kovan first, I'm attaching the main.rs template file here (Cargo.toml the same as in ganache-pending).
[kovan-main-rs.zip](https://github.com/gakonst/ethers-rs/files/5702505/kovan-main-rs.zip)

gakonst:
[...]

PR #104, opened 2020-12-17
fix: reschedule waker even if receipt is not available

gakonst:
Fixes https://github.com/gakonst/ethers-rs/issues/102

PR #103 did not address the Future hanging when the receipt was not
immediately available, e.g. in non dev environments


gakonst:
Merging - this is a flaky Celo tests which I haven't pinned down yet.

Issue #105, opened 2020-12-17
Flaky Celo Test


gakonst:
**Version**
`5b7578296b9f695300a5ef7f931e3aa2883d68fd`

**Description**

```
running 1 test
test celo_tests::deploy_and_call_contract ... FAILED

failures:

---- celo_tests::deploy_and_call_contract stdout ----
thread 'celo_tests::deploy_and_call_contract' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', ethers-contract/tests/contract.rs:395:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

[This test is flaky](https://github.com/gakonst/ethers-rs/blob/master/ethers-contract/tests/contract.rs#L394-L395). Sometimes it succeeds, sometimes it fails. The transaction gets broadcast properly, but it runs out of gas for some reason.

Example failing tx: https://alfajores-blockscout.celo-testnet.org/tx/0x33192056516baec27694f01607586e7043ef90c07dab255d2a379ab897dcd117/internal_transactions. 

Example success: https://alfajores-blockscout.celo-testnet.org/tx/0x4cdf69696ab2f62877dbd04b1a3056f89fe315c1881c38c5de33e866914c56f5/internal_transactions

PR #107, opened 2020-12-17
Returning a `PendingTransaction` after sending a tx

gakonst:
Improves the ergonomics of sending & waiting for a transaction to be mined. You can now do

```rust

// Before
let tx_hash = provider.send_transaction(...).await?
dbg!(tx_hash);
let receipt = provider.pending_transaction(tx_hash).await?;

// After:
let receipt = provider.send_transaction(...).await?.await?;

// or:
let pending_tx = provider.send_transaction(...).await?;
dbg!(*pending_tx);
let receipt = pending_tx.await?;

// you can still use the syntax for N confs like so:
let receipt = provider.send_transaction(...).await?.confirmations(6).await?;
```

# THIS IS A BREAKING CHANGE

* If you use previously used the tx_hash in the return values of a send_transaction call, you should deref it since `PendingTransaction` derefs to `TxHash`
* Smart contract calls need to be split in the "create call" and "send call":

```rust
// Before:
let pending_tx = contract
            .method::<_, H256>("setValue", "hi".to_owned())
            .unwrap()
            .send()
            .await
            .unwrap();

// After:
let call = contract.method::<_, H256>("setValue", "hi".to_owned()).unwrap();
let pending_tx = call.send().await.unwrap();

### DIFF
--- a/ethers-contract/src/call.rs
+++ b/ethers-contract/src/call.rs
@@ -1,9 +1,9 @@
 use super::base::{decode_fn, AbiError};
 use ethers_core::{
     abi::{Detokenize, Function, InvalidOutputType},
-    types::{Address, BlockNumber, Bytes, TransactionRequest, TxHash, U256},
+    types::{Address, BlockNumber, Bytes, TransactionRequest, U256},
 };
-use ethers_providers::Middleware;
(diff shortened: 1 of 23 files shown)


gakonst:
Merging, CI failure is related to #105

PR #109, opened 2020-12-18
feat: derives &, Arc and Box for Middleware

gakonst:
As title, nice ergonomics upgrade, h/t @vorot93 for https://github.com/auto-impl-rs/auto_impl

### DIFF
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -149,7 +149,7 @@ pub trait FromErr<T> {
 }
 
 #[async_trait]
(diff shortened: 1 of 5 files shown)


ethers-providers/src/lib.rs
```diff
@@ -147,6 +149,7 @@ pub trait FromErr<T> {
 }
 
 #[async_trait]
+#[auto_impl(&,Box, Arc)]
```
gakonst, line 152:
```suggestion
#[auto_impl(&, Box, Arc)]
```

PR #163, opened 2021-01-15
Handle ping/pong messages for Ws provider

SamWilsn:
## Motivation

Fixes disconnects with WebSocket subscriptions when Geth sends a `Ping` message.

## Solution

Reply with a `Pong` message, and ignore `Pong` messages since they can be spuriously generated.


prestwich:
LGTM

gakonst (approved):
Great catch, thank you

PR #165, opened 2021-01-16
feat: Transformer middleware with DsProxy impl

roynalnaruto:
## Motivation

This PR adds support for proxy wallets, eg. [DsProxy](https://github.com/dapphub/ds-proxy)

## Solution

* The `Transformer` trait must be implemented by a wallet like `DsProxy`
* The `TransformerMiddleware` can intercept and transform tx requests to send them as the appropriate txs that the proxy contract can execute.

### Usage
* Using DsProxy as a Transformer
```rust
// instantiate wallet and provider
// instantiate signer middleware
let middleware = SignerMiddleware::new(provider, wallet);
let middleware = Arc::new(middleware);

// instantiate DsProxy as the transformer
let ds_proxy = DsProxy::build::<HttpWallet, Arc<HttpWallet>>(
    Arc::clone(&middleware),
    Some(ds_proxy_factory.address()),
    provider.address(),
).await?;
let transformer_middleware = TransformerMiddleware::new(Arc::clone(&middleware), ds_proxy);

// send txs via DsProxy
transformer_middleware.send_transaction(tx_request, None).await?.await?;
```
* Using DsProxy standalone
```rust
// instantiate DsProxy as the transformer
let ds_proxy = DsProxy::build::<HttpWallet, Arc<HttpWallet>>(
    Arc::clone(&middleware),
    Some(ds_proxy_factory.address()),
    provider.address(),
).await?;

// execute(bytes memory _code, bytes memory _data)
ds_proxy
    .execute::<HttpWallet, Arc<HttpWallet>>(
        Arc::clone(&middleware),
        AddressOrBytes::Bytes(bytecode),
        calldata,
    )
    .await;

// execute(address _target, bytes memory _data)
ds_proxy
    .execute::<HttpWallet, Arc<HttpWallet>>(
        Arc::clone(&middleware),
        AddressOrBytes::Address(target),
        calldata,
    )
    .await;
```


ethers-middleware/src/proxy_wallet/ds_proxy/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+            target = addr;
+        } else {
+            return Err(ProxyWalletError::Dummy);
+        }
+
+        // fetch the data field.
+        let data = tx.data.unwrap_or(vec![].into());
+
+        // encode data as the ABI encoded data for DSProxy's execute method.
+        let ds_proxy_base: BaseContract = parse_abi(&[DS_PROXY_EXECUTE_SIGNATURE])?.into();
```
gakonst, line 41: Probably better for us to move this `BaseContract` in the DsProxy constructor, instead of re-parsing the ABI each time.

ethers-middleware/src/proxy_wallet/ds_proxy/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+    address: Address,
+}
+
+impl DsProxy {
+    /// Create a new instance of DsProxy by providing the address of the DsProxy contract that has
+    /// already been deployed to the Ethereum network.
+    pub fn new(address: Address) -> Self {
+        Self { address }
+    }
+}
```
gakonst, line 22: Should we also add an extra method which calls `function execute(bytes _code, bytes _data)`?

ethers-middleware/src/proxy_wallet/ds_proxy/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+mod tests {
+    use super::*;
+    use hex::FromHex;
+
+    #[test]
+    fn test_get_proxy_tx() {
+        let ds_proxy_addr = Address::random();
+        let ds_proxy = DsProxy::new(ds_proxy_addr);
+
+        let tx: TransactionRequest = serde_json::from_str(
```
gakonst, line 62: Where is this data from? A [transaction request](https://github.com/gakonst/ethers-rs/blob/master/ethers-core/src/types/transaction.rs#L22) does not have an`transaction_index` or a `chain_id`

A better way to structure this test could be:

1. create a `BaseContract` with a `set(uint256 x)` function
2. create the calldata for the call by encoding the function data for a sample call
3. call `get_proxy_tx` with the calldata from 2 and a dummy address to get the DsProxy's calldata
4. decode the data from 3 and assert that it matches the calldata from 2 and the dummy address

ethers-middleware/src/proxy_wallet/gnosis_safe/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+    fn get_proxy_tx(&self, tx: TransactionRequest) -> Result<TransactionRequest, ProxyWalletError> {
+        let mut proxy_tx = tx.clone();
+
+        // TODO: update the `data` field of the transaction to be the ABI encoded data for
+        // GnosisSafe's method.
+        let _gnosis_safe_base: BaseContract = parse_abi(&["function execTransaction( \
+                address to, \
+                uint256 value, \
+                bytes calldata data, \
+                Enum.Operation operation, \
```
gakonst, line 28:
```suggestion
                Enum.Operation operation, \
```

You might need to change this to a `uint8`

ethers-middleware/src/proxy_wallet/gnosis_safe/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+}
+
+impl ProxyWallet for GnosisSafe {
+    fn get_proxy_tx(&self, tx: TransactionRequest) -> Result<TransactionRequest, ProxyWalletError> {
+        let mut proxy_tx = tx.clone();
+
+        // TODO: update the `data` field of the transaction to be the ABI encoded data for
+        // GnosisSafe's method.
+        let _gnosis_safe_base: BaseContract = parse_abi(&["function execTransaction( \
```
gakonst, line 24: It'd be best if we moved this in the constructor

ethers-middleware/src/proxy_wallet/ds_proxy/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+
+const DS_PROXY_EXECUTE_SIGNATURE: &str =
+    "function execute(address target, bytes data) public payable returns (bytes memory response)";
+
+#[derive(Debug)]
+pub struct DsProxy {
+    address: Address,
+}
+
+impl DsProxy {
```
gakonst, line 16: Could be cool to add a few other convenience methods, e.g. a `create` method which allows using the DsProxyRegistry to create a new proxy easily

ethers-middleware/src/proxy_wallet/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+    #[error(transparent)]
+    AbiParseError(#[from] ParseError),
+
+    #[error(transparent)]
+    AbiError(#[from] AbiError),
+}
+
+pub trait ProxyWallet: Send + Sync + std::fmt::Debug {
+    fn get_proxy_tx(&self, tx: TransactionRequest) -> Result<TransactionRequest, ProxyWalletError>;
+}
```
gakonst, line 23: I wonder if there's some better naming here, maybe `TransformerMiddleware` and `fn transform`?

gakonst (commented):
So far so good! Left some comments. Thank you for the contribution

PR #182, opened 2021-01-27
Feat/is middleware signer

roynalnaruto:
## Motivation

It would be useful to have a way to know if a middleware can sign data.

A way to know if one of the two is true:
* The middleware stack contains `SignerMiddleware`
* The JSON-RPC provider has an unlocked key and it supports the `eth_sign` call

## Solution

* Add `is_signer` method to the `Middleware` trait
* `is_signer` is always true for `SignerMiddleware`
* `is_signer` is true if the JSON-RPC provider has an unlocked key and supports `eth_sign` call

### DIFF
--- a/ethers-providers/src/provider.rs
+++ b/ethers-providers/src/provider.rs
@@ -333,7 +333,20 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
     ) -> Result<Signature, ProviderError> {
         let data = utils::serialize(&data.into());
         let from = utils::serialize(from);
-        self.request("eth_sign", [from, data]).await
+
+        // get the response from `eth_sign` call and trim the 0x-prefix if present.
+        let sig: String = self.request("eth_sign", [from, data]).await?;
+        let sig = sig.strip_prefix("0x").unwrap_or(&sig);
+
+        // deserialize the signature from its {r, s, v} components.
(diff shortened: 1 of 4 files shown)


gakonst (approved):

ethers-providers/src/provider.rs
```diff
@@ -324,7 +333,20 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+
+        // deserialize the signature from its {r, s, v} components.
+        let v = sig[128..130].parse::<u64>().unwrap();
+        let sig_json = format!(
+            r#"{{"r": "0x{}", "s": "0x{}", "v": {}}}"#,
+            &sig[..64],
+            &sig[64..128],
+            v
+        );
+        Ok(serde_json::from_str(&sig_json).map_err(ProviderError::SerdeJson)?)
```
gakonst, line 349: Instead of this, should we instead use the [Signature TryFrom](https://github.com/gakonst/ethers-rs/blob/fc093362fbf77273ea6d1dcfd8d6e026bb6f6377/ethers-core/src/types/signature.rs#L153)?

```rust
let sig = hex::decode(sig)?
Ok(Signature::try_from(sig)?)
```

roynalnaruto: Agreed 👍

PR #200, opened 2021-02-19
make generated contract abi code reproducible

lerencao:
## Motivation

Each time I run abigen, the generated rust code changes for a same contract abi json.
I'd like to make it reproducible.

## Solution

Use BtreeMap to hold methods and events definition.


gakonst:
Excellent idea!

PR #225, opened 2021-03-11
Refactor human readable abi parsing

mattsse:
This includes a more sophisticated human readable abi parser that should be on par with `ethers-js`'s interface parser.

## Motivation
The existing parser followed the grammar loosely; 
False positives were possible, so that for example `event SetValue_function_called()` would be detected as `Function`.

## Solution
All parser function are replaced with a handwritten parser that follows `ethers-js` abi parsing rules.
This also detects `StateMutability` and constructors.
All tests are successfully updated with the new parsers.

## Up for discussion

* **mandatory `returns` statement**: The new parser will fail if a function lacks a `returns` statement before its outputs. The `contract_human_readable` example did not succeed because a missing `returns` statement (`function getValue() external view (string)`) was detected and fixed in #
f3bd156 . I'm not sure wether the preferential behavior would be to allow a missing `returns` or require it. `ethers-js` treats a missing `returns` before ouputs as an `INVALID_ARGUMENT` exception.
* **error handling**: This introduces a lot more error checking. Right now there is not a lot feedback about what was the root cause of an error, most of the time a simple `ParseError::ParseError(super::Error::InvalidData)` is returned. I'm open to make this responsive by introducing a `Message(String)` error type or provide the `ParseError::ParseError(super::Error::Other(anyow::Error))` with a message.
* **Side effects of constructor parsing**: `constructor`s are handled as `Constructor` now and asigned to `Contract::constructor`, I'm not sure about potential side effects of the constructor now missing from `Contract::functions`

### DIFF
--- a/ethers-core/src/abi/human_readable.rs
+++ b/ethers-core/src/abi/human_readable.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use thiserror::Error;
 
 use super::{
-    Abi, Event, EventParam, Function, Param, param_type::Reader, ParamType, StateMutability,
+    Abi, Event, EventParam, Function, Param, param_type::Reader, ParamType, StateMutability, Constructor
 };
 
 /// Parses a "human readable abi" string vector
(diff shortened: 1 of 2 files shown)


gakonst (approved):
Amazing, thank you.

PR #226, opened 2021-03-13
feat: add struct parsing support for human readable ABI

mattsse:
This adds support for parsing solidity structs in the humand readable ABI.
A new `AbiParser` can parse structs, turn them into tuples and substitute their references in events and functions.
Also this improves the error handling with textual feedback about the error that occurred.
Additional tests are tests are added as well.

## Motivation

Support for structs in the human readable ABI makes it easier to copy definitions from your solidity contract without further modifications.

## Solution

The `SolStruct` can parse solidity structs, supported features are:

* elementary datatypes (`ethabi::ParamType`), like `int25 _y;`
* struct references as `String`, like `MyStruct m;` and `MostOuter.Outer.Inner m;`
* struct arrays, fixed and unbounded: `MyStruct[] m;` and `MyStruct[10] m;`
* mappings and nested mappings like `mapping(string => string)` and  `mapping(string => mapping(string => string))`

A new `AbiParser` is introduced that handles struct parsing and replacing them as tuples, either within structs or in event and function arguments.

Substitution is based on the following rules:

* a struct becomes a tuple: `struct S {int x; int y;}` => `tuple(int,int)`
* a struct array becomes an array: `function(S[] _s)` => `tuple(int,int)[]` and `function(S[10] _s)` => `tuple(int,int)[10]`
* nested structs become nested tuples: `struct X {S s;  address a;}` => `tuple(tuple(int,int),address)
* mappings are disallowed and lead to errors in the `AbiParser`

The `parse` api remaines unchanged but its code is replaced `AbiParser::default().parse(s)`

### DIFF
--- a/ethers-core/src/abi/error.rs
+++ b/ethers-core/src/abi/error.rs
@@ -1,0 +1,25 @@
+//! Boilerplate error definitions.
+use thiserror::Error;
+
+/// A type alias for std's Result with the Error as our error type.
+pub type Result<T, E = ParseError> = std::result::Result<T, E>;
+
+#[derive(Error, Debug)]
+pub enum ParseError {
+    #[error("{0}")]
+    Message(String),
+    #[error(transparent)]
+    ParseError(#[from] super::Error),
+}
+
+macro_rules! _format_err {
+    ($($tt:tt)*) => {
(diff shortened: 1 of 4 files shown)


gakonst (approved):
This is great, thank you.

PR #229, opened 2021-03-15
fix: make EthEvent abi attribute work for tuple inputs

mattsse:
## Motivation
In #227 I mentioned a limitation in parser when the provided attribute contains tuples: (`#[ethevent(abi = "Event((string,address),string)")]` which resulted in a failed derivation of the `EthEvent` trait.

## Solution
This has been fixed by adding an attempt to parse the `abi` input as tuple and validated by a new test


gakonst (approved):
LGTM, awesome.

Issue #230, opened 2021-03-16
Add support for EIP-234 and EIP-1898


GCdePaula:
[EIP-234](https://eips.ethereum.org/EIPS/eip-234): Add `blockHash` to JSON-RPC filter options.
[EIP-1898](https://eips.ethereum.org/EIPS/eip-1898): Add `blockHash` to JSON-RPC methods which accept a default block parameter.

These two EIPs are closely related, and allow for querying the blockchain's state at a specified block hash. I've a draft implementation of these two ready and tested :)

There are two issues I'd like some guidance on, though.

The first is that, the way I was approaching it,EIP-1898 requires changing the signatures of a reasonable amount of methods of the trait `etheres_providers::Middleware`. Specifically, changing the types of `Option<BlockNumber>` to `Option<BlockId>`. Is this reasonable? I've also tried changing the type to something like `Option<T> where T: Into<BlockId>`. However, this has the downside that every time we pass a `None`, we have to type annotate the calls. Thoughts?

The second relates to testing. Ganache does not implement EIP-1898 (trufflesuite/ganache-cli#745). It can be tested with geth, however, which is what I did.

How should I proceed? Should I open a PR?

Cheers!

gakonst:
I personally find the `None<()>` annotation to be super ugly, so I'd more supportive of this first approach. This would be a breaking change (and that's fine since we're pre-1.0), and would be easily fixable by passing `BlockNumber::Latest.into()` (or similar) wherever the API is consumed

Testing the feature with geth SGTM!

PR #231, opened 2021-03-16
feat: add support for EIP-234 and EIP-1898

GCdePaula:
Related issue: #230

## Motivation

These two EIPs are closely related, and allow for querying the blockchain's state at a specified block hash. Which is quite useful when dealing chain reorgs and allows for more robust clients.

## Solution

For EIP-234, add method `at_block_hash` for `Event` and `Filter`.

For EIP-1898, modify `Middleware` functions affected by the EIP (that is, most that receive a `BlockNumber`), changing their signatures to receive a `BlockId` instead. Also, change the method `ContractCall::block` to receive a `BlockId`, instead of a `BlockNumber`.

### DIFF
--- a/ethers-contract/src/call.rs
+++ b/ethers-contract/src/call.rs
@@ -1,7 +1,7 @@
 use super::base::{decode_function_data, AbiError};
 use ethers_core::{
     abi::{Detokenize, Function, InvalidOutputType},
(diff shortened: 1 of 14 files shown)


gakonst (approved):
LGTM, thank you.

Issue #228, opened 2021-03-15
Implement the `EthEvent` trait for generated event code in abigen


gakonst:
In https://github.com/gakonst/ethers-rs/pull/227, @mattsse introduced the `EthEvent` trait, for implementing event bindings.

Currently, we only codegen the [method implementations without the trait](https://github.com/gakonst/ethers-rs/blob/732ff29d1428e059803c2394fd85315a6e2850b7/ethers-contract/ethers-contract-abigen/src/contract/events.rs#L106-L121). We should make it so that the codegen also generates the trait.

Another idea could be to add a new method, which would allow us to do something like what we do [here](https://github.com/gakonst/ethers-rs/blob/master/ethers-contract/src/contract.rs#L183-L193).

```rust
trait EthEvent {
    fn new<M: Middleware>(filter: &Filter, provider: &M) -> Result<Vec<Self>, SomeError> {
          let filter = filter.clone().event(Self::abi_signature());
          Event {
               filter,
               provider,
               datatype: PhantomData,
               event: <... something that's compatible with the ABI definition>
         }
    }

    // <... other methods>
}

#[derive(Debug, Clone, EthEvent)]
struct MyEvent {
     address foo;
}

let filter = Filter::new().address(MY_ADDRESS).from_block(0).to_block(100);
let provider = Provider::<Http>::try_from("...")?;
let events = MyEvent::new(filter, provider).query().await?;
```

This allows us to conveniently query and subscribe to strongly typed events, without having to instantiate a full contract object (like we show in the [examples](https://github.com/gakonst/ethers-rs/blob/master/ethers/examples/contract_with_abi.rs#L59-L64))

gakonst:
Closed in #232

PR #232, opened 2021-03-16
feat(abigen): include ethevent proc macro in abigen code gen workflow

mattsse:
This integrates the `EthEvent` trait in the `abigen` code generation process.
## Motivation
The `EthEvent` makes the output of the `abigen` macro more consistent for events, and allows for generating further helper methods, as mentioned in #228.

## Solution
Instead of generating the methods `name`, `signature` and `abi_signature` for each event provided in the ABI, the `EthEvent` trait is simply implemented via a `#[derive(EthEvent)]` during the `abigen` workflow. In addition all structs that are declared in the human readable abi are generated aswell. If a solidity event contains variables with solidity structs as tyoe , the generated rust struct variables have the corresponding rust. This has currently a limitation where the build fails if a solidity event has an array of a struct as input, however https://github.com/gakonst/ethabi/pull/1 should remove this limitation.

## Open Discussion
* The generated event struct still use a `Filter` name suffix, I think this is not necessary and could be removed.
* To use the `EthEvent` methods as before, an import of the `EthEvent` is required, right now the generated contract module does not `pub EthEvent`, which would be breaking as well as the minor change from `&'static str` to `Cow<'static, str>` in `EthEvent::name` and `EthEvent::abi_signature`, I'd think that is negligible.

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/contract.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract.rs
@@ -2,11 +2,13 @@
 mod common;
 mod events;
 mod methods;
+mod structs;
 mod types;
 
 use super::util;
 use super::Abigen;
 use anyhow::{anyhow, Context as _, Result};
+use ethers_core::abi::AbiParser;
 use ethers_core::{
     abi::{parse_abi, Abi},
     types::Address,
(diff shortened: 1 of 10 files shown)


gakonst:
Merged, thank you. Supportive of removing the `Filter` suffix from the struct definition

PR #239, opened 2021-03-18
feat(abigen): extend ethevent trait methods and decoding

### LINKED ISSUE #98: Events from Receipts
Ideally for the following contract's events:

```solidity 
contract C {
    event X(...);
    event Y(...);
}
```

The resulting codegen should be

```rust
#[derive(EthEvent)]
struct X {
   ...
}

#[derive(EthEvent)]
struct Y {
  ...
}

#[derive(EthEvent)]
enum XY {
       X(X),
       Y(Y),
}

// parse_logs will try to decode each individual log inside the receipt as either X or Y 
let ev: Vec<XY> = ethers_contract::parse_logs(tx_receipt)?;
```

It should be possible to use the enum as a union type to parse event logs of various types, allowing us to e.g. loop over a transaction's logs. Also, it'd be nice if we had this in a function like [`all_events`](https://github.com/gnosis/ethcontract-rs/blob/main/ethcontract-generate/src/contract/events.rs#L400)  (see [here](https://github.com/gnosis/ethcontract-rs/blob/main/ethcontract-generate/src/contract/events.rs#L387-L446) for the macro def).

### LINKED ISSUE #234: EthEvent::new method for getting typed event data without instantiating a contract

We could add a new method to `EthEvent`, which would allow us to do something like what we do [here](https://github.com/gakonst/ethers-rs/blob/master/ethers-contract/src/contract.rs#L183-L193).

```rust
trait EthEvent {
    fn new<M: Middleware>(filter: &Filter, provider: &M) -> Result<Vec<Self>, SomeError> {
          let filter = filter.clone().event(Self::abi_signature());
          Event {
               filter,
               provider,
               datatype: PhantomData,
               event: <... something that's compatible with the ABI definition>
         }
    }

    // <... other methods>
}

#[derive(Debug, Clone, EthEvent)]
struct MyEvent {
     address foo;
}

let filter = Filter::new().address(MY_ADDRESS).from_block(0).to_block(100);
let provider = Provider::<Http>::try_from("...")?;
let events = MyEvent::new(filter, provider).query().await?;
```

This allows us to conveniently query and subscribe to strongly typed events, without having to instantiate a full contract object (like we show in the [examples](https://github.com/gakonst/ethers-rs/blob/master/ethers/examples/contract_with_abi.rs#L59-L64))

mattsse:
## Motivation
This follows up on #234 and also provides a solution for #98

## Solution
The `EthEvent` trait is extended with a method to create an `Event` builder directly (closes #234).
The `EthEvent` proc macro now derives a `decode_log(log: &RawLog)` that parses the log message into an instance of the type it's derived on.
This introduces another trait `EthLogDecode` which only purpose is to create a common `RawLog -> Self` interface for all the generated `EthEvent`s of a contract and the generated enum wrapper that is generated for events that have more than 1 event in their abi (closes #98)

### Example
if a contract is created as follows

```rust
abigen!(
    SimpleContract,
    r#"[
    event ValueChanged(address indexed author, string oldValue, string newValue)
    event OwnerChanged(address indexed author, address indexed author)
    event Log(string msg)
]"#
);
```
an enum 

```
pub enum SimpleContractEvents {
    LogFilter(LogFilter),
    OwnerChangedFilter(OwnerChangedFilter),
    ValueChangedFilter(ValueChangedFilter),
}
```

is created. 
In addition, if the abi has at least one event the generated contract gets a new method `events(&self) -> Event<M,  Events>`, where `Events` is either the sole event for contracts with only a single event or the generated enum for contracts with more than one event. 
In the latter case, the protocol decoding works in such a way that all variants are tried to be decoded one after the other and the first successful attempt is returned

## Breaking changes
* Renamed the `Contract::event(&self, &str) -> Result<Event<_>,Error>` to `event_for_name` and introduced `Contract::event<T:EthEvent>(&self) -> Event<_>` + another convenience function

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/contract.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract.rs
@@ -79,15 +79,14 @@ impl Context {
             #[allow(clippy::too_many_arguments)]
             mod #name_mod {
                 #imports
-
                 #struct_decl
 
-                impl<'a, M: Middleware> #name<M> {
+                impl<'a, M: ethers_providers::Middleware> #name<M> {
                     /// Creates a new contract instance with the specified `ethers`
                     /// client at the given `Address`. The contract derefs to a `ethers::Contract`
                     /// object
-                    pub fn new<T: Into<Address>>(address: T, client: Arc<M>) -> Self {
-                        let contract = Contract::new(address.into(), #abi_name.clone(), client);
+                    pub fn new<T: Into<ethers_core::types::Address>>(address: T, client: ::std::sync::Arc<M>) -> Self {
+                        let contract = ethers_contract::Contract::new(address.into(), #abi_name.clone(), client);
                         Self(contract)
                     }
 
--- a/ethers-contract/ethers-contract-abigen/src/contract/common.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract/common.rs
@@ -15,11 +15,12 @@ pub(crate) fn imports(name: &str) -> TokenStream {
         use std::sync::Arc;
         use ethers::{
             core::{
+                self as ethers_core,
                 abi::{Abi, Token, Detokenize, InvalidOutputType, Tokenizable},
                 types::*, // import all the types so that we can codegen for everything
             },
-            contract::{Contract, builders::{ContractCall, Event}, Lazy},
-            providers::Middleware,
+            contract::{self as ethers_contract, Contract, builders::{ContractCall, Event}, Lazy},
+            providers::{self as ethers_providers,Middleware},
         };
     }
 }
--- a/ethers-contract/ethers-contract-abigen/src/contract/common.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract/common.rs
@@ -31,12 +31,12 @@ pub(crate) fn struct_declaration(cx: &Context, abi_name: &proc_macro2::Ident) ->
 
     let abi_parse = if !cx.human_readable {
         quote! {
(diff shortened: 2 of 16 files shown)


ethers-contract/ethers-contract-abigen/src/contract.rs
```diff
@@ -79,15 +79,14 @@ impl Context {
             #[allow(clippy::too_many_arguments)]
             mod #name_mod {
                 #imports
-
                 #struct_decl
 
-                impl<'a, M: Middleware> #name<M> {
+                impl<'a, M: ethers_providers::Middleware> #name<M> {
```
gakonst, line 84: Having fully qualified syntax here requires that we have to declare ethers_providers/core/contract in the consuming crate's cargo toml. It'd be ideal if we avoided that. Could you please revert this change?

gakonst (changes requested):
1 comment as discussed on the fully qualified syntax, otherwise looks great to me.

ethers-contract/ethers-contract-abigen/src/contract.rs
mattsse: Instead of removing the fully qualified syntax, I simply renamed the top level imports to their crate name, which achieves the same effect, plus has the benefit, that the `EthEvent` trait, which resides in `ethers_contract` is usable with `ethers_contract` as crate dependency.

gakonst (approved):

ethers-contract/ethers-contract-abigen/src/contract.rs
gakonst: Awesome, thank you, merged.

PR #241, opened 2021-03-21
feat: add async setup utility functions

mattsse:
## Motivation
This adds async helper functions for compiling contracts and setting up `geth` or `ganache`.
Using the async `compile_and_launch_ganache` will only take as long as the longest running command (`solc`, `ganache-cli` ). 
This will shave off the solidity compile time for the updated examples, because this in those cases the compile time is very short. And it will shave off the `~1.8s` it takes to launch ganache for those cases where `solc` takes longer than that.

## Solution
Feature gate a `setup` module in `ethers-core` that simply runs `tokio::task::blocking` for the commands (`solc`, `geth`, `ganache`) and join those task when using `compile_and_launch_ganache` or `compile_and_launch_geth`.
By default the `setup` feature of the `ethers-core` crate is enabled in `ethers` exposing those helper functions from the `ethers::util` module.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -693,6 +693,7 @@ dependencies = [
  "ethereum-types",
  "ethers",
  "funty",
+ "futures-util",
  "generic-array",
  "glob",
  "hex",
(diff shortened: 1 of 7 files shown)


gakonst (approved):
LGTM! This should become more useful once/when somebody decides to build a full dev framework on top of ethers-rs, so that it compiles the code + launches the client!

Issue #106, opened 2020-12-17
IPC Support


gakonst:
Rust-web3 added an IPC transport https://github.com/tomusdrw/rust-web3/issues/415 which we can port over.

PR #260, opened 2021-04-06
feat: implement IPC transport support

austinabell:
## Motivation

#106 I'm just playing with it to learn more about IPC connections and Tokio

## Solution

This is very rough as is. Just opening draft and I'll hopefully cleanup tomorrow and do more extensive testing. ~The connection seems to happen alright, but a warning is coming up that the method doesn't exist, so I'll look into that tomorrow too~ Basic requests are now functional, but will test subscription based requests later. Is there any specific workflow that's ideal to test this?


gakonst:
You can launch a `geth` instance like we do [here](https://github.com/gakonst/ethers-rs/blob/a22f1f9aa06836f7ab76bf65b9f06a6d1d234709/ethers-providers/tests/txpool.rs#L10), specify a temp ipc socket directory, connect to it via provider and test it that way. Ideally the socket file should be removed at the end of the test too!

ethers-providers/Cargo.toml
```diff
@@ -24,6 +24,7 @@ serde_json = { version = "1.0.64", default-features = false }
 thiserror = { version = "1.0.24", default-features = false }
 url = { version = "2.2.1", default-features = false }
 auto_impl = { version = "0.4.1", default-features = false }
+futures = { version = "0.3.5", default-features = false, optional = true }
```
gakonst, line 27: You do not need this dep, you can get `StremExt` from `futures-util` https://docs.rs/futures-util/0.3.13/futures_util/stream/trait.StreamExt.html

austinabell:
Is there a reason I'm missing for why a BTreeMap is used for the request ID lookups in the ws impl? I don't understand why the BTreeMap overhead is needed for lookups. I'm going to use a regular HashMap, but happy to switch back if there is a good reason.

ethers-providers/Cargo.toml
```diff
@@ -24,6 +24,7 @@ serde_json = { version = "1.0.64", default-features = false }
 thiserror = { version = "1.0.24", default-features = false }
 url = { version = "2.2.1", default-features = false }
 auto_impl = { version = "0.4.1", default-features = false }
+ahash = { version = "0.7", default-features = false, features = ["std"], optional = true }
```
tarcieri, line 27: If you really need higher performance than `SipHasher` (I would suggest measuring), and don't care about hashDoS (the `ahash` crate is not designed to be hashDoS resistant), I would suggest using the [`twox-hash` crate](https://crates.io/crates/twox-hash) which implements the XXHash algorithm.

austinabell: Yeah I'm aware of this, but figured this usage wouldn't require that property, but maybe I'm overlooking some concern. Thanks for the link, I'll look into that, but I'll just change to whatever @gakonst prefers since I don't have an opinion.

austinabell: I've switched it to siphash as a default, but happy to switch to whatever is preferred :)

tarcieri: Sounds good. I think using what the standard library provides is a reasonable starting point, and if you need better performance, it would be good to measure whether a 3rd-party hashing crate actually provides an improvement.

ethers/examples/ipc.rs
```diff
@@ -0,0 +1,17 @@
+use ethers::prelude::*;
+use std::time::Duration;
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    // TODO switch to some default IPC after
+    let ws = Ipc::new("/home/austinabell/.ethereum/geth.ipc").await?;
```
gakonst, line 7:
```suggestion
    let ws = Ipc::new("~/.ethereum/geth.ipc").await?;
```

ethers-providers/src/transports/ws.rs
```diff
@@ -347,6 +347,8 @@ mod tests {
 
         // Subscribing requires sending the sub request and then subscribing to
         // the returned sub_id
+        let block_num: U256 = ws.request("eth_blockNumber", ()).await.unwrap();
+        println!("block num2 {}", block_num);
```
gakonst, line 351:
```suggestion
```

gakonst (approved):

gakonst:
Merged, thank you! If you want to join, we're discussing the library in [this chatroom](https://t.me/ethers_rs).

PR #288, opened 2021-05-08
feature: Signature derives Copy trait

prestwich:
derives `Copy` on `Signature`

## Motivation

I'd like `Signature` to be `Copy`

## Solution

Derive `Copy`

## Drive-by change

Set network ID in signer in Celo tests


prestwich:
Pretty sure the IPC test is a race condition?

PR #294, opened 2021-05-12
Expose `LogMeta` type and add more fields, and expose FilterBlockOption

GCdePaula:
## Motivation

The types `LogMeta` and `FilterBlockOption` are not public, and I believe they should be. Also, the `LogMeta` is rather terse, and could use more metadata.

## Solution

I've exported the two types and added more fields to `LogsMeta`.

However, there's one important consideration on adding more fields. The procedure that maps the type `Log` into `LogMeta` heavily uses `expect` on Option types. This behavior follows what was previously there, just extended to the new fields. If we look at the [json-rpc API](https://eth.wiki/json-rpc/API#eth_getfilterchanges), it says these fields are only `null` when the block is pending. In other words, this is at least as safe it was previously, and it is probably safe. One possible alternative is implementing `TryFrom`, instead of `From`. But maybe this can be done in another PR.

Cheers!

### DIFF
--- a/ethers-contract/src/event.rs
+++ b/ethers-contract/src/event.rs
@@ -2,7 +2,7 @@ use crate::{stream::EventStream, ContractError, EthLogDecode};
 
 use ethers_core::{
     abi::{Detokenize, RawLog},
-    types::{BlockNumber, Filter, Log, TxHash, ValueOrArray, H256, U64},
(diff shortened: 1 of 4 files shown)


gakonst (approved):
LGTM. Thank you for the change. I don't think this would be an issue for mined blocks for the reasons you wrote. If somebody encounters this in the future, we could rename the current function to something like `query_with_meta_unchecked` and make the `query_with_meta` return a `Result`.

PR #296, opened 2021-05-22
fix: off by one error during log topic decoding

mattsse:
## Motivation

Fixes a bug, that is essentially an off by one error, that resulted in errors when decoding events with no indexed entries. 

## Solution

Remove the bad `is_empty` requirement and test against real log.

### DIFF
--- a/ethers-contract/ethers-contract-derive/src/lib.rs
+++ b/ethers-contract/ethers-contract-derive/src/lib.rs
@@ -432,7 +432,7 @@ fn derive_decode_from_log_impl(
                 let flat_topics = topics.iter().skip(1).flat_map(|t| t.as_ref().to_vec()).collect::<Vec<u8>>();
             },
             quote! {
-                if topic_tokens.is_empty() || topic_tokens.len() != topics.len() - 1 {
+                if topic_tokens.len() != topics.len() - 1 {
                     return Err(ethers_core::abi::Error::InvalidData);
                 }
             },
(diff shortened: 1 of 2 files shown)


ethers-contract/ethers-contract-derive/src/lib.rs
```diff
@@ -432,7 +432,7 @@ fn derive_decode_from_log_impl(
                 let flat_topics = topics.iter().skip(1).flat_map(|t| t.as_ref().to_vec()).collect::<Vec<u8>>();
             },
             quote! {
-                if topic_tokens.is_empty() || topic_tokens.len() != topics.len() - 1 {
+                if topic_tokens.len() != topics.len() - 1 {
```
prestwich, line 435: can't this subtraction underflow if the `LOG0` opcode is used?

mattsse: The macro generates different code for anonymous events:
https://github.com/gakonst/ethers-rs/blob/94797f526936613c6b4c987b9518d5ba61e1bd67/ethers-contract/ethers-contract-derive/src/lib.rs#L411-L440

only for non anonymous it makes sure that the signature is the first entry in `topics`
https://github.com/gakonst/ethers-rs/blob/94797f526936613c6b4c987b9518d5ba61e1bd67/ethers-contract/ethers-contract-derive/src/lib.rs#L426
so `topics.len() - 1` can't underflow. 

Most of this is adapted from:
https://github.com/rust-ethereum/ethabi/blob/970d47018e0f457440851752a7b46a4dbf793a0b/ethabi/src/event.rs#L122-L173

gakonst (approved):

PR #305, opened 2021-06-01
feat: make I256::from_raw() as const

guanqun:
I'd like to make from_raw() a const fn.


guanqun:
Added another commit to fix the `cargo fmt` issue in CI.

gakonst (approved):

gakonst:
Merged - thank you:)

PR #314, opened 2021-06-11
feat: extend Middleware trait customized for celo

hanyunx:
## Motivation
We would like to have a customized `provider` for Celo so as to add more functions exclusively provided by Celo blockchain.

## Solution
A new trait `CeloMiddleware` is added to extend the `Middleware` for provider, so that functions in `Middleware` will query APIs compatible with Ethereum while functions in `CeloMiddleware` will query those exclusively in Celo.

The method `get_validators_bls_public_keys()` is added as an example.

### DIFF
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -561,3 +561,17 @@ pub trait Middleware: Sync + Send + Debug {
             .map_err(FromErr::from)
     }
 }
+
+#[cfg(feature = "celo")]
+#[async_trait]
+pub trait CeloMiddleware: Middleware {
+    async fn get_validators_bls_public_keys(
+        &self,
+        block_number: String,
+    ) -> Result<Vec<String>, ProviderError> {
+        self.provider()
+            .get_validators_bls_public_keys(block_number)
+            .await
+            .map_err(FromErr::from)
+    }
+}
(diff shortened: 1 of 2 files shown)


ethers-providers/src/lib.rs
```diff
@@ -561,3 +561,17 @@ pub trait Middleware: Sync + Send + Debug {
             .map_err(FromErr::from)
     }
 }
+
+#[cfg(feature = "celo")]
+#[async_trait]
+pub trait CeloMiddleware: Middleware {
+    async fn get_validators_bls_public_keys(
+        &self,
```
gakonst, line 569: This should probably be an `Option<BlockId>`? https://github.com/gakonst/ethers-rs/blob/b4f490844b5d81c40038b2c6383b0292cc8c99ed/ethers-providers/src/lib.rs#L252

ethers-providers/src/lib.rs
```diff
@@ -561,3 +561,17 @@ pub trait Middleware: Sync + Send + Debug {
...
+        block_number: String,
+    ) -> Result<Vec<String>, ProviderError> {
```
gakonst, line 571: How long are the keys? Is there any more specific type we could use instead of String?

gakonst (changes requested):
Supportive of having a Celo middleware. Minor nits but looks good otherwise.

ethers-providers/src/lib.rs
hanyunx: The public keys are 96 bytes. Maybe we can use `Vec<u8>` instead.

PR #316, opened 2021-06-13
feat: add filter utility function

mattsse:
## Motivation
Makes setting block ranges more convenient by adding a new function (`select`) to the `Filter` struct and convenient `From` impls.

Not sure about:
* the name
* whether to include `From<U256>` since `at_block_hash` does the same already.

### DIFF
--- a/ethers-core/src/types/log.rs
+++ b/ethers-core/src/types/log.rs
@@ -4,6 +4,7 @@ use crate::{
     utils::keccak256,
 };
 use serde::{ser::SerializeStruct, Deserialize, Serialize, Serializer};
+use std::ops::{Range, RangeFrom, RangeTo};
 
 /// A log produced by a transaction.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


ethers-core/src/types/log.rs
```diff
@@ -187,6 +247,72 @@ impl Filter {
...
+    /// Match a range of blocks
+    ///
+    /// ```rust
+    /// # use ethers::types::{Filter, H256};
+    /// # fn main() {
+    /// let filter = Filter::new().select(0u64..100u64);
+    /// # }
+    /// ```
+    ///
+    /// Match a all blocks in range `(1337..BlockNumber::Latest)`
```
gakonst, line 294:
```suggestion
    /// Match all blocks in range `(1337..BlockNumber::Latest)`
```

ethers-core/src/types/log.rs
```diff
@@ -187,6 +247,72 @@ impl Filter {
...
+    ///
+    /// ```rust
+    /// # use ethers::types::{Filter, H256};
+    /// # fn main() {
+    /// let filter = Filter::new().select(1337u64..);
+    /// # }
+    /// ```
+    ///
+    /// Match a all blocks in range `(BlockNumber::Earliest..1337)`
```
gakonst, line 303:
```suggestion
    /// Match all blocks in range `(BlockNumber::Earliest..1337)`
```

gakonst (approved):

PR #319, opened 2021-06-17
fix: implement the arithematic shr

guanqun:
for the shr, we should take the sign into account, unless we explicitly say that the 'shr' is logical shift right operation.


gakonst:
Closing - am reluctant to changing the behavior, added a note that it's a logical shift operation in #329

guanqun:
Sure, should we add an arithematic shr?

gakonst:
Sounds good to me!

PR #320, opened 2021-06-17
fix: change the type of block number and pub key

hanyunx:
A follow up on #314, changed the type of block number to the generic `Into<BlockId>` so that the client can query with the input of type `u64` other than converting `u64` to `BlockId` manually. 

Changed the type of returned public key to `String` since the client cannot deserialize the response data otherwise.


hanyunx:
The clippy failed in files that were not touched, any idea why this happened? @gakonst

gakonst:
Clippy error is unrelated. Will fix.

PR #327, opened 2021-07-03
feature: PendingTransaction returns Option

### LINKED ISSUE #326: `PendingTransaction` hangs indefinitely if the tx never confirms
**Version**
`main`

**Description**
`PendingTransaction` implementation of `Future` doesn't account for the case where the TX is dropped from the mempool. As a result, dropped txns cause the pending tx to flip between states `PausedGettingReceipt` and `GettingReceipt` forever without resolving

prestwich:
## Motivation

closes #326 

## Solution

- Add 2 new states to the PendingTransaction future
- Change the output type to `Option<TransactionReceipt>`
- `None` is returned when the tx drops from the mempool without confirming
- Refactor `PendingTransaction` internals to prevent repeated cloning of the receipt
- Refactor code that relies on `PendingTransaction` to handle the `None` case

### DIFF
--- a/ethers-contract/src/factory.rs
+++ b/ethers-contract/src/factory.rs
@@ -45,7 +45,8 @@ impl<M: Middleware> Deployer<M> {
         let receipt = pending_tx
             .confirmations(self.confs)
             .await
-            .map_err(|_| ContractError::ContractNotDeployed)?;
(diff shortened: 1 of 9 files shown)


prestwich:
some failing tests are related to changes in streaming tests. can someone with more context take a look?

gakonst (approved):

Issue #325, opened 2021-06-28
Add `rustls` as a feature flag or use it by default.


shekohex:
**Is your feature request related to a problem? Please describe.**
Currently, `ethers-providers` crate uses `tokio-tungstenite` with `tls` feature flag, which would use `async-native-tls` crate.
in Linux that means Openssl .. and you know what happens when you use Openssl and try to compile your cli/project to `musl` target!.

**Describe the solution you'd like**
I would like to have an option to enable the `rustls-tls` feature instead of the `tls` default one.
see: https://docs.rs/crate/tokio-tungstenite/0.14.0/features

**Describe alternatives you've considered**
Enable `rustls-tls` by default, which is more secure and more importantly easy to cross compile.

**Additional context**
CI Fails because of this: https://github.com/webb-tools/webb-rs/runs/2910394664?check_suite_focus=true

gakonst:
@shekohex thanks for the issue! Could you check if #330 fixes it for you?

shekohex:
Yes, that is awesome.
Thank you!

PR #330, opened 2021-07-05
feat: allow using native-tls or rustls

gakonst:
Closes https://github.com/gakonst/ethers-rs/issues/325:
* Makes Rust TLS the default feature being used with tokio-tungstenite/reqwest
* Adds an `openssl` feature for using native bindings otherwise


shekohex (approved):
LGTM!

PR #333, opened 2021-07-06
Update crypto deps for ethers

zmanian:
Update RustCcrypto deps for ethers.

<s>Waiting on @roynalnaruto to merge roynalnaruto/eth-keystore-rs#4 then will update this PR.</s>


gakonst (approved):
LGTM - test is flaky

PR #353, opened 2021-07-30
feat: eip2930/1559 response type adjustments (part 1)

gakonst:
* Refactors the tx types to its own subdirectory, to prepare for adding more tx request types / serialization helpers
* Adds `base_fee_per_gas` to `Block`
* Adds `access_list`, `transaction_type`, `max_priority_fee_per_gas`, `max_fee_per_gas` to `Transaction`
* Adds `transaction_type`, `effective_gas_price` to `TransactionReceipt`
* `SignerMiddleware::sign_transaction` now returns an rlp encoded signed transaction instead of a `Transaction`
* `Middleware::send_raw_transaction` now receives a signed rlp transaction instead of a `Transaction`

### DIFF
--- a/ethers-core/src/types/transaction/eip2930.rs
+++ b/ethers-core/src/types/transaction/eip2930.rs
@@ -1,8 +1,5 @@
-use super::request::TransactionRequest;
-use super::NUM_TX_FIELDS;
-use crate::types::{Address, Bytes, Signature, H256, U64};
+use crate::types::{Address, H256};
 
-use rlp::RlpStream;
 use rlp_derive::RlpEncodable;
 use serde::{Deserialize, Serialize};
 
--- a/ethers-core/src/types/transaction/eip2930.rs
+++ b/ethers-core/src/types/transaction/eip2930.rs
@@ -1,0 +1,27 @@
+use super::request::TransactionRequest;
+use super::NUM_TX_FIELDS;
+use crate::types::{Address, Bytes, Signature, H256, U64};
+
+use rlp::RlpStream;
+use rlp_derive::RlpEncodable;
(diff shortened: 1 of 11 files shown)


ethers-core/src/types/transaction/eip2930.rs
```diff
@@ -0,0 +1,24 @@
+use crate::types::{Address, H256};
+
+use rlp_derive::RlpEncodable;
+use serde::{Deserialize, Serialize};
+
+/// Access list
+#[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize, RlpEncodable)]
```
guanqun, line 7: I think I hit one bug here, `RlpEncodable` should be `RlpEncodableWrapper`.  unless you're using `.0` to do the rlp encoding, otherwise you got an extra "[]".

mattsse (approved):
splitting the transaction mod definitely makes it more maintainable and provides more structure.

ethers-core/src/types/transaction/response.rs
```diff
@@ -0,0 +1,290 @@
...
+    pub gateway_fee_recipient: Option<Address>,
+
+    /// Gateway fee amount (None for no gateway fee paid)
+    #[cfg(feature = "celo")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "celo")))]
+    #[serde(skip_serializing_if = "Option::is_none", rename = "gatewayFee")]
+    pub gateway_fee: Option<U256>,
+
+    // EIP2718
+    /// Transaction type, Some(1) for AccessList transaction, None for Legacy
```
onbjerg, line 90: Should probably also mention that `Some(2)` is EIP1559

PR #354, opened 2021-07-30
feat: stream_with_meta

gakonst:
As title, allows one to stream logs with metadata, by calling `with_meta` on the stream before polling it.


shekohex (approved):
LGTM!
Thank you!

PR #355, opened 2021-07-31
feat: 1559/2930 txs (part 2)

gakonst:
* Introduces the eip1559/2930 tx types 
* Introduces a `TypedTransaction` enum w/ `From` implementations for each type, which will replace `TransactionRequest` in the rest of the codebase
* Fixes a bug pointed out by @guanqun on the `AccessList` RLP encoding derivation method
* Adds tests to ensure that RLP encoding and the sighash matches Geth's

### DIFF
--- a/ethers-core/src/types/transaction/eip2718.rs
+++ b/ethers-core/src/types/transaction/eip2718.rs
@@ -1,0 +1,70 @@
+use super::eip2930::Eip2930TransactionRequest;
+use crate::{
+    types::{TransactionRequest, H256, U64},
+    utils::keccak256,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Debug)]
+#[serde(tag = "type")]
+pub enum TypedTransaction {
+    // 0x00
+    #[serde(rename = "0x00")]
+    Legacy(TransactionRequest),
+    // 0x01
+    #[serde(rename = "0x01")]
+    Eip2930(Eip2930TransactionRequest),
+}
+
+impl TypedTransaction {
+    /// Hashes the transaction's data with the provided chain id
+    pub fn sighash<T: Into<U64>>(&self, chain_id: T) -> H256 {
+        let encoded = match self {
+            TypedTransaction::Legacy(ref tx) => {
+                let mut encoded = vec![0];
+                encoded.extend_from_slice(tx.rlp(chain_id).as_ref());
+                encoded
+            }
+            TypedTransaction::Eip2930(ref tx) => {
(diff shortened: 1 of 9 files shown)


ethers-core/src/types/transaction/eip2718.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Debug)]
+#[serde(tag = "type")]
+pub enum TypedTransaction {
+    // 0x00
+    #[serde(rename = "0x00")]
+    Legacy(TransactionRequest),
+    // 0x01
+    #[serde(rename = "0x01")]
+    Eip2930(Eip2930TransactionRequest),
```
guanqun, line 16: Can I request a rename of these transaction names?

I prefer python's version of "AccessListTransactionRequest" and "DynamicFeeTransactionRequest" over these number named one.

Eip2930 and Eip1559 are not straightforward, it's hard to figure out the underlying meaning.

guanqun: A sample definition here: https://github.com/guanqun/ethers-rs/blob/feat/typed-transactions/ethers-core/src/types/typed_transaction.rs#L42

mattsse (approved):
only checked syntax, not the actual encoding.

guanqun:
Thank you for the feature.

My only comment is the naming. These EIP numbers are just magic numbers, we should avoid using magic numbers.

How do you think? @gakonst @mattsse

ethers-core/src/types/transaction/eip2718.rs
gakonst: Hey - we discussed it in the dev chat, I preferred the EIPxxx numbering, confirmed by @prestwich @roynalnaruto @mattsse, so we'll keep the EIPXXX numbering for tx types!

gakonst:
@guanqun replied above, let's stick with the EIP numbers, they're numbers that everyone in Ethereum knows about (EIP-1559 is very specific about what it gives, vs "Dynamic Fee" can mean literally anything)

guanqun:
Thanks for the info.

I agree that EIP-1559 is popular, but from my own view, other two EIP-2930 and EIP-2718 are less popular. I doubt that people would immediately know the meaning behind them.

That said, you have the final call.

guanqun:
FYI. A few small changes are needed to make it work on goerli network. The biggest one missing here is the leading header byte. I hack around and come up with a patch here, if it's useful to others. https://github.com/guanqun/ethers-rs/commit/68de2a26680d7e14a6eca858ebdac50430a64103

PR #356, opened 2021-07-31
chore: cargo clippy --fix && rustfmt format_code_in_doc_comments wrap_comments

mattsse:
## Motivation
Run cargo clippy fix with all features enabled and format with more rustfmt configs

## Solution

`cargo clippy --fix --all-features && cargo fmt --all -- --config format_code_in_doc_comments=true wrap_comments=true `


gakonst:
@mattsse given the 1559 changes were happening, I didn't want to merge it back then because it'd cause a lot of conflicts. Do you want to re-apply the same commands on top of master?

gakonst:
Closing due to multiple changes in the codebase :S Bad timing, if not much effort and you want to re-execute the command that works for me!

PR #358, opened 2021-08-01
feature: signer using aws kms

prestwich:
Adds an AWS KMS signer to ethers-signers

Some notes:
1. haven't found a mock KMS that supports ecdsa signing. So can't test this outside of a real aws account
2. It is working with my AWS account
3. added under the `aws` feature
4. The `utils` functions are NOT for general consumption
5. The public key is retrieved on instantiation. ergo, instantiation is async

TODO:
- [x] documentation
- [x] instantiation notes and examples

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,6 +67,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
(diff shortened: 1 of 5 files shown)


gakonst (approved):
legendary

PR #362, opened 2021-08-02
feat(ethers-contract): typed txs (part 4)

gakonst:
* Adds `TypedTransaction` inside the `ContractCall` object
* Uses EIP1559 Transactions by default (Legacy if `legacy` feature flag is on), allows falling back to legacy ones via the `legacy` helper function
* Most tests need to use `legacy()` because ganache does not support Berlin or London yet (?!?)

### DIFF
--- a/ethers-contract/src/call.rs
+++ b/ethers-contract/src/call.rs
@@ -1,7 +1,9 @@
 use super::base::{decode_function_data, AbiError};
 use ethers_core::{
     abi::{Detokenize, Function, InvalidOutputType},
-    types::{Address, BlockId, Bytes, TransactionRequest, U256},
+    types::{
+        transaction::eip2718::TypedTransaction, Address, BlockId, Bytes, TransactionRequest, U256,
+    },
 };
 use ethers_providers::{Middleware, PendingTransaction, ProviderError};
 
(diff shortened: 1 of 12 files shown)


ethers-contract/src/call.rs
```diff
@@ -60,25 +62,36 @@ pub struct ContractCall<M, D> {
...
+        self.tx.set_from(from.into());
+        self
+    }
+
+    /// Uses a Legacy transaction instead of an EIP-1559 one to execute the call
+    pub fn legacy(mut self) -> Self {
+        if let TypedTransaction::Eip1559(inner) = self.tx.clone() {
+            let tx: TransactionRequest = inner.into();
+            self.tx = TypedTransaction::Legacy(tx);
+        }
```
mattsse, line 74:
```suggestion
        self.tx = match self.tx {
            TypedTransaction::Eip1559(inner) => {
                let tx: TransactionRequest = inner.into();
                TypedTransaction::Legacy(tx)
            }
            other => other,
        };
```

ethers-contract/src/factory.rs
```diff
@@ -31,6 +34,15 @@ impl<M: Middleware> Deployer<M> {
         self
     }
 
+    /// Uses a Legacy transaction instead of an EIP-1559 one to do the deployment
+    pub fn legacy(mut self) -> Self {
+        if let TypedTransaction::Eip1559(inner) = self.tx.clone() {
+            let tx: TransactionRequest = inner.into();
+            self.tx = TypedTransaction::Legacy(tx);
+        }
+        self
```
mattsse, line 43: same as above

mattsse (approved):
lgtm, minor nits.

aidenbenner:
Thanks for the amazing work on this library!
Is this ready to merge? This would be very useful now that the London fork has landed.

gakonst:
I'll be working on the final bits to merge these PRs -today-, there's currently a serialization bug (along with some extra tests against testnets)

PR #363, opened 2021-08-02
fix: ensure tuple arguments are encoded as tuples

mattsse:
## Motivation
- single solidity struct arguments resulted in an invalid type check when encoding the function parameters, such as
```solidity
 function addLiquidityEthForUniV3(CellarAddParams calldata cellarParams)
```

## Solution
- Modify the generated function body so that tuples are treated as tuples when `encode_function_data` is invoked internally
-


zmanian:
This is test fixes the invoking a contract calls with tuple args for me.

gakonst (approved):

PR #366, opened 2021-08-05
feat: detect ethers crate paths in derive macros

mattsse:
## Motivation
- The `EthEvent` and `EthAbiType` derive macros are usable with `ethers_contract` and `ethers_core` only, therefore it depends on `ethers_core` which requires `use ethers::core as ethers_core;` alias to compile.

## Solution
- parse the project's `Cargo.toml` and check the dependencies for `ethers` dependency, if the dependency is present then `EthEvent` relies on `ethers::core`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -329,6 +329,37 @@ dependencies = [
 ]
 
 [[package]]
+name = "camino"
+version = "1.0.5"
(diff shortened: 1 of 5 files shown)


gakonst (approved):
absolute legend - thank you. LGTM.

PR #369, opened 2021-08-10
feat: fee estimation with custom/default fn

roynalnaruto:
## Motivation

Fee estimation/recommendation not available for EIP-1559 type transactions

## Solution

Default estimator is a port of [this](https://github.com/MyCryptoHQ/MyCrypto/blob/master/src/services/ApiService/Gas/eip1559.ts)

* Add `estimate_eip1559_fees` method to `Middleware` and implement in the `JsonRpcProvider`
* I believe none of the Gas oracle APIs support EIP-1559 fee recommendation (so that's simply an Err at the moment)
* Ethers-rs users can even provide their own `fn` to estimate fees using the `base_fee_per_gas` and `fee_history.reward`

### DIFF
--- a/ethers-contract/tests/contract.rs
+++ b/ethers-contract/tests/contract.rs
@@ -399,6 +399,7 @@ mod eth_tests {
         let _receipt = contract
             .method::<_, H256>("setValue", "hi".to_owned())
             .unwrap()
+            .legacy()
             .send()
             .await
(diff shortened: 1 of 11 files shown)


gakonst (changes requested):
So far so good - I like the approach. Could we also get some tests and remove any debug output logs?

PR #370, opened 2021-08-12
Support multiple addresses in filters

sebastinez:
## Motivation

According to the [ETH Wiki](https://eth.wiki/json-rpc/API#eth_newfilter) a filter should be able to define a contract address or a list of addresses from which logs should originate. The current state allows to define only one single address. 

## Solution

I added the `ValueOrArray` enum to the Log types and refactored some `ethers-contract` code with the mentioned enum.
To check the replacements there is also a new integration test in the `ethers-contract` package, that will deploy multiple contracts and check if the logs that originate are from the corresponding addresses.

### DIFF
--- a/ethers-contract/src/contract.rs
+++ b/ethers-contract/src/contract.rs
@@ -7,7 +7,7 @@ use crate::{
 
 use ethers_core::{
     abi::{Abi, Detokenize, Error, EventExt, Function, Tokenize},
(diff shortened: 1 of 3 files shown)


sebastinez:
> hey - I wanted to make some formatting changes and rebase this to master, but couldn't push to your branch. closing and re-opening there!

No problem @gakonst  thanks for maintaining my commit 👍

PR #378, opened 2021-08-14
feat: generate rust structs from solidity JSON ABI

mattsse:
## Motivation
Currently, parameters of solidity functions that are solidity structures are generated to rust tuples, which then requires some extra boilerplate code if the user wants to use rust structs within the ethers program.

## Solution
Since solidity 0.5.11 the solidity compiler adds an extra fields `internalType` to the inputs/outputs objects in the JSON ABI, which also contains structs with their qualified path (`struct SomeLibrary.StructName`). We can use this to essentially turn the tuple into a rust struct.

However this is not trivial because `ethabi` strips the internalType field entirely (https://github.com/rust-ethereum/ethabi/issues/230), therefor we need to parse the abi string again but keep the internalType. I made a minimal representation auf the abi (`RawAbi`) which supports that.

Next challenge are nested structs, again since the ethabi totally removes any information about the internaltype, we need to do additional work. Essentially we start with the structs that are used as inputs and outputs and check their fields recursively for structs. Since all structs boil down to elementary type, we will eventually end up with structs that only use elementary types. 
This tedious step is necessary, so that we don't end up with lots of struct duplicates.

- [x] turn JSON ABI tuples into rust structs
- [x] use them as input parameters for the corresponding contract calls
- [ ] use them as return types for the corresponding contract calls
- [ ] test some edge cases, like name collisions `LibraryA.Point` `LibraryB.Point`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -846,6 +846,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "reqwest",
+ "serde",
  "serde_json",
  "syn",
  "url",
--- a/ethers-contract/ethers-contract-abigen/Cargo.toml
+++ b/ethers-contract/ethers-contract-abigen/Cargo.toml
@@ -19,6 +19,7 @@ quote = "1.0"
 syn = "1.0.12"
 url = "2.1"
 serde_json = "1.0.61"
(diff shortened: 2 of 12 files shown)


gakonst:
As discussed, this seems to work well so far so merging as-is.

PR #380, opened 2021-08-14
feat: hardhat util

georgewhewell:
## Motivation
ganache-cli is not working well post-london hardfork

## Solution

add new `utils::hardhat` module copying `utils::ganache`

### DIFF
--- a/ethers-core/src/utils/hardhat.rs
+++ b/ethers-core/src/utils/hardhat.rs
@@ -1,0 +1,189 @@
+use crate::{
+    types::Address,
+    utils::{secret_key_to_address, unused_port},
+};
+use k256::{ecdsa::SigningKey, SecretKey as K256SecretKey};
+use std::{
+    io::{BufRead, BufReader},
+    process::{Child, Command},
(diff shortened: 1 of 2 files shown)


gakonst:
Hey @georgewhewell thanks for the contribution. @roynalnaruto also took a stab at this here: https://github.com/roynalnaruto/ethers-rs-1/blob/feat(util)/hardhat-node/ethers-core/src/utils/hardhat.rs

The issue with both approaches is that we need to have a `hardhat.config.js` somewhere in order for the process to work and a _local_ installation of hardhat, which means npm-installing stuff etc.

The Hardhat team is working on a global installation w/ no hardhat.config.js required, so ideally let's wait for that. An alternative would be to run a Hardhat Docker image.

georgewhewell:
ah cool- this implementation looks a bit more featured than mine so will close and wait that. fwiw i often have `node_modules` and `hardhat.config.js` in my projects already so don't need ethers to provide another but i see motivation for having this functionality work out-of-box

PR #385, opened 2021-08-17
add uncle related missing api

alexlee85:
## Motivation
missing uncle block related api

## Solution
add uncle block related missing api

### DIFF
--- a/ethers-providers/src/provider.rs
+++ b/ethers-providers/src/provider.rs
@@ -227,6 +227,47 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
         self.get_block_gen(block_hash_or_number.into(), true).await
     }
 
+    /// Gets the block uncle count at `block_hash_or_number`
+    async fn get_uncle_count<T: Into<BlockId> + Send + Sync>(
+        &self,
+        block_hash_or_number: T,
+    ) -> Result<U256, Self::Error> {
+        let id = block_hash_or_number.into();
+        Ok(match id {
+            BlockId::Hash(hash) => {
+                let hash = utils::serialize(&hash);
+                self.request("eth_getUncleCountByBlockHash", [hash]).await?
+            }
+            BlockId::Number(num) => {
+                let num = utils::serialize(&num);
+                self.request("eth_getUncleCountByBlockNumber", [num])
+                    .await?
+            }
+        })
+    }
+
+    /// Gets the block uncle at `block_hash_or_number` and `idx`
+    async fn get_uncle<T: Into<BlockId> + Send + Sync>(
+        &self,
+        block_hash_or_number: T,
+        idx: U64,
+    ) -> Result<Option<Block<H256>>, ProviderError> {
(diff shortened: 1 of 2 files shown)


ethers-providers/src/provider.rs
```diff
@@ -164,6 +164,40 @@ impl<P: JsonRpcClient> Provider<P> {
...
+            }
+            BlockId::Number(num) => {
+                let num = utils::serialize(&num);
+                self.request("eth_getUncleCountByBlockNumber", [num])
+                    .await?
+            }
+        })
+    }
+
+    async fn get_uncle_gen(
```
gakonst, line 182:
```suggestion
    async fn get_uncle(
```

The `_gen` suffix was added to the `get_block_gen` function, so that you can choose between getting just the transaction hashes or the full transaction set due to the boolean parameter supplied to it. In this case it shouldn't be necessary since uncle blocks are always w/o transactions.

ethers-providers/src/provider.rs
```diff
@@ -164,6 +164,40 @@ impl<P: JsonRpcClient> Provider<P> {
             }
         })
     }
+
+    async fn get_uncle_count_gen(&self, id: BlockId) -> Result<U256, ProviderError> {
```
gakonst, line 168:
```suggestion
    async fn get_uncle_count(&self, id: BlockId) -> Result<U256, ProviderError> {
```
Same comment as below

ethers-providers/src/provider.rs
```diff
@@ -227,6 +261,23 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
         self.get_block_gen(block_hash_or_number.into(), true).await
     }
 
+    /// Gets the block uncle count at `block_hash_or_number`
+    async fn get_uncle_count<T: Into<BlockId> + Send + Sync>(
+        &self,
+        block_hash_or_number: T,
+    ) -> Result<U256, Self::Error> {
+        self.get_uncle_count_gen(block_hash_or_number.into()).await
```
gakonst, line 269:
```suggestion
        self.get_uncle_count(block_hash_or_number.into()).await
```

ethers-providers/src/provider.rs
```diff
@@ -227,6 +261,23 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+    }
+
+    /// Gets the block uncle at `block_hash_or_number` and `idx`
+    async fn get_uncle<T: Into<BlockId> + Send + Sync>(
+        &self,
+        block_hash_or_number: T,
+        idx: U256,
+    ) -> Result<Option<Block<H256>>, ProviderError> {
+        self.get_uncle_gen(block_hash_or_number.into(), idx).await
```
gakonst, line 278:
```suggestion
        self.get_uncle(block_hash_or_number.into(), idx).await
```

gakonst (changes requested):

gakonst (approved):
LGTM - thank you.

PR #387, opened 2021-08-17
bug: log streams properly end when internal stream closes

prestwich:
Changes `<EventStream as Stream>::poll_next` to properly end the stream when the internal stream ends

Need to consider, how do we pass permanent/temporary failures from `Provider` -> `FilterWatcher` -> `EventStream`


gakonst:
Merged - opened a new issue to track the actual problem

PR #388, opened 2021-08-17
sign_transaction_sync

auterium:
## Motivation
`Wallet` struct implements the `Signer` trait to expose the `sign_transaction()` function which doesn't do any async calls, but it's still exposed through the async interface. This makes it complex to use in environments where there's no async runtime.

## Solution
Created a new public, sync function and moved the code from the async funtion to this one. Wasn't sure what name to use on the new function, but open to suggestions


prestwich (approved):
seems good to me

gakonst (approved):
LGTM

PR #395, opened 2021-08-19
fix: backwards compatibility for eth_feeHistory

roynalnaruto:
## Motivation

Geth v1.10.6 expects `blockCount` parameter in the RPC call to `eth_feeHistory` to be an unsigned integer, whereas since v1.10.7 it allows both unsigned integer as well as hex encoded uint. Hardhat allows only hex encoded uint.

The response format has also changed in Geth v1.10.7, whereby the field `oldestBlock` is now a hex-encoded uint.

## Solution

Deserialisation of the `FeeHistory` struct is made backwards-compatible with Geth v1.10.6 as well, and in case the call to Hardhat (or any other client only supporting hex-encoded form for `blockCount` ) fails, we fallback to the unsigned integer serialiastion while making the RPC call.

Note: This is for convenience for more users since a few networks (at this point in time Infura) are running Geth v1.10.6, while others may have upgraded. We don't want the `eth_feeHistory` to fail, since the EIP-1559 fee estimator makes use of this RPC method.

### DIFF
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -84,8 +84,8 @@ pub use pubsub::{PubsubClient, SubscriptionStream};
 use async_trait::async_trait;
 use auto_impl::auto_impl;
 use ethers_core::types::transaction::{eip2718::TypedTransaction, eip2930::AccessListWithGasUsed};
-use serde::{de::DeserializeOwned, Deserialize, Serialize};
-use std::{error::Error, fmt::Debug, future::Future, pin::Pin};
+use serde::{de::DeserializeOwned, Deserialize, Deserializer, Serialize};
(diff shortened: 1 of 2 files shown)


ethers-providers/src/lib.rs
```diff
@@ -706,7 +706,7 @@ pub trait Middleware: Sync + Send + Debug {
 
     async fn fee_history(
         &self,
-        block_count: u64,
+        block_count: U256,
```
gakonst, line 709: Nit: Make this `T: Into<U256>` to make this easier to call

gakonst (approved):
LGTM - one small nit if you want to make it slightly easier to use

ethers-providers/src/lib.rs
roynalnaruto: Definitely, makes sense!

Issue #391, opened 2021-08-19
WS Event Stream may hang if provider closes


gakonst:
**Version**

All versions w/ websocket support

**Description**

From Nathan Barnavon:

I'm trying to listen to events while connected to a provider. The problem I am running into is I cannot seem to detect when the connection with a provider has closed, and so I will wait forever. Any suggestions on how I should detect if the provider has closed the stream?

Example
```
async fn stream_event_loop(client: EvmClient, contract: AnchorContract<Provider<Ws>>) -> Result<(), anyhow::Error> {
    let current_block_number = client.provider.get_block_number().await.unwrap();
    let events_filter = contract.deposit_filter().from_block(current_block_number);
    let events = events_filter.subscribe().await?;
    let mut stream = events.with_meta();

    while let Some(v) = stream.try_next().map_err(Error::from).await? {
        let (_e, log) = v;
        println!("Got new Event at block {}", log.block_number);
    }
    Ok(())
}
```

**Proposed Solution**
1. Partially fixed in #387 so that the the outer stream ends when the internal stream ends
2. AFAIK there's no way to detect if the other side of the websocket connection is closed. The only way for us to do that would be to add a periodic Ping message and return an error in the internal filter watcher.

mattsse:
I guess this happens because we actually never do anything if the underlying websocket was closed in the spawned server
https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/transports/ws.rs#L213
since the websocket is unusable after that, should we close the and drop all channels, ergo closing alls subscriptions?

gakonst:
Fixed in #396.

vinceniko:
This issue still appears for me on master when subscribing to an abi autogenerated event filter.

### example
Using the abi auto generated filter doesn't work:
```{rust}
let filter = contract.event_filter();
let mut event_stream = filter.subscribe().await?;

let event: Event = event_stream
    .next()
    .await
    .ok_or("failed")??
```

Manually subscribing to the filter works correctly, i.e. an Err Result variant is returned when the ws connection is dropped:
```{rust}
let filter = contract.event_filter();
let mut event_stream = client.subscribe_logs(&filter.filter).await?;

let event: Event = filter.parse_log(
    event_stream
        .next()
        .await
        .ok_or("failed")?,
)?;
```

PR #396, opened 2021-08-19
feature: drop WS server on when WS connection closes

prestwich:
related to #391 

- causes the `WsServer` to drop if the `ws` stream ends
- causes `SubscriptionStream` to end if the internal channel ends

includes refactors to
- handling incoming message types (binary, close, etc)
- handling incoming text (`Incoming` enum)
- handling consumer instruction (`Instruction` enum)
- naming clarity
- enhance `SubscriptionStream` docs

### DIFF
--- a/ethers-providers/src/pubsub.rs
+++ b/ethers-providers/src/pubsub.rs
@@ -44,7 +44,13 @@ where
     P: PubsubClient,
     R: DeserializeOwned,
 {
-    /// Creates a new subscription stream for the provided subscription id
+    /// Creates a new subscription stream for the provided subscription id.
+    ///
+    /// ### Note
+    /// Most providers treat `SubscriptionStream` IDs as global singletons.
+    /// Instanitating this directly with a known ID will likely cause any
+    /// existing streams with that ID to end. To avoid this, start a new stream
+    /// using [`Provider::subscribe`] instead of `SubscriptionStream::new`.
     pub fn new(id: U256, provider: &'a Provider<P>) -> Result<Self, P::Error> {
         // Call the underlying PubsubClient's subscribe
         let rx = provider.as_ref().subscribe(id)?;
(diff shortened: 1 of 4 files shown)


ethers-providers/src/pubsub.rs
```diff
@@ -44,7 +44,13 @@ where
     P: PubsubClient,
     R: DeserializeOwned,
 {
-    /// Creates a new subscription stream for the provided subscription id
+    /// Creates a new subscription stream for the provided subscription id.
+    ///
+    /// ### Note
+    /// Most providers treat `SubscriptionStream` IDs as global singletons.
+    /// Instanitating this directly with a known ID will likely cause any
```
gakonst, line 51:
```suggestion
    /// Instantiating this directly with a known ID will likely cause any
```

ethers-providers/src/transports/ws.rs
```diff
@@ -189,101 +205,135 @@ where
...
+    }
+
+    async fn handle_text(&mut self, inner: String) -> Result<(), ClientError> {
+        let message = serde_json::from_str::<Incoming>(&inner);
+        if message.is_err() {
+            // Ignore malformatted server responses
+            return Ok(());
+        }
+
+        match message.unwrap() {
```
gakonst, line 286: You can do `match message { Ok(Incoming::Response(resp) => { ...}, Ok(Incoming::Notification(notification) }` etc. if you want to avoid the double nesting of `match` and not use `.unwrap()`

gakonst (approved):
Largely LGTM - thank you. A nit would be if you wanted to split the PR in 2 components, first the bugfix then the refactor (or the opposite), for clarity. No worries if not.

ethers-providers/src/transports/ws.rs
prestwich: dbf6de0

ethers-providers/src/pubsub.rs
prestwich: bc9783a

prestwich:
@gakonst I split it. this PR is the bugfix, and #396 is the refactor

PR #399, opened 2021-08-20
fix: drop WS server if work complete

mattsse:
## Motivation
This fixes another edge case that leads to active WS servers with not possibility of future work.
If all instruction senders are dropped there are no more subscriptions or pending requests, then this leaves the Ws server with nothing todo and since the receiver can't receive any more instructions it's effectively useless, running in the background with, still connected to the ws server and occasionally receiving/sending pings/pongs.
Also Subscriptions with a closed channel also remained active.

## Solution
* Remove a subscription from the `subscriptions` if its channel was closed
* break the tick loop so the server gets dropped if the instructions receiver channel is closed and no more subscriptions are active or requests are pending.
*


mattsse:
Some parts work, however anything related to deploying and interacting hangs, could be due to misconfig on my end 🤷‍♂️ 

To test the minimal wasm test:
1. start ganache in separate terminal : `ganache-cli`
2. run wasm-pack (installation reuqired `cargo install wasm-pack`)
```
cd ethers-providers
wasm-pack test --chrome
```
go to http://127.0.0.1:8000/ and check browser console
or run in headless (only firefox works)
`wasm-pack test --headless --firefox`

mattsse:
# Update

added an `is_done` check in the tick loop instead of checking this abort condition in the tick function

gakonst (approved):
LGTM - nice work.

PR #400, opened 2021-08-23
feat: add basic policy middleware

mattsse:
## Motivation
Support a way to enforce certain rules for transactions

## Solution
Add a new `PolicyMiddleware` type which wrapps a middleware and includes an instance of the new `Policy` trait, which is a single function to ensure a transaction is in compliance with this policy.

### Further work
common use cases could be to look up the receiver in a database and check if its address is whitelisted. Not sure yet what the most flexible implementation would be since a DB adapter is database specific and also how to check for a whitelisted address may be database model specific. So we could abstract basic use cases via some kind of `PolicyBackend` trait or smth which implements common functions for checking `is_whitelist(&self, recv: Address)` and also for checking the calldata maybe. We could also add a default persistent implementation of that (postgres for example) and a `HashSet` based one and feature gate this.

### DIFF
--- a/ethers-middleware/src/lib.rs
+++ b/ethers-middleware/src/lib.rs
@@ -78,3 +78,8 @@ pub mod transformer;
 /// instead of using eth_sendTransaction and eth_sign
 pub mod signer;
 pub use signer::SignerMiddleware;
+
+/// The [Policy](crate::PolicyMiddleware) is used to ensure transactions comply with the rules
+/// configured in the `PolicyMiddleware` before sending them.
+pub mod policy;
+pub use policy::PolicyMiddleware;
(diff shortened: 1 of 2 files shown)


gakonst (approved):
Excellent, this interface looks great: `async fn ensure_can_send(&self, tx: TypedTransaction) -> Result<TypedTransaction, Self::Error>;`

I don't think we really need the policy backend trait given the quirks of each individual platform. Let's start with a simple in-memory one, and we can start adding databases afterwards.

Simple functionalities:
* `to` can only be in whitelist
* `from` can only be in whitelist
* `value` cannot be over X ETH
* `data` can only follow certain patterns (e.g. match first 8 bytes to limit function selector, or even more granularly limited the args based on the function selector)

PR #401, opened 2021-08-23
fix: use once_cell with feature=std in abigen

nhynes:
This PR fixes the `once_cell` dependency of `ethers-contract-abigen` which has a dependency on `once_cell::sync`, but doesn't pull in this mod due to the `std` feature being turned off. If this worked before, it's because all of the tests are run using Cargo with `resolver = 1`, which compiles one version of a dependency with all features enabled by all peer dependencies.


gakonst (approved):
LGTM - nice catch! I wonder if that was what was causing the failure in CI for abigen recently...

PR #390, opened 2021-08-18
feat: wasm support

mattsse:
## Motivation
Add wasm support
- Close #96 

## Solution
- [x] make everything compile with `--target wasm32-unknown-unknown`. the biggest challenge is probably disabling transports.
- [x] implement a wasm compatible WS provider using `web_sys::WebSocket`
- [x] make sure all abi generated stuff is wasm compatible.

Afterwards we can look at how to shrink it etc.

PR #403, opened 2021-08-23
feat: add wasm http provider support

mattsse:
## Motivation
Followup from #390

## Solution
* enable Http Provider in wasm. 
* Http support in `contract-derive-abigen` remains disabled, because wasm http requires async. however this is not a limitation to the actual `abigen!` macro which doesn't get executed with wasm anyways.
* fixed the double execute issue with webpack. Since I couldn't immediately find how to manually bundle wasm-pack output I kept it for now. But yarn serve doesn't throw errors anymore


examples/ethers-wasm/src/lib.rs
```diff
@@ -1,90 +1,90 @@
...
 
-/// key[0] of ganache with custom seed `ethers-wasm-seed`
-pub const KEY: &str = "817169e55f14ede54f4fd6a4f2ab4209db14aeeb1b9972b3b28f1560af0a061a";
+/// some keys of ganache with custom seed `ethers-wasm-seed`
+pub const KEYS: [&str; 4] = [
+    "817169e55f14ede54f4fd6a4f2ab4209db14aeeb1b9972b3b28f1560af0a061a",
+    "375715b8ced8bd9b7386ba5dc72efa518aa4379d6a4676d3e26d8f5ff5e7469c",
+    "de7c5d8e884fbe9f0915703ff2c123f4cda56f148eb22ca46d47392acf52bcec",
+    "0bd6bf22f84f96b39258a46ac2a7c482d0b8e1c5f8af0c07fa304a8d875158ec",
+];
```
gakonst, line 39: FWIW you can also use the [`Mnemonic`](https://github.com/gakonst/ethers-rs/blob/master/ethers-signers/src/wallet/mnemonic.rs#L78-L80) if you want to make this a bit more robust

gakonst (approved):
LGTM - one minor nit. Feel free to merge!

examples/ethers-wasm/src/lib.rs
mattsse: used mnemonic instead.

gakonst (approved):

PR #409, opened 2021-08-27
feat: add quorum provider

mattsse:
## Motivation
Support multiple provider endpoints

## Solution
Adds the `QuorumProvider` type that can run internal multiple providers internally and only returns a response if a quorum was reached. The internal providers can be weighted.
This required two Wrapper traits for `JsonRpcClient` `PubsubClient` to get rid of the generic types.
How a quorum is reached can be configured with `Quorum`, default is `Quorum::Majority` which reaches a quorum if >50% of the weighted providers returned the same value

### DIFF
--- a/ethers-providers/src/transports/quorum.rs
+++ b/ethers-providers/src/transports/quorum.rs
@@ -166,7 +166,7 @@ impl<T: JsonRpcClientWrapper> QuorumProvider<T> {
     /// Normalizes the request payload depending on the call
     async fn normalize_request(&self, method: &str, params: &mut Value) {
         match method {
-            "eth_Call"
+            "eth_call"
             | "eth_createAccessList"
             | "eth_getStorageAt"
             | "eth_getCode"
--- a/ethers-providers/src/transports/quorum.rs
+++ b/ethers-providers/src/transports/quorum.rs
@@ -1,0 +1,587 @@
+use std::{
+    fmt,
+    fmt::Debug,
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use crate::{provider::ProviderError, JsonRpcClient, PubsubClient};
+use async_trait::async_trait;
+use ethers_core::types::U256;
+use futures_core::Stream;
+use futures_util::{FutureExt, StreamExt};
+use serde::{de::DeserializeOwned, Serialize};
+use serde_json::Value;
+use thiserror::Error;
+
+/// A provider that bundles multiple providers and only returns a value to the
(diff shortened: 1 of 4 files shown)


ethers-providers/src/transports/quorum.rs
```diff
@@ -0,0 +1,640 @@
...
+    fn weight<T>(self, providers: &[WeightedProvider<T>]) -> u64 {
+        match self {
+            Quorum::All => providers.iter().map(|p| p.weight).sum::<u64>(),
+            Quorum::Majority => {
+                let total = providers.iter().map(|p| p.weight).sum::<u64>();
+                let rem = total % 2;
+                total / 2 + rem
+            }
+            Quorum::Percentage(p) => {
+                providers.iter().map(|p| p.weight).sum::<u64>() * (p as u64) / 100
```
gakonst, line 226: Maybe add a `if p > 100 { panic!() }` here?

ethers-providers/src/transports/quorum.rs
```diff
@@ -0,0 +1,640 @@
...
+        }
+    }
+}
+
+impl<'a, T> Future for QuorumRequest<'a, T> {
+    type Output = Result<Value, QuorumError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.get_mut();
+        for n in (0..this.requests.len()).rev() {
```
gakonst, line 281: Why are we iterating in reverse here?

ethers-providers/src/transports/quorum.rs
```diff
@@ -0,0 +1,640 @@
...
+}
+
+#[cfg(test)]
+#[cfg(not(target_arch = "wasm32"))]
+mod tests {
+    use super::{Quorum, QuorumProvider, WeightedProvider};
+    use crate::{Middleware, MockProvider, Provider};
+    use ethers_core::types::U64;
+
+    async fn test_quorum(q: Quorum) {
```
gakonst, line 569: Nice usage of the mock provider

gakonst (approved):
In general LGTM. I think it'd be useful if we also had an integration test in the middleware crate showing that Streams / Subscriptions work when paired with 1 Infura and 1 Alchemy node

ethers-providers/src/transports/quorum.rs
mattsse: because how swap_remove works, which replaces the removed element with the last element. but this also means that we start at `0` we would miss the value the last index, therefor we need to start at the last index.

PR #412, opened 2021-08-28
explore provider api

mattsse:
## Motivation
Get rid of all associated types of the `Middleware`

## Solution
* Merge `JsonRpcClient` and `Middleware` into a single trait `JsonRpcProvider`.
* use `eyre::Result` instead of `Middleware::Error`

Provider nesting would work as follows

```rust
pub struct SignerMiddleware<Signer, Provider> {
    pub(crate) inner: Provider,
    pub(crate) signer: Signer,
    pub(crate) address: Address,
}

#[async_trait]
impl<Signer: Send + Sync, Provider: JsonRpcProvider> JsonRpcProvider
    for SignerMiddleware<Signer, Provider>
{
    async fn request<T, R>(&self, method: &str, params: T) -> Result<R>
    where
        Self: Sized,
        T: Debug + Serialize + Send + Sync,
        R: Serialize + DeserializeOwned,
    {
        self.inner.request(method, params).await
    }

    fn default_sender(&self) -> Option<Address> {
        Some(self.address)
    }
}
```

we could keep the `JsonRpcClient` (as `Transport` ?) for Http, Ws clients, but in order to keep all default `JsonRpcProvider` function implementation we'd need to restrict `JsonRpcProvider: Transport`, so in addition to `impl JsonRpcProvider` the Providers would need to impl `Transport` as well, which is probably fine but extra work

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -969,6 +969,7 @@ dependencies = [
  "bytes",
  "ethers",
  "ethers-core",
+ "eyre",
  "futures-channel",
  "futures-core",
  "futures-timer",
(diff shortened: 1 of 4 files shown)


mattsse:
closing this as this is totally not viable this way due to rust's object safety rules.

PR #415, opened 2021-08-29
Correctly encode non-legacy mempool transactions

onbjerg:
## Motivation

If a non-legacy transaction was fetched from the mempool and then RLP-encoded for inclusion in e.g. a flashbots bundle, then the bundle would fail because the encoding only worked for legacy transactions. See https://github.com/onbjerg/ethers-flashbots/issues/11#issuecomment-907672337

## Solution

RLP encode `Transaction` response types based on their `transaction_type`. I'm not super confident about this piece of code since it is a lot of duplicate logic from other files, but I'm not super well known in the codebase yet so I'm not sure what the best way is to improve it. Please let me know if you have any suggestions!

### DIFF
--- a/ethers-core/src/types/transaction/response.rs
+++ b/ethers-core/src/types/transaction/response.rs
@@ -1,5 +1,5 @@
 //! Transaction types
-use super::{eip2930::AccessList, rlp_opt, NUM_TX_FIELDS};
+use super::{eip2930::AccessList, rlp_opt};
 use crate::{
     types::{Address, Bloom, Bytes, Log, H256, U256, U64},
     utils::keccak256,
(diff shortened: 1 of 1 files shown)


gakonst (approved):

avocadochicken:
Not working yet. Please take code from here:
https://github.com/onbjerg/ethers-flashbots/issues/11#issuecomment-908130611

@onbjerg might do a PR soon.

onbjerg:
> Not working yet. Please take code from here:
> [onbjerg/ethers-flashbots#11 (comment)](https://github.com/onbjerg/ethers-flashbots/issues/11#issuecomment-908130611)
> 
> @onbjerg might do a PR soon.

Yeah, I have some tests locally and some fixes too, but I want to add one more before pushing. Thanks for the PR on my fork, I'll take a look today :)

onbjerg:
Just asked in the Flashbots Discord, it seems to be the return value of the called function, so I'll have to adjust that.

onbjerg:
Should be good to go @gakonst

ethers-core/src/types/transaction/response.rs
```diff
@@ -141,22 +141,78 @@ impl Transaction {
...
+                encoded.extend_from_slice(&[0x1]);
+                encoded.extend_from_slice(rlp_bytes.as_ref());
+                encoded.into()
+            }
+            Some(x) if x == U64::from(2) => {
+                encoded.extend_from_slice(&[0x2]);
+                encoded.extend_from_slice(rlp_bytes.as_ref());
+                encoded.into()
+            }
+            _ => rlp_bytes,
```
gakonst, line 214: WDYT about matching `Some(x) if x == U64::from(0)` and encoding with the 0x0 prefix, and default to `rlp_bytes` only when `self.transaction_type` is not present OR if the `legacy` feature is turned on?

gakonst (approved):

ethers-core/src/types/transaction/response.rs
onbjerg: AFAIK `0x0` is reserved for new tx types, not legacy: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2718.md#backwards-compatibility

gakonst: NVM. You are right, forgot that we skip it for legacy txs: https://github.com/gakonst/ethers-rs/blob/404e9c24163c35ff4055bfa3ce3f7107d163e7c7/ethers-core/src/types/transaction/eip2718.rs#L153-L155

PR #427, opened 2021-09-01
fix: Enable solc optimization

wolflo:
## Motivation

Currently, the solc bindings ignore the `optimizer` parameter.

## Solution

Add the optimization arguments to the executed command.

This also adds some options around using the optimizer, some of which may be a little awakward as implemented. Ideally, in addition to enabling the optimizer, users should be able to:
1. Compile without the optimizer
2. Set the optimizer runs
3. Set optimizer runs to 0 (i.e. we can't use 0 as a hack for turning the optimizer off)

As implemented, the current examples will work as expected (with optimization now).

```
Solc::new("./contracts/*").optimizer(200).build()?;
```

However, the optimizer no longer defaults to the optimizer ON with 200 runs (which is what the documentation used to suggest). Also, you now need to know the runs parameter you want to use in order to enable optimization.

Alternatively, we could add a `.optimize(bool)` method, with the downside just being verbosity. If a breaking change was acceptable, `optimizer()` could take an `Option<usize>`.

## PR Checklist

- [ ] Added Tests
- [X] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/utils/solc.rs
+++ b/ethers-core/src/utils/solc.rs
@@ -35,7 +35,7 @@ pub struct CompiledContract {
 /// Assumes that `solc` is installed and available in the caller's $PATH. Any calls
 /// will **panic** otherwise.
 ///
-/// By default, it uses Istanbul as the EVM version.
+/// By default, it uses 200 optimizer runs and Istanbul as the EVM version
 ///
 /// # Examples
 ///
@@ -45,7 +45,7 @@ pub struct CompiledContract {
 /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
 /// // Give it a glob
 /// let contracts = Solc::new("./contracts/*")
-///     .optimizer(200)
+///     .optimizer(Some(200))
 ///     .build()?;
 ///
 /// // this will return None if the specified contract did not exist in the compiled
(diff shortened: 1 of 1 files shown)


gakonst:
Thanks for the PR! I'd be supportive of as many breaking changes around that as you want. So far so good - do you think you'd want to merge as-is, or do you want to add additional stuff?

wolflo:
My preference would probably be to have `optimize()` take an `Option<usize>`, with `None` meaning no optimization. That would give you the flexibility to have the common default (`--optimize --optimize-runs 200`) while still being able to set on / off / runs without chaining another method. 

This would break things that previously used `.optimize(200)`, but would restore the "expected" default of 200 runs.

What do you think about also taking arbitrary string args the way ganache does?
https://github.com/gakonst/ethers-rs/blob/584b683c1240fed5ab6d34fb9ba0d6b6aa20ad6a/ethers-core/src/utils/ganache.rs#L122-L126

gakonst:
Both Option<usize> and arbitrary string args SGTM!

wolflo:
The breaking change here is that `optimizer()` now takes an `Option<usize>` instead of `usize`, and `optimizer(None)` turns the optimizer off.

The other addition is `.arg()` and `.args()` which take raw argument strings (or an iterator of raw argument strings) and passes them to the solc command. This is going to have some rough edges because of the strict parsing of the output. Anything that changes the structure of the output by adding an additional field to the combined json (e.g. `--asm`) will result in
```
Error: trailing characters at line 3 column 1
```

Probably the output parsing could be made more general, but at some point it's easier to just build the command and handle the output yourself.

```rust
// No optimization
let contracts = Solc::new("./contracts/*")
    .optimizer(None)
    .build()?;

// Some(200) is default, optimizer on with 200 runs
// .arg() allows passing arbitrary args to solc command
let optimized_contracts = Solc::new("./contracts/*")
    .optimizer(Some(200))
    .arg("--metadata-hash=none")
    .build()?;
```

### Tests

This could use some, but I'm not sure the most robust way to test Solc given that it's all reliant on the locally installed version.

PR #451, opened 2021-09-14
test: fix duplicate tx flakes by rotating through a list of wallets

gakonst:
Is this the day our tests stop being flaky? Maybe? We now rotate over private keys associated with a mnemonic, instead of having a bunch of keys lying around and hoping we don't duplicate them somewhere & introduce race conditions.

I've pre-funded the wallets generated as documented [here](https://github.com/gakonst/ethers-rs/compare/fix/flaky-tests?expand=1#diff-be734beb2ff927faeaf4f9733808604d8fdf182a633d164c6168247ac475abddR117-R120).

### DIFF
--- a/ethers-middleware/tests/signer.rs
+++ b/ethers-middleware/tests/signer.rs
@@ -1,9 +1,22 @@
-use ethers_providers::{Http, Middleware, Provider};
+use ethers_providers::{Http, JsonRpcClient, Middleware, Provider};
 
-use ethers_core::types::TransactionRequest;
+use ethers_core::{
+    types::{BlockNumber, TransactionRequest},
+    utils::{parse_ether, parse_units},
+};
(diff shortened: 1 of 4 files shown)


ethers-middleware/tests/signer.rs
```diff
@@ -300,3 +309,63 @@ async fn deploy_and_call_contract() {
...
+        let addr = provider.address();
+
+        let mut nonce = provider.get_transaction_count(addr, None).await.unwrap();
+        let mut pending_txs = Vec::new();
+        for addr in addrs {
+            println!("Funding wallet {:?}", addr);
+            let tx = TransactionRequest::new()
+                .nonce(nonce)
+                .to(addr)
+                // 0.1 eth per wallet
```
gakonst, line 341:
```suggestion
                // 1 eth per wallet
```

PR #459, opened 2021-09-17
Add getProof to provider

drewstone:
## Motivation
Adding `eth_getProof` RPC support so that we can serve storage proofs for various applications.

s/o @nepoche for contribution as well.

## Solution
Adding the RPC call, methods, and respective types. Adding a test to ensure it works.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


ethers-providers/src/provider.rs
```diff
@@ -804,6 +827,89 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
             )
             .await)
     }
+
+    async fn fill_transaction(
```
drewstone, line 831: not really sure how this ended up here, I didn't touch it.

PR #465, opened 2021-09-23
fix: use Transaction instead of TransactionRequest for txpool

x04:
## Motivation

I was attempting to monitor mempool for notifications of certain transactions becoming pending and I noticed that I could not get access to certain fields, mainly the transaction hash, despite them being shown in geth documentation. I noticed that for some reason the TransactionRequest struct is used rather than the full Transaction struct, which I believe isn't really ideal because there is more information that we receive that TransactionRequest does not account for but Transaction does.

## Solution

Use full Transaction struct rather than TransactionRequest for txpool_content call.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog (looking for input as to how this should be done, what version should it be bumped to etc)


x04:
Oh it seems I overlooked a test however I'm unsure what would be the best way to go about rewriting it to work for this scenario, I would think it would be best to get the expected transaction hash and compared to the hash of the transaction in the mempool and that should suffice? Or is there a better way possibly I'm just not thinking of?

gakonst:
Hey @x04 apologies for the delay. I think comparing the tx hash in the pool against the one from get_transaction makes sense.

gakonst:
Closing as stale, lmk if any interest in picking #576 back up

Issue #446, opened 2021-09-12
Human Readable ABI improvements


gakonst:
Currently, the human readable abi does not support "shorthand" declarations e.g. `greeting()(string)` or `hello(uint256)()`. Instead it requires both the `function` and the `returns` keyword. It'd be nice if our parser could support all styles.

gakonst:
Fixed in #472

PR #472, opened 2021-09-27
feat(core, abi): support shorthand function declarations

shekhirin:
## Motivation

See https://github.com/gakonst/ethers-rs/issues/446

## Solution

Support shorthand function declarations

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1478,9 +1478,9 @@ dependencies = [
 
 [[package]]
 name = "instant"
-version = "0.1.10"
+version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


gakonst:
This is great. Thank you. Paging @mattsse in case he's available to review this, since he wrote the parser, otherwise I'll try to take a closer look tomorrow.

shekhirin:
Another thing that may worth the issue: redesign human readable ABI parsing, so it doesn't split the strings 9999 times but iterate the tokens instead 😄

mattsse:
Definitely, when I first wrote this I used the internal ethabi parser as the basis for that.
Tokenization would be much nicer, but at least the human readable abi does only add to compile-time in `abigen!` 
+1 for follow up issue, parts of [solang](https://github.com/hyperledger-labs/solang/blob/main/src/parser/lexer.rs) could be useful for that,

mattsse (approved):
awesome

PR #477, opened 2021-09-29
fix(human-readable-abi): allow functions without output args

gakonst:
Allows functions wihtout output args in the human readable ABI, previously it'd throw an error. cc @mattsse @ shekhirin


mattsse (approved):
good catch!

PR #481, opened 2021-10-01
derive-eip712: initial implementation of eip712 derive macro

Ryanmtate:
This commit provides an initial implementation for a derive macro
to encode typed data according to EIP-712, https://eips.ethereum.org/EIPS/eip-712

Additionally, this commit introduces a new signer trait method:

```rust
    async fn sign_typed_data<T: Eip712 + Send + Sync>(
        &self,
        payload: &T,
    ) -> Result<Option<Signature>, Self::Error>;
```

And implements the new method for each of the signers (wallet, ledger,
aws).

At the moment, derive does not recurse the primary type to find nested
Eip712 structs.

Last, this commit does not test against ethereum abi.encodePacked()
values and is therefore considered unreliable and is NOT production
ready. This commit is considered WIP and should not be released into
production code until tested.

## PR Checklist

- [ ] Added Tests
- [X ] Added Documentation
- [ ] Updated the changelog


derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+    parsed_fields
+}
+
+// Convert hash map of field names and types into a type hash corresponding to enc types;
+fn make_type_hash(primary_type: String, fields: &HashMap<String, String>) -> String {
+    let parameters = fields
+        .iter()
+        .map(|(k, v)| format!("{} {}", v, k))
+        .collect::<Vec<String>>()
+        .join(",");
```
Ryanmtate, line 262: Note to myself, these parameters should be sorted alphabetically.

derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+        .path
+        .segments
+        .first()
+        .expect("field must have a type");
+
+    let mut field_type = segment.ident.to_string();
+
+    if let syn::PathArguments::AngleBracketed(arguments) = &segment.arguments {
+        field_type.push('<');
+        for arg in &arguments.args {
```
Ryanmtate, line 204: Currently, this works well if there is only one argument within the angle bracket. Should likely throw an error if there are more than one arguments. Trying to think of a use case where multiple arguments (e.g. HashMap<String, String>) would be mapped to an ethereum type.

derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+    // a work around for passing in the struct fields;
+    let fields: String = serde_json::to_string(&parsed_fields)
+        .expect("failed to serialize parsed fields into JSON string");
+
+    // Compute the type hash for the derived struct using the parsed fields from above;
+    let type_hash = make_type_hash(primary_type.clone().to_string(), &parsed_fields);
+
+    let implementation = quote! {
+        #[derive(Debug, thiserror::Error)]
+        pub enum Eip712Error {
```
Ryanmtate, line 295: Move to `ethers_core::transaction::eip712`. Multiple derivations while erroneously call multiple declarations.

derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+            #[error("Failed to serialize serde JSON object")]
+            SerdeJsonError(#[from] serde_json::Error),
+            #[error("Failed to decode hex value")]
+            FromHexError(#[from] hex::FromHexError),
+            #[error("Failed to make struct hash from values")]
+            FailedToEncodeStruct
+        }
+
+        fn make_struct_hash<T: std::fmt::Debug + serde::Serialize>(
```
Ryanmtate, line 304: move contents of this function into the `encode_eip712()` method.

Cargo.toml
```diff
@@ -19,6 +19,7 @@ members = [
     "ethers-signers",
     "ethers-core",
     "ethers-middleware",
+    "derive-eip712"
```
Ryanmtate, line 22: move this crate under ethers_core

derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+
+/// This method provides the hex encoded domain type hash for EIP712Domain type;
+/// This is used by all Eip712 structs.
+fn eip712_domain_type_hash() -> [u8; 32] {
+    keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)")
+}
+
+/// Eip712 Domain attributes used in determining the domain separator;
+#[derive(Debug, Default)]
+struct Eip712Domain {
```
Ryanmtate, line 78: Move types to ethers_core::transaction::eip712 so they are available outside of the macro crate.

ethers-core/src/types/transaction/eip712.rs
```diff
@@ -0,0 +1,276 @@
...
+        };
+
+        domain
+    }
+}
+
+// Convert rust types to enc types. This is used in determining the type hash;
+// NOTE: this is not an exhaustive list, and there may already be an existing mapping
+// in another library.
+pub fn parse_field_type(field_type: String) -> String {
```
Ryanmtate, line 168: use ethabi::token::Token or ParamType instead of String

https://docs.rs/ethabi/15.0.0/ethabi/token/enum.Token.html

ethers-core/ethers-derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+                            ethers_core::types::U256::from(&hex::decode(#type_hash)?[..]),
+                        )];
+
+                        for key in keys {
+                            if let Some(v) = obj.get(&key) {
+                                if let Some(ty) = fields.get(&key) {
+                                    if let serde_json::Value::String(value) = v{
+                                        if let serde_json::Value::String(field_type) = ty {
+                                            // convert encoded type;
+                                            let item = match field_type.as_ref() {
```
Ryanmtate, line 124: Look into using ethers_core::abi::Token::TryFrom<T>

ethers-core/src/types/transaction/eip712.rs
```diff
@@ -0,0 +1,276 @@
...
+        "H128" => "bytes16",
+        "H160" => "address",
+        "H256" => "bytes32",
+        "String" => "string",
+        "Boolean" => "boolean",
+        "Bytes" => "bytes",
+        "Vec<u8>" => "bytes",
+        "Vec<U128>" => "uint128[]",
+        "Vec<U256>" => "uint256[]",
+        "Vec<H128?" => "bytes16[]",
```
gakonst, line 181: question mark is a typo here right?

ethers-core/src/types/transaction/eip712.rs
```diff
@@ -0,0 +1,276 @@
...
+        "Vec<H160>" => "address[]",
+        "Vec<H256?" => "bytes32[]",
```
gakonst, line 183: question mark is a typo here right?

ethers-signers/src/lib.rs
```diff
@@ -93,6 +95,21 @@ pub trait Signer: std::fmt::Debug + Send + Sync {
...
+    /// Encodes and signs the typed data according EIP-712
+    /// payload must implement Eip712 trait.
+    ///
+    /// NOTE: Returning is Option<Signature> instead of Signature
+    /// due to use of std::convert::Infallible as an error type
+    /// in the Wallet implementation, which conflicts with
+    /// the error types used within the method, which cannot be mapped
+    /// to Infallible type. To avoid breaking changes and changing the error
+    /// type, this method returns `None` if there is an error that restricts
+    /// the payload from being encoded, and signed.
```
gakonst, line 107: Mhm this is a bit weird, we should be able to do it normally

ethers-signers/src/wallet/mod.rs
```diff
@@ -83,6 +87,23 @@ impl<D: Sync + Send + DigestSigner<Sha256Proxy, RecoverableSignature>> Signer fo
         Ok(self.sign_transaction_sync(tx))
     }
 
+    async fn sign_typed_data<T: Eip712 + Send + Sync>(
+        &self,
+        payload: &T,
+    ) -> Result<Option<Signature>, Self::Error> {
+        if let Ok(encoded) = payload.encode_eip712() {
```
gakonst, line 94: Why can you not do `payload.encode_eip712()?` etc for the other methods?

ethers-core/ethers-derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+                #type_hash.to_string()
+            }
+
+            fn domain_separator() -> String {
+                #domain_separator.to_string()
+            }
+
+            fn encode_eip712(&self) -> Result<String, Self::Error> {
+                // Ok(make_struct_hash(self, #domain_separator, #type_hash, #fields)?.to_string())
+                let json: serde_json::Value = serde_json::from_str(#fields)?;
```
mattsse, line 101: this is a clever hack lol
this looks rather similar to the `EthEvent` implementation.
I might be wrong but all this does is essentially encode the whole struct (its fields) into a `Vec<Token>` right?
can't we mandate that this struct implements `Tokenizable` which provides `Tokenizable::into_token() ->Token` which is a tuple for structs. This is all implemented automatically with `EthAbiType` macro, examples https://github.com/gakonst/ethers-rs/blob/master/ethers-contract/tests/common/derive.rs
This will require a clone however, we could get rid of this with some effort but I guess cloning would be negligeble.

mattsse (commented):
some comments/questions/recommendations

ethers-core/src/types/transaction/eip712.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// for which the user is targeting. In the derive macro, these attributes
+    /// are passed in as arguments to the macro. When manually deriving, the user
+    /// will need to know the name of the domain, version of the contract, chain ID of
+    /// where the contract lives and the address of the verifying contract.
+    fn domain_separator() -> String;
+
+    /// This method is used for calculating the hash of the type signature of the
+    /// struct. The field types of the struct must map to primitive
+    /// ethereum types or custom types defined in the contract.
+    fn type_hash() -> String;
```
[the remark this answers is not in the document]
Ryanmtate: The original error I was getting was due to trying to pass the compile time [u8; 32] into the method using #type_hash. However, this does not work due to ToTokens not being implemented for [u8; 32].

That said, I did manage to update the trait to return `Result<[u8; 32], Eip712Error>` rather than `String` simply by performing the hex encoding -> String at compile time and decoding the `'static &str` -> `[u8;32]` within the methods. It's a hack, but it allows us to return the desired `[u8; 32]` type, avoiding the user having to worry about hex encoding/decoding.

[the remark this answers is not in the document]
Ryanmtate: I'm also not opposed to returning Bytes.

ethers-signers/src/wallet/mod.rs
Ryanmtate: Ideally, we could call `payload.encode_eip712()?`, but I am still trying to figure out how to map the error to the `std::convert::Infallible` type. Would changing the trait Error type for the wallet signer cause any downstream issues? It seems like there could be cases where an error may happen, which doesn't seem infallible to me.

ethers-core/src/types/transaction/eip712.rs
```diff
@@ -0,0 +1,276 @@
...
+            Token::Uint(U256::from(eip712_domain_type_hash())),
+            Token::Uint(U256::from(keccak256(&self.name))),
+            Token::Uint(U256::from(keccak256(&self.version))),
+            Token::Uint(self.chain_id),
+            Token::Address(self.verifying_contract),
+        ])))
+    }
+}
+
+// Parse the AST of the struct to determine the domain attributes
```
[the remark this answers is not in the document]
Ryanmtate: Definitely like your pattern better; will implement the same for the `Eip712Domain` struct.

ethers-core/src/types/transaction/eip712.rs
Ryanmtate: Thanks @mattsse! Will look into using this!

ethers-core/src/types/transaction/eip712.rs
```diff
@@ -0,0 +1,276 @@
...
+/// implement the trait for a given struct. See documentation
+/// for `derive_eip712` for more information and example usage.
+///
+/// For those who wish to manually implement this trait, see:
+/// https://eips.ethereum.org/EIPS/eip-712
+///
+/// Any rust struct implementing Eip712 must also have a corresponding
+/// struct in the verifying ethereum contract that matches its signature.
+///
+/// NOTE: Due to limitations of the derive macro not supporting return types of
```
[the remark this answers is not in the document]
Ryanmtate: Problem was not so much returning bytes, but my misunderstanding of passing compile time values into the derived trait methods.

ethers-core/src/types/transaction/eip712.rs
```diff
@@ -0,0 +1,276 @@
...
+        let field_type = parse_field(f);
+
+        parsed_fields.insert(field_name, field_type);
+    });
+
+    parsed_fields
+}
+
+/// Convert hash map of field names and types into a type hash corresponding to enc types;
+pub fn make_type_hash(primary_type: String, fields: &HashMap<String, String>) -> String {
```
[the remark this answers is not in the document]
Ryanmtate: I agree ordering is something to watch out for. This method currently sorts the HashMap keys (field names) alphabetically, but the rule is that whatever the ordering, it needs to match the ordering of the type hash in the solidity contract.

Sorting the fields alphabetically makes sense, but I could see where others might mimic the ordering of the struct fields in the rust and solidity types, not necessarily in alphabetical order.

Accessing the fields by name is helpful when encoding the struct for the field type, but we might be able to get around needing a HashMap and avoiding serializing into / from JSON.

ethers-core/ethers-derive-eip712/src/lib.rs
Ryanmtate: Oh nice! `EthAbiType` looks great! Will look for solutions to avoid needing to serialize the hash map here.

derive-eip712/src/lib.rs
Ryanmtate: No longer using a hashmap, fields are sorted based on their position in the struct.

derive-eip712/src/lib.rs
Ryanmtate: No longer using this method; deprecating in favor of `find_parameter_type`

derive-eip712/src/lib.rs
Ryanmtate: moved

derive-eip712/src/lib.rs
Ryanmtate: moved

Cargo.toml
Ryanmtate: moved.

derive-eip712/src/lib.rs
Ryanmtate: moved.

ethers-core/src/types/transaction/eip712.rs
Ryanmtate: Thanks again for this! Ended up copying this method. Deprecated the old method.

ethers-core/ethers-derive-eip712/src/lib.rs
Ryanmtate: Using `EthAbiType` macro to derive types.

ethers-core/ethers-derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+                    }
+                }
+
+                // Reached Error:
+                Err(ethers_core::types::transaction::eip712::Eip712Error::FailedToEncodeStruct)
+
+            }
+
+            fn eip712_domain_type_hash() -> String {
+                #domain_type_hash.to_string()
```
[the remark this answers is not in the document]
Ryanmtate: Returns `[u8; 32]` instead of `String`

ethers-core/src/types/transaction/eip712.rs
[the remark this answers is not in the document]
Ryanmtate: Updated the attribute parsing for the outer helper attribute! Thanks again!

ethers-core/src/types/transaction/eip712.rs
Ryanmtate: no longer need this method, deprecated.

ethers-signers/src/lib.rs
Ryanmtate: Still looking for a better solution than the following:

```rust
async fn sign_typed_data<T: Eip712 + Send + Sync>(
        &self,
        payload: T,
    ) -> Result<Signature, Self::Error> {
        let encoded = match payload.encode_eip712() {
            Ok(e) => e,
            Err(_) => {
                // NOTE: Still looking for a better solution for mapping to Infallible error type;
                unreachable!()
            }
        };

        Ok(self.sign_hash(H256::from(encoded), false))
    }
```

ethers-core/src/types/transaction/eip712.rs
```diff
@@ -0,0 +1,276 @@
...
+            Token::Uint(U256::from(keccak256(&self.version))),
+            Token::Uint(self.chain_id),
+            Token::Address(self.verifying_contract),
+        ])))
+    }
+}
+
+// Parse the AST of the struct to determine the domain attributes
+impl From<&syn::DeriveInput> for Eip712Domain {
+    fn from(input: &syn::DeriveInput) -> Eip712Domain {
```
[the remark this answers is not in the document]
Ryanmtate: Implementing `TryFrom` now and returning compile_error; removed core dump (panic!) behavior.

ethers-core/src/types/transaction/eip712.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+    .to_string()
+}
+
+/// Parse the field type from the derived struct
+pub fn parse_field(field: &syn::Field) -> String {
+    let field_path = match &field.ty {
+        syn::Type::Path(p) => p,
+        _ => {
+            panic!("field type must be a path")
```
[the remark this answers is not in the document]
Ryanmtate: Methods now returning `syn::parse::Error`

ethers-core/src/types/transaction/eip712.rs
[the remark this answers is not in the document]
Ryanmtate: Using `Vec<(_,_)>` now instead of HashMap and no longer sorting alphabetically. Only using the order defined in the struct.

ethers-core/ethers-derive-eip712/src/lib.rs
Ryanmtate: No longer using JSON to pass values around. Using `EthAbiType` for tokenizing the struct. Thanks for the tip!

ethers-core/src/types/transaction/eip712.rs
Ryanmtate: deprecated

ethers-core/src/types/transaction/eip712.rs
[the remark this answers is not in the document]
Ryanmtate: decided against using bytes due to conversion overhead, using `[u8;32` return type.

ethers-core/src/types/transaction/eip712.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// User defined error type;
+    type Error: std::error::Error + Send + Sync + std::fmt::Debug;
+
+    /// The eip712 domain is the same for all Eip712 implementations,
+    /// This method does not need to be manually implemented, but may be overridden
+    /// if needed.
+    fn eip712_domain_type_hash() -> String {
+        hex::encode(crate::utils::keccak256(
+            "EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)",
+        ))
```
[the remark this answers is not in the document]
Ryanmtate: Now providing constant type hashes for domain:

```rust
/// Pre-computed value of the following statement:
///
/// ```
/// keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)")
/// ```
///
pub const EIP712_DOMAIN_TYPE_HASH: [u8; 32] = [
    139, 115, 195, 198, 155, 184, 254, 61, 81, 46, 204, 76, 247, 89, 204, 121, 35, 159, 123, 23,
    155, 15, 250, 202, 169, 167, 93, 82, 43, 57, 64, 15,
];

/// Pre-computed value of the following statement:
///
/// ```
/// keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract,bytes32 salt)")
/// ```
///
pub const EIP712_DOMAIN_TYPE_HASH_WITH_SALT: [u8; 32] = [
    216, 124, 214, 239, 121, 212, 226, 185, 94, 21, 206, 138, 191, 115, 45, 181, 30, 199, 113, 241,
    202, 46, 220, 207, 34, 164, 108, 114, 154, 197, 100, 114,
];
```

ethers-signers/src/wallet/mod.rs
Ryanmtate: I've replaced `std::convert::Infallible` with `WalletError`: 

`type Error = WalletError;`

Added an enum for the `<T as Eip712>::Error` type for mapping the error to WalletError:

```rust
async fn sign_typed_data<T: Eip712 + Send + Sync>(
        &self,
        payload: T,
    ) -> Result<Signature, Self::Error> {
        let encoded = payload
            .encode_eip712()
            .map_err(|e| Self::Error::Eip712Error(e.to_string()))?;

        Ok(self.sign_hash(H256::from(encoded), false))
    }
```

Decided against implementing a trait bound on `WalletError` to satisfy `<T as Eip712>::Error` and decided to use a message string:

```rust
pub enum WalletError {

    ...

    /// Error type from Eip712Error message
    #[error("error encoding eip712 struct: {0:?}")]
    Eip712Error(String),
}
```

Also open to other suggestions for better error handling.

Ryanmtate:
@gakonst @mattsse  Looking to finish up on the example and tests today and open for a review.


One last issue I am battling is with the `example/eip712.rs` final encoding. 

Able to confirm `domain separator`, `type hash` and `struct hash` are all matching their solidity equivalents, but for some reason the final encoding does not match between the contract and the Eip712 trait derived macro implementing.

`examples/DeriveEip712Test.sol` and `examples/derive_eip712_abi.json` contain the full example, but the problem method is, `encodeEip712`:

```solidity
function encodeEip712(FooBar memory fooBar) public pure returns (bytes32) {
        return
            keccak256(
                abi.encodePacked(
                    "\\x19\\x01",
                    domainSeparator(),
                    structHash(fooBar)
                )
            );
    }
```

`examples/eip712.rs` file contains full text, but the test seems to fail on `encodeEip712` comparison:

```rust
let domain_separator = contract.domain_separator().call().await?;
    let type_hash = contract.type_hash().call().await?;
    let struct_hash = contract.struct_hash(derived_foo_bar.clone()).call().await?;
    let encoded = contract
        .encode_eip_712(derived_foo_bar.clone())
        .call()
        .await?;
    let verify = contract
        .verify_foo_bar(wallet.address(), derived_foo_bar, r, s, v)
        .call()
        .await?;

    assert_eq!(
        domain_separator,
        FooBar::domain_separator()?,
        "domain separator does not match contract domain separator!"
    );

    assert_eq!(
        type_hash,
        FooBar::type_hash()?,
        "type hash does not match contract struct type hash!"
    );

    assert_eq!(
        struct_hash,
        foo_bar.clone().struct_hash()?,
        "struct hash does not match contract struct struct hash!"
    );

    assert_eq!(
        encoded,
        foo_bar.encode_eip712()?,
        "Encoded value does not match!"
    );
```

mattsse:
hmm,
just to make sure I understand that snippet,
- [x] domain seprator matches
- [x] struct hash matches

but `encode_eip712` fails?

this would mean that 

```solidity
                abi.encodePacked(
                    "\\x19\\x01",
                    domainSeparator(),
                    structHash(fooBar)
                )
```
is not identical to 

```rust
               let digest_input = [
                    &[0x19, 0x01],
                    &Self::domain_separator()?[..],
                    &self.struct_hash()?[..]
                ].concat();
```

and since the values (domainSeparator, structHash) are identical (ensured by the ), this would indicate `encodePacked` is not identical to `concat?`
could you check that?

gakonst:
I think `abi.encode` is what's equivalent to `concat` I Believe. I think encode packed is a non-standard encoding. It may be exposed via ethabi though I haven't investigated.

mattsse:
> Great news! The problem was much simpler. Let it be known that `\\x19\\x01` does not encode to `&[x19, x01]`, but `\x19\x01` does.
> 
> Going to open this up for review after resolving merge conflicts.

does &[x19, x01][..] do?

Ryanmtate:
> > Great news! The problem was much simpler. Let it be known that `\\x19\\x01` does not encode to `&[x19, x01]`, but `\x19\x01` does.
> > Going to open this up for review after resolving merge conflicts.
> 
> does &[x19, x01][..] do?

both `&[0x19, 0x01][..]` and `&[0x19, 0x01]` are equivalent and map to `\x19\x01` in solidity. 

however, both do not work for `\\x19\\x01` in solidity, which I had seen used in an [example](https://medium.com/metamask/eip712-is-coming-what-to-expect-and-how-to-use-it-bb92fd1a7a26) for implementing eip-712.

ethers-core/ethers-derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+// Main implementation macro, used to compute static values and define
+// method for encoding the final eip712 payload;
+fn impl_eip_712_macro(ast: &syn::DeriveInput) -> TokenStream {
+    // Primary type should match the type in the ethereum verifying contract;
+    let primary_type = &ast.ident;
+
+    // Computer domain separator
+    let domain = match eip712::EIP712Domain::try_from(ast) {
+        Ok(attributes) => attributes,
+        Err(e) => return TokenStream::from(e),
```
mattsse, line 85: `syn::Error`

ethers-core/ethers-derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+                let mut items = vec![ethers_core::abi::Token::Uint(
+                    ethers_core::types::U256::from(&Self::type_hash()?[..]),
+                )];
+
+                if let ethers_core::abi::Token::Tuple(tokens) = self.clone().into_token() {
+                    for token in tokens {
+                        match &token {
+                            ethers_core::abi::Token::Tuple(t) => {
+                                // TODO: check for nested Eip712 Type;
+                                // Challenge is determining the type hash
```
mattsse, line 131: should this fail for the time being?

ethers-core/ethers-derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+            fn encode_eip712(self) -> Result<[u8; 32], Self::Error> {
+                // encode the digest to be compatible with solidity abi.encodePacked()
+                // See: https://github.com/gakonst/ethers-rs/blob/master/examples/permit_hash.rs#L72
+                let digest_input = [
+                    &[0x19, 0x01],
+                    &Self::domain_separator()?[..],
+                    &self.struct_hash()?[..]
+                ].concat();
+
+                // let digest_input = &[0x19, 0x01];
```
mattsse, line 156:
```suggestion
```

ethers-core/ethers-derive-eip712/src/lib.rs
Ryanmtate: I think it is okay to do this here. This is returning a syn::Error, however, we don't have access to a span on the `DeriveInput` input. I see this is also how `ethers-contract-derive` is handling it, [here](https://github.com/gakonst/ethers-rs/blob/master/ethers-contract/ethers-contract-derive/src/lib.rs#L128).

Open to exploring directly using `syn::Error`, but not sure what the idiomatic way is to handle the error without the span, and changing the `proc_macro::TokenStream` to `proc_macro2::TokenStream`.

ethers-core/ethers-derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+        Ok(attributes) => attributes,
+        Err(e) => return TokenStream::from(e),
+    };
+
+    let domain_separator = hex::encode(domain.separator());
+
+    // Must parse the AST at compile time.
+    let parsed_fields = match eip712::parse_fields(ast) {
+        Ok(fields) => fields,
+        Err(e) => return TokenStream::from(e),
```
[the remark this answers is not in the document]
Ryanmtate: see previous comment.

ethers-core/ethers-derive-eip712/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+//!
+use std::convert::TryFrom;
+
+use ethers_core::types::transaction::eip712;
+use proc_macro::TokenStream;
+use quote::quote;
+
+#[proc_macro_derive(Eip712, attributes(eip712))]
+pub fn eip_712_derive(input: TokenStream) -> TokenStream {
+    let ast = syn::parse(input).expect("failed to parse token stream for Eip712 derived struct");
```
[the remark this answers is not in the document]
Ryanmtate: perhaps using `let ast = parse_macro_input!(input);` is preferred here?

ethers-core/ethers-derive-eip712/src/lib.rs
Ryanmtate: fails with error now, but should be unreachable in practice. Will think about a solution for the nested struct, but it might require some re-working to get it done. Hopefully this is a good start.

Cargo.toml
```diff
@@ -19,7 +19,8 @@ members = [
     "ethers-signers",
     "ethers-core",
     "ethers-middleware",
-    "ethers-etherscan"
+    "ethers-etherscan",
+    "ethers-core/ethers-derive-eip712",
```
gakonst, line 23: Is this needed? I think we should just be re-exporting the macro/traits from ethers-core crate

Cargo.toml
```diff
@@ -71,13 +72,13 @@ yubi = ["ethers-signers/yubi"]
...
 abigen = ["ethers-contract/abigen"]
 
-
 [dependencies]
 ethers-contract = { version = "^0.5.0", default-features = false, path = "./ethers-contract" }
 ethers-core = { version = "^0.5.0", default-features = false, path = "./ethers-core", features = ["setup"] }
 ethers-providers = { version = "^0.5.0", default-features = false, path = "./ethers-providers" }
 ethers-signers = { version = "^0.5.0", default-features = false, path = "./ethers-signers" }
 ethers-middleware = { version = "^0.5.0", default-features = false, path = "./ethers-middleware" }
+ethers-derive-eip712 = { version = "0.1.0", path = "./ethers-core/ethers-derive-eip712"}
```
gakonst, line 81: no need here if we re-export the macro

ethers-signers/src/lib.rs
```diff
@@ -93,6 +95,13 @@ pub trait Signer: std::fmt::Debug + Send + Sync {
     /// Signs the transaction
     async fn sign_transaction(&self, message: &TypedTransaction) -> Result<Signature, Self::Error>;
 
+    /// Encodes and signs the typed data according EIP-712
```
gakonst, line 98:
```suggestion
    /// Encodes and signs the typed data according EIP-712.
```

examples/eip712.rs
```diff
@@ -0,0 +1,125 @@
...
+    bar: U256,
+    fizz: Vec<u8>,
+    buzz: [u8; 32],
+    far: String,
+    out: Address,
+}
+
+abigen!(
+    DeriveEip712Test,
+    "./examples/derive_eip712_abi.json",
```
gakonst, line 30: This makes me realize we may want abigen to also work directly with solidity file so that it can generate both contract factories and just have easier UX

examples/eip712.rs
```diff
@@ -0,0 +1,125 @@
...
+    utils::{compile_and_launch_ganache, keccak256, Ganache, Solc},
+};
+use ethers_derive_eip712::*;
+
+#[derive(Debug, Clone, Eip712, EthAbiType)]
+#[eip712(
+    name = "Eip712Test",
+    version = "1",
+    chain_id = 1,
+    verifying_contract = "0x0000000000000000000000000000000000000001",
```
gakonst, line 16: Do you think it'd be worth allowing functions which specify `chain_id` and `verifying_contract` at runtime? That way if Im testing on both mainnet and testnet, I don't need to modify my code and can just provide it as a param.

gakonst (commented):
Overall LGTM - pending @mattsse's approval. Some nits / suggestions.

ethers-core/ethers-derive-eip712/src/lib.rs
mattsse: yeah you're right, sgtm

Cargo.toml
mattsse: +1 on that, perhaps feature gated in core? default is enabled

examples/eip712.rs
mattsse: Perhaps we can make the `EIP712Domain` type part of the trait so that instead of the hash, `domain_type` returns `EIP712Domain` which has `EIP712Domain::encode` and additional setters. This way we can add a type that wraps around a `EIP712Domain` and takes an `Eip712` type as reference when encoding the actual eip712.

examples/eip712.rs
mattsse: I have a draft for that somewhere already that went stale a month ago or so lol, will revive that

Cargo.toml
Ryanmtate: I don't think we can re-export the macro since it depends on ethers-core. From my understanding, proc-macros can only have one public export in the lib. We can remove the dependency here if we move the example into the `ethers-contracts/test` folder, which I think is preferable anyway.

examples/eip712.rs
Ryanmtate: I like the flexible to pass in runtime values for the domain. After considering the above solution, I opted for providing an extra parameter for the domain when calling encode: 

```rust
fn encode_eip712(self, domain: Option<EIP712Domain>) -> Result<[u8; 32], Self::Error>;
```

I found this to be a simple solution, without needing to implement a Eip712 type, which may not actually correspond to the domain type hash. I think using the `EIP712Domain` type is more defensible against user error.

The macro implementation looks like:

```rust
fn encode_eip712(self, domain: Option<ethers_core::types::transaction::eip712::EIP712Domain>) -> Result<[u8; 32], Self::Error> {
                // encode the digest to be compatible with solidity abi.encodePacked()
                // See: https://github.com/gakonst/ethers-rs/blob/master/examples/permit_hash.rs#L72

                let domain_separator = if let Some(d) = domain {
                    d.separator()
                } else {
                    Self::domain_separator()?
                };


                let digest_input = [
                    &[0x19, 0x01],
                    &domain_separator[..],
                    &self.struct_hash()?[..]
                ].concat();

                return Ok(ethers_core::utils::keccak256(digest_input));
}
```

This also means we need to expose the parameter to the signer trait for `sign_typed_data`, like:

```rust
/// Encodes and signs the typed data according EIP-712.
    /// Payload must implement Eip712 trait.
    /// * `domain` - Optional Eip712 domain struct to override eip712 macro attribute helpers for Eip712 Type `T`;
    async fn sign_typed_data<T: Eip712 + Send + Sync>(
        &self,
        payload: T,
        domain: Option<EIP712Domain>,
    ) -> Result<Signature, Self::Error>;
```

Overall, I think this might be quick solution to the request, but has the downside of needed an extra parameter, but I don't think that is too large a burden.

Open to other feedback here, but we probably should look to open a new PR to introduce such a request and get the current works landed.

examples/eip712.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+abigen!(
+    DeriveEip712Test,
+    "./examples/derive_eip712_abi.json",
+    event_derives(serde::Deserialize, serde::Serialize)
+);
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
Ryanmtate: moved to `ethers-contract/tests/`

Cargo.toml
Ryanmtate: removed dependency here.

Cargo.toml
Ryanmtate: Added a feature gate under `ethers_core::types::transaction::mod`:

```rust
#[cfg(feature = "eip712")]
pub mod eip712;
```

Also included a quick comment in the docs for it, see c911087

ethers-signers/src/ledger/mod.rs
```diff
@@ -25,6 +29,16 @@ impl Signer for LedgerEthereum {
...
     }
 
+    async fn sign_typed_data<T: Eip712 + Send + Sync>(
+        &self,
+        payload: T,
+        domain: Option<EIP712Domain>,
+    ) -> Result<Signature, Self::Error> {
+        let hash = payload.encode_eip712(domain)?;
+
+        Ok(self.sign_hash(hash.into(), false))
```
sebastinez, line 39: It seems that `sign_hash` is not available on `LedgerEthereum` which you are implementing for here. Should this be eventually `sign_message?`

ethers-signers/src/ledger/mod.rs
```diff
@@ -25,6 +29,16 @@ impl Signer for LedgerEthereum {
         self.sign_tx(message).await
     }
 
+    async fn sign_typed_data<T: Eip712 + Send + Sync>(
+        &self,
+        payload: T,
+        domain: Option<EIP712Domain>,
+    ) -> Result<Signature, Self::Error> {
+        let hash = payload.encode_eip712(domain)?;
```
sebastinez, line 37: As I said earlier out of my comfort zone for now hehe.
But it seems that the Eip712 Error cannot be cast to a `LedgerError` here, Eventually we can handle the Result differently?
```
the trait `From<<T as Eip712>::Error>` is not implemented for `ledger::types::LedgerError`
```

gakonst:
```
---- eth_tests::test_derive_eip712 stdout ----
thread 'eth_tests::test_derive_eip712' panicked at 'called `Result::unwrap()` on an `Err` value: SolcError("Error: Source file requires different compiler version (current compiler is 0.6.6+commit.6c089d02.Linux.g++ - note that nightly builds are considered to be strictly less than the released version\n --> tests/solidity-contracts/DeriveEip712Test.sol:2:1:\n  |\n2 | pragma solidity ^0.8.0;\n  | ^^^^^^^^^^^^^^^^^^^^^^^\n\n")', ethers-contract/tests/common/mod.rs:35:10
error: test failed, to rerun pass '-p ethers-contract --test contract'
```

Just need to fix the solc version to be the same one as the one installed in CI. 

The WASM test is also seemingly failing, is this a flake?

mattsse:
> ```
> ---- eth_tests::test_derive_eip712 stdout ----
> thread 'eth_tests::test_derive_eip712' panicked at 'called `Result::unwrap()` on an `Err` value: SolcError("Error: Source file requires different compiler version (current compiler is 0.6.6+commit.6c089d02.Linux.g++ - note that nightly builds are considered to be strictly less than the released version\n --> tests/solidity-contracts/DeriveEip712Test.sol:2:1:\n  |\n2 | pragma solidity ^0.8.0;\n  | ^^^^^^^^^^^^^^^^^^^^^^^\n\n")', ethers-contract/tests/common/mod.rs:35:10
> error: test failed, to rerun pass '-p ethers-contract --test contract'
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> Just need to fix the solc version to be the same one as the one installed in CI.
> 
> The WASM test is also seemingly failing, is this a flake?

weird, only the http provider wasm test fails due to a 400 ` MiddlewareError(MiddlewareError(JsonRpcClientError(SerdeJson { err: Error("invalid type: integer `400`, expected struct Response", line: 1, column: 3), text: "400 Bad Request" })))`
still works locally -.-

Ryanmtate:
Thanks @gakonst @mattsse for the help and pointers!

gakonst:
@Ryanmtate need to enable abi encoder v2 i think

PR #482, opened 2021-10-01
feat: support human readable struct inputs

mattsse:
## Motivation
Support custom structs as contract params in human readable abi

## Solution
* keep track of the function's types (input, output) in the abi parser, this is added on top the existing parser and adds a bit more verbosity in to the param parse functions...
* reuse the existing `InternalStructs` when determine function inputs

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog


gakonst (approved):
godlike

Issue #485, opened 2021-10-03
Add dedicated etherscan client


mattsse:
Tracking issue for complete `ethers-etherscan` client crate

## Endpoints

- [ ] accounts
- [ ] contracts
- [ ] transactions
- [ ] blocks
- [ ] tokens
- [ ] gas tracker
- [ ] stats

Overview https://docs.etherscan.io/

Initial WIP https://github.com/gakonst/dapptools-rs/pull/82

## Other todos
- [ ] replace the current `ethers` internal usage in `abigen`

gakonst:
Transactions were implemented in #512

PR #486, opened 2021-10-04
feat: add etherscan client crate

mattsse:
## Motivation
Add etherscan.io client,
Ref #485 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-etherscan/src/lib.rs
+++ b/ethers-etherscan/src/lib.rs
@@ -222,7 +222,8 @@ pub struct VerifyContract {
         skip_serializing_if = "Option::is_none"
     )]
     pub constructor_arguments: Option<String>,
-    pub evmversion: Option<String>,
+    #[serde(rename = "evmversion")]
+    pub evm_version: Option<String>,
     #[serde(flatten)]
     pub other: HashMap<String, String>,
 }
@@ -238,7 +239,7 @@ impl VerifyContract {
             optimization_used: None,
             runs: None,
             constructor_arguments: None,
-            evmversion: None,
+            evm_version: None,
             other: Default::default(),
         }
     }
@@ -276,8 +277,8 @@ impl VerifyContract {
         self
     }
 
-    pub fn evmversion(mut self, evmversion: impl Into<String>) -> Self {
-        self.evmversion = Some(evmversion.into());
+    pub fn evm_version(mut self, evm_version: impl Into<String>) -> Self {
+        self.evm_version = Some(evm_version.into());
         self
     }
 
@@ -446,9 +447,6 @@ mod tests {
                 .optimization(true)
                 .runs(200);
 
-        let _resp = client
-            .submit_contract_verification(&contract)
-            .await
-            .unwrap();
+        let resp = client.submit_contract_verification(&contract).await;
     }
 }
--- a/ethers-etherscan/src/lib.rs
+++ b/ethers-etherscan/src/lib.rs
@@ -1,0 +1,454 @@
+//! Bindings for [etherscan.io web api](https://docs.etherscan.io/)
+
+use ethers_core::abi::{Abi, Address};
+use reqwest::{header, Url};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::{borrow::Cow, collections::HashMap};
+
+/// The Etherscan.io api client.
+#[derive(Clone)]
+pub struct Client {
+    /// The client that executes the http requests
+    client: reqwest::Client,
+    /// The etherscan api key
+    api_key: String,
+    /// API endpoint like https://api(-chain).etherscan.io/api
+    etherscan_api_url: Url,
+    /// Base etherscan endpoint like https://etherscan.io/address
+    etherscan_url: Url,
+}
+
+impl Client {
+    /// Create a new client with the correct endpoints based on the chain.
+    ///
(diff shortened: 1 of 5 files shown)


ethers-etherscan/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+    }
+
+    #[tokio::test]
+    #[ignore]
+    async fn can_verify_contract() {
+        // TODO this needs further investigation
```
gakonst, line 430: Is this working now or does it still require more investigation?

ethers-etherscan/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+
+        // https://etherscan.io/address/0x9e744c9115b74834c0f33f4097f40c02a9ac5c33#code
+        let contract = include_str!("../resources/UniswapExchange.sol");
+        let address = "0x9e744c9115b74834c0f33f4097f40c02a9ac5c33"
+            .parse()
+            .unwrap();
+        let compiler_version = "v0.5.17+commit.d19bba13";
+        let constructor_args = "0x000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000005f5e1000000000000000000000000000000000000000000000000000000000000000007596179537761700000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000035941590000000000000000000000000000000000000000000000000000000000";
```
gakonst, line 438: I wonder if we should move `BaseContract` to ethers-core now to encode constructor arguments in a nice way

ethers-etherscan/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+        self.optimization_used = Some("0".to_string());
+        self
+    }
+
+    pub fn code_format(mut self, code_format: CodeFormat) -> Self {
+        self.code_format = code_format;
+        self
+    }
+
+    pub fn evmversion(mut self, evmversion: impl Into<String>) -> Self {
```
gakonst, line 279:
```suggestion
    pub fn evm_version(mut self, evm_version: impl Into<String>) -> Self {
```

ethers-etherscan/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+    optimization_used: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub runs: Option<String>,
+    /// NOTE: there is a typo in the etherscan API `constructorArguements`
+    #[serde(
+        rename = "constructorArguements",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub constructor_arguments: Option<String>,
+    pub evmversion: Option<String>,
```
gakonst, line 225:
```suggestion
    pub evm_version: Option<String>,
```

ethers-etherscan/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+}
+
+/// The type that gets serialized as query
+#[derive(Debug, Serialize)]
+struct Query<'a, T: Serialize> {
+    apikey: Cow<'a, str>,
+    module: Cow<'a, str>,
+    action: Cow<'a, str>,
+    #[serde(flatten)]
+    other: T,
```
gakonst, line 198: smart. never thought about that.

ethers-etherscan/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+    /// Get Contract Source Code for Verified Contract Source Codes
+    /// ```no_run
+    /// # use ethers_etherscan::Client;
+    ///
+    /// # #[tokio::main]
+    /// # async fn main() -> anyhow::Result<()> {
+    ///     let client = Client::new("mainnet", "API_KEY").unwrap();
+    ///     let meta = client
+    ///         .contract_source_code("0xBB9bc244D798123fDe783fCc1C72d3Bb8C189413".parse().unwrap())
+    ///         .await?;
```
gakonst, line 167: Should we also compile this with Solc? Or provide some helper way to convert this to a `CompiledContract`

ethers-etherscan/src/lib.rs
mattsse: still get this response _instantly_, so I think I messed up the payload somehow, but tried various, alternatives like numbers as strings, etc.., will need to take a closer look at dapptools itself or any other etherscan client
```
    Response {
        status: "0",
        message: "NOTOK",
        result: "Error!",
    },
```

ethers-etherscan/src/lib.rs
mattsse: that'd be useful, I think we touched on abstracting functions as a trait via `abigen!` some time ago, I think it makes sense to follow up on this and move all general abstractions to core.

ethers-etherscan/src/lib.rs
mattsse: the etherscan api doesn't seem to provide the bytecode directly, however it is present in the html reponse with `div[id=verifiedbytecode2]` -.-
Additional tooling to instantly compile this is probably useful, I'll add this to the tracking issue

gakonst (approved):
Gg

PR #514, opened 2021-10-16
test(middleware): fix flapping gas oracle

shekhirin:
## Motivation

[CI tests are flapping](https://github.com/gakonst/ethers-rs/runs/3914532490) because gas price changes between retrieving from Gas Station as expected and using it in middleware.

## Solution

Implement fake gas price oracle, so we can be sure price doesn't change during the test.

## PR Checklist

- [x] Implement `FakeGasOracle`


gakonst:
Nice one! Thank you.

PR #518, opened 2021-10-17
fix: eip712 signing with ledger hw

sebastinez:
# Motivation
This PR fixes the way a EIP712 derived struct is sent to the Ledger HW to obtain the signature.

# Solution
Based on the following specification: https://github.com/LedgerHQ/app-ethereum/blob/master/doc/ethapp.asc#sign-eth-eip-712
- Adds a new `INS` variant = `SIGN_ETH_EIP_712`
- Replaces `self.sign_message` with `self.signed_typed_struct`
- `signed_typed_struct` creates a payload consisting of the domain separator and the typed data hash.
- This allows the Ledger to display both in its display to be signed by the user.
- It also checks if the Ledger ETH app is at least 1.6.0 since this is a requirement, according to https://github.com/LedgerHQ/app-ethereum/issues/105#issuecomment-765316999.
  - If you prefer a different way to handle this issue please let me know!

# PR Checklist
- [ ]  Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

I tested the generated signature on a deployed contract on Rinkeby and it works great.
Please let me know if you have any feedback @gakonst @prestwich @Ryanmtate.

### DIFF
--- a/ethers-contract/tests/contract.rs
+++ b/ethers-contract/tests/contract.rs
@@ -660,7 +660,7 @@ mod eth_tests {
         };
 
         let sig = wallet
-            .sign_typed_data(foo_bar.clone())
+            .sign_typed_data(&foo_bar)
             .await
             .expect("failed to sign typed data");
 


sebastinez:
> Can we add a test case to the [tests](https://github.com/gakonst/ethers-rs/blob/master/ethers-signers/src/ledger/app.rs#L200) module?

Yeah should be no problem, the thing I'm missing is since I'm not able to mock a Ledger (at least I think so) it's kinda hard to assert something.
We could eventually check that an error gets thrown when the Eth app version does not match semver.

Did you have some specific test in mind?

gakonst:
I think just writing a simple test for any 712 struct that you ignore by default with the Ledger signer is sufficient. Just want something that I could run locally and check that it works!

gakonst:
OK this looks great. Thank you for the work on it. I made the test slightly tighter and manually verified that it works!

PR #539, opened 2021-10-28
feat: improved solc management

mattsse:
## Motivation
provide a way to configure a project root with cache support.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts.rs
+++ b/ethers-solc/src/artifacts.rs
@@ -311,11 +311,17 @@ impl Source {
             .collect()
     }
 
+    /// Generate a non-cryptographically secure checksum of the file's content
     pub fn content_hash(&self) -> String {
         let mut hasher = md5::Md5::new();
         hasher.update(&self.content);
         let result = hasher.finalize();
         hex::encode(result)
+    }
+
+    /// Returns all import statements of the file
+    pub fn parse_imports(&self) -> Vec<&str> {
+        utils::find_import_paths(self.as_ref())
     }
 }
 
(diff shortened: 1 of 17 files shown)


ethers-solc/src/artifacts.rs
```diff
@@ -397,7 +397,7 @@ impl<'a> OutputDiagnostics<'a> {
 impl<'a> fmt::Display for OutputDiagnostics<'a> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         if !self.has_error() {
-            f.write_str("Compiler run successful")?;
+            f.write_str("Compiler run successful\n")?;
```
mattsse, line 400:
```suggestion
            writeln!(f, "Compiler run successful")?;
```

ethers-solc/src/artifacts.rs
```diff
@@ -379,28 +382,38 @@ impl CompilerOutput {
...
-            f.write_str("Compiler run successful\n")?;
+            f.write_str("Compiler run successful")?;
         }
-        for err in self.0 {
-            writeln!(f, "{}", err)?;
+        for err in self.errors {
+            // Do not log any ignored error codes
+            if let Some(error_code) = err.error_code {
+                if !self.ignored_error_codes.contains(&error_code) {
+                    writeln!(f, "\n{}", err)?;
```
gakonst, line 412: I prefer the newline here so that if there's no errors there's no extra newline in the Compiler run successful log, otherwise it adds an empty line which is redundant

gakonst (approved):
Looks great.

To summarize what we have from this PR:

Allows configuring repo structures for Hardhat & DappTools like below:

```rust
use ethers_solc::{Project, ProjectPathsConfig};

fn main() {
    eprintln!("Building");

    // configure the project with all its paths, solc, cache etc.
    let project = Project::builder()
        // replace with `ProjectPathsConfig::dapptools` for the src/lib/out repo structure
        .paths(ProjectPathsConfig::hardhat(env!("CARGO_MANIFEST_DIR")).unwrap())
        .build()
        .unwrap();
    let output = project.compile().unwrap();
    eprintln!("{}", output);
}
```

This outputs artifacts with ABI, Bytecode, Runtime-Bytecode under `artifacts/` as well as a file under the`cache/` dir which is used to check which files have changed since the last time. 

This PR _does not_ have full support for:
1. Incremental compilation (any file changing => everything changes), this should be easy to fix in the `is_changed` function
2. Smart remappings detection
3. A way for a consumer to read the artifact associated with a contract programmatically from memory w/o reading from disk
4. Automatic Solc detection

PR #542, opened 2021-10-30
test(middleware): get rinkeby private key from env var

shekhirin:
## Motivation

Test is failing because account needs to have some ether in it and we got no ether in it, sad. If we'll send some ether to the address used in the test, it'll be drained quickly, no bueno.

## Solution

Use private key for rinkeby account from environment variable (@gakonst needs to set it for GitHub Actions and send some rinkeby ether to this account)


gakonst:
Gonna set this up today

PR #546, opened 2021-10-30
feat: use ethers_solc::Solc instead of ethers_core::utils::Solc

gakonst:
Description to come in the morning


examples/contract_human_readable.rs
```diff
@@ -19,13 +17,30 @@ abigen!(
...
+        .ephemeral()
+        .artifacts(ArtifactOutput::Nothing)
+        .build()
+        .unwrap();
+    // compile the project and get the artifacts
+    let compiled = solc.compile().unwrap();
+    let compiled = match compiled {
+        ProjectCompileOutput::Compiled((output, _)) => output,
+        _ => panic!("expected compilation artifacts"),
+    };
```
mattsse, line 37: I feel like we should think about either:

* adding an error type that treats a CompilerOutput with a hard error as `Err(CompilerOutput)`, this will be more convenient for build scripts and gets rid of the `println!("{}", output);` to print diagnostics
* and an `CompilerOutput::into_result` function

PR #548, opened 2021-10-31
fix: preserve underscores in case of collisions

### LINKED ISSUE #530: handle underscore in overloaded functions
Currently, methods can be assigned an alias in order to support overloaded functions, but

```solidity
  function method()...
  function _method()
```
are different functions, but conflict due to the way `Inflector` works, this causes a clash during EthCall and contract expansion

I that case also prefix the function and EthCall structs with an `_`

**Additional context**
Ref #505

mattsse:
## Motivation
Close #530

## Solution
leading, Trailing `_` are preserved, only if they would cause conflicts. See new test for how different conflicts are handled,

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
amazing

PR #551, opened 2021-11-02
feat(solc): Multiple Solc Version detection

gakonst:
We port over the logic for detecting and automatically using the right solc version from [dapptools-rs solc](https://github.com/gakonst/dapptools-rs/tree/master/solc), leveraging the SVM and Async features. If these features are disabled, no automatic solc detection will be done, and it will use the solc version as defined in `ethers_solc::Solc`'s rules.

### DIFF
--- a/ethers-solc/src/compile.rs
+++ b/ethers-solc/src/compile.rs
@@ -35,8 +35,13 @@ pub const BERLIN_SOLC: Version = Version::new(0, 8, 5);
 /// https://blog.soliditylang.org/2021/08/11/solidity-0.8.7-release-announcement/
 pub const LONDON_SOLC: Version = Version::new(0, 8, 7);
 
-#[cfg(all(feature = "svm", feature = "async"))]
+#[cfg(any(test, all(feature = "svm", feature = "async")))]
 use once_cell::sync::Lazy;
+
+#[cfg(any(test, feature = "tests"))]
+use std::sync::Mutex;
+#[cfg(any(test, feature = "tests"))]
+static LOCK: Lazy<Mutex<()>> = Lazy::new(|| Mutex::new(()));
 
 #[cfg(all(feature = "svm", feature = "async"))]
 /// A list of upstream Solc releases, used to check which version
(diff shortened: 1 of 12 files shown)


ethers-solc/src/lib.rs
```diff
@@ -93,8 +93,58 @@ impl Project {
...
+    /// NB: If the `svm` feature is enabled, this function will automatically detect
+    /// solc versions across files.
     pub fn compile(&self) -> Result<ProjectCompileOutput> {
-        let mut sources = self.sources()?;
+        let sources = self.sources()?;
+
+        #[cfg(not(all(feature = "svm", feature = "async")))]
+        {
+            self.compile_with_version(&self.solc, sources)
+        }
```
mattsse, line 105: should this be also configureable via some Project settings flag?

mattsse (approved):
lgtm

ethers-solc/src/lib.rs
gakonst: I don't think so? Isn't it always better to support many?

ethers-solc/src/compile.rs
```diff
@@ -87,6 +112,74 @@ impl Solc {
         Ok(solc)
     }
 
+    /// Assuming the `versions` array is sorted, it returns the first element which satisfies
```
mattsse, line 115: Assuming versions are sorted asc, right?
```suggestion
    /// Assuming the `versions` array is sorted, it returns the latest element which satisfies
```

ethers-solc/src/lib.rs
mattsse: you're right, that's better!

ethers-solc/src/compile.rs
```diff
@@ -34,10 +35,34 @@ pub const BERLIN_SOLC: Version = Version::new(0, 8, 5);
...
+
+#[cfg(all(feature = "svm", feature = "async"))]
+/// A list of upstream Solc releases, used to check which version
+/// we should download.
+pub static RELEASES: Lazy<Vec<Version>> = Lazy::new(|| {
+    // Try to download the releases, if it fails default to empty
+    match tokio::runtime::Runtime::new()
+        .expect("could not create tokio rt to get remote releases")
+        // TODO: Can we make this future timeout at a small time amount so that
+        // we do not degrade startup performance if the consumer has a weak network?
```
mattsse, line 54: additional timeout is great, but we also need to consider that there might no network at all. I guess reqwest would simply fail so this wouldn't really be an issue is therefore negligible?

gakonst: I think so, still leaving the todo there just in case

PR #552, opened 2021-11-03
Fix missing fmt argument in error message

dtolnay:
Without this, the error message would literally be "root path '{}' is not absolute" with curly braces in it instead of the root path.

Test plan:

`cd ethers-contract/ethers-contract-abigen`
`cargo check`
`cargo check --target wasm32-unknown-unknown`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Unreleased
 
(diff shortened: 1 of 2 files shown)


gakonst (approved):
Ah! Excellent catch! Thank you.

PR #554, opened 2021-11-03
Feat/add call deployer

frjnn:
## Motivation

As it can be useful, for plentiful reasons, to perform one or several dry test runs of a contract's deployment, I hereby propose an implementation of `.call()` for the `Deployer` struct.

## Solution

The implementation is straightforward, as it consist in invoking the client's`.call()` method from the `Middleware` trait to simulate the deployment. Such doing leaves plenty of room to handle reverts in a more structured way in the future.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Unreleased
 
-- Add `.call()` method to `Deployer` for performing dry runs of contract deployments.
+- Add `.call()` method to `Deployer` for performing dry runs of contract deployments. [#554](https://github.com/gakonst/ethers-rs/pull/554)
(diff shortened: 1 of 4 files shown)


gakonst (approved):
LGTM - thank you

frjnn:
Looks like something's wrong with [this test](https://github.com/gakonst/ethers-rs/blob/23fb877c16c70a81825506ba54ea1c2bf915e19a/ethers-middleware/tests/nonce_manager.rs#L20), I don't understand what's causing it, seems unrelated to me.

gakonst:
It's unrelated, merged, thank you!

Issue #494, opened 2021-10-09
create ethers-macros


gakonst:
We should create a new ethers-macros crate which contains useful macros for parsing solidity datatypes and generating ethers types.

_Originally posted by @Ryanmtate in https://github.com/gakonst/ethers-rs/issues/16#issuecomment-938735219_

sebastinez:
This issue has been solved in #555 and can be closed

PR #555, opened 2021-11-04
Refactor crate determination in new ethers-macro crate

sebastinez:
## Motivation
Currently when wanting to implement the EIP712 derive macro, one has to import `ethers_core` as a separate dependency, or declare it like `ethers::core as ethers_core`.
This is cumbersome and not clean.

## Solution
By making use of the convenience function in `ethers::contract::abigen::ethers_core_crate()` we can compute a path and show it in the derive macro.
By doing this we can avoid this additional `ethers_core` dependency in external projects.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/Cargo.toml
+++ b/ethers-core/Cargo.toml
@@ -14,6 +14,7 @@ rlp = { version = "0.5.0", default-features = false }
 ethabi = { version = "15.0", default-features = false, features = ["full-serde", "rlp"] }
 arrayvec = { version = "0.7.2", default-features = false }
 rlp-derive = { version = "0.1.0", default-features = false }
+cargo_metadata = "0.14.0"
 
 # crypto
 ecdsa = { version = "0.12.3", default-features = false, features = ["std"] }
@@ -55,6 +56,7 @@ celo = ["legacy"] # celo support extends the transaction format with extra field
 setup = ["tokio", "futures-util"] # async support for concurrent setup
 legacy = []
 eip712 = ["convert_case", "syn", "quote", "proc-macro2"]
+macros = ["syn"]
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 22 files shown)


sebastinez:
Hey @gakonst, this should solve the issue with needing to import `ethers_core` as separate dependency.
Let me know if you have any feedback regarding this PR.

mattsse:
this would be an opportunity to think about `ethers-macros` #494 again.
At first, we could simply create it and move the create detection there and then figure out how to proceed.

gakonst:
Yeah so: Let's make an `ethers-macros` crate and pull in the crate detection functions there, and use them in both ethers-contract-abigen and ethers-712

sebastinez:
> Yeah so: Let's make an `ethers-macros` crate and pull in the crate detection functions there, and use them in both ethers-contract-abigen and ethers-712. Code changes look good otherwise beyond the dependencies.

Okay no problem!
Will do a push today with these changes

gakonst:
Amazing. Thank you.

ethers-core/Cargo.toml
```diff
@@ -14,6 +14,7 @@ rlp = { version = "0.5.0", default-features = false }
 ethabi = { version = "15.0", default-features = false, features = ["full-serde", "rlp"] }
 arrayvec = { version = "0.7.2", default-features = false }
 rlp-derive = { version = "0.1.0", default-features = false }
+cargo_metadata = "0.14.0"
```
gakonst, line 17: Should this be `optional = true` and enabled only if `macros` is enabled?

gakonst (approved):

ethers-core/Cargo.toml
sebastinez: Yeah that makes sense since it's only used by the macros crate for now

PR #562, opened 2021-11-06
add GethTrace

frjnn:
## Motivation

Currently ethers-rs lacks support for transaction tracing when using the Geth client. This pull request extends the `Middleware` trait to include such functionalities.

## Solution

The implementation consists in the `GethTrace` and `OpCodeLog` structs, which handle the trace type, and the `geth_trace_transaction` method, which sends a `debug_traceTransaction` json RPC request to the node.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/types/trace/filter.rs
+++ b/ethers-core/src/types/trace/filter.rs
@@ -1,5 +1,5 @@
 //! Types for the Parity Transaction-Trace Filtering API
-use crate::types::{Address, BlockNumber, Bytes, H160, H256, U256};
+use crate::types::{trace::BTreeMap, Address, BlockNumber, Bytes, H160, H256, U256};
 use serde::{Deserialize, Serialize};
 
 /// Trace filter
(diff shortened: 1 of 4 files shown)


ethers-core/src/types/trace/filter.rs
```diff
@@ -266,6 +266,32 @@ pub enum RewardType {
     External,
 }
 
+/// Basic trace type for the Geth client.
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct GethTrace {
+    failed: bool,
+    gas: i64,
```
mattsse, line 273: lets make gas `U256`, that's consistent across other ethers types

ethers-core/src/types/trace/filter.rs
```diff
@@ -266,6 +266,32 @@ pub enum RewardType {
...
+    depth: u64,
+    error: Option<String>,
+    gas: u64,
+    #[serde(rename = "gasCost")]
+    gas_cost: u64,
+    memory: Option<Bytes>,
+    op: String,
+    pc: usize,
+    stack: Option<Vec<U256>>,
+    storage: Option<BTreeMap<H256, H256>>,
```
mattsse, line 292: I'd prefer to use default her for both stack and storage, getting rid of the option

mattsse (changes requested):
only looked at syntax

ethers-providers/src/transports/ipc.rs
```diff
@@ -236,6 +236,7 @@ where
     ) -> Result<(), IpcError> {
         // Extend buffer of previously unread with the new read bytes
         read_buffer.extend_from_slice(&bytes);
+        dbg!(read_buffer.len());
```
gakonst, line 239:
```suggestion
```

gakonst (changes requested):

ethers-core/src/types/trace/filter.rs
frjnn: Done.

PR #566, opened 2021-11-08
Prestwich/super pending

prestwich:
## Motivation

Provide a reasonable and working alternative to gas escalator middleware. 

## Solution

- Introduce a new Future `EscalatingPending` that broadcasts a series of transactions
    - functions similarly to `PendingTransaction` but with a simpler (but less informative) state machine
    - it returns the _first_ receipt to confirm.
    - it also returns the _first_ RPC error to occur
    - it allows users to configure polling and broadcast intervals
- Add a middleware method `send_escalating` 
    - allows users to instantiate with a TypedTransaction and an escalation policy
    - may be better as a middleware extension trait?

## things I don't like
- Invariant enforcement is done in the Middleware method rather than the constructor. 
    - This is brittle
    - it's hard to see how to sign and nonce correctly otherwise
- Naming

## PR Checklist

- [ ] Added Tests
- [X] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -93,6 +93,7 @@ use std::{error::Error, fmt::Debug, future::Future, pin::Pin, str::FromStr};
 
 pub use provider::{FilterKind, Provider, ProviderError};
 
+/// A simple gas escalation policy
 pub type EscalationPolicy = Box<dyn Fn(U256, usize) -> U256 + Send + Sync>;
 
 // Helper type alias
@@ -292,9 +293,9 @@ pub trait Middleware: Sync + Send + Debug {
     /// Send a transaction with a simple escalation policy.
     ///
     /// `escalation` should be a boxed function that maps `original_gas_price`
-    /// and `number_of_previous_escalations` -> `new_gas_price`
+    /// and `number_of_previous_escalations` -> `new_gas_price`.
     ///
-    /// e.g. ```Box::new(|start, escalations| start * 1250.pow(escalations) / 1000.pow(escalations))```
+    /// e.g. `Box::new(|start, escalations| start * 1250.pow(escalations) / 1000.pow(escalations))`
     ///
     async fn send_escalating<'a>(
         &'a self,
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -296,7 +296,6 @@ pub trait Middleware: Sync + Send + Debug {
     /// and `number_of_previous_escalations` -> `new_gas_price`.
     ///
     /// e.g. `Box::new(|start, escalations| start * 1250.pow(escalations) / 1000.pow(escalations))`
-    ///
     async fn send_escalating<'a>(
         &'a self,
         tx: &TypedTransaction,
@@ -307,18 +306,17 @@ pub trait Middleware: Sync + Send + Debug {
         self.fill_transaction(&mut original, None).await?;
         let gas_price = original.gas_price().expect("filled");
         let chain_id = self.get_chainid().await?.low_u64();
-        let reqs: Vec<_> = (0..5)
+        let sign_futs: Vec<_> = (0..5)
             .map(|i| {
                 let new_price = escalation(gas_price, i);
                 let mut r = original.clone();
                 r.set_gas_price(new_price);
                 r
             })
+            .map(|req| async move {
+                self.sign(req.rlp(chain_id), from).await.map(|sig| req.rlp_signed(chain_id, &sig))
+            })
             .collect();
-
-        let sign_futs = reqs.into_iter().map(|req| async move {
-            self.sign(req.rlp(chain_id), from).await.map(|sig| req.rlp_signed(chain_id, &sig))
-        });
 
         // we reverse for convenience. Ensuring that we can always just
         // `pop()` the next tx off the back later
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -64,6 +64,7 @@
 //! # }
 //! ```
 mod transports;
+use futures_util::future::join_all;
 pub use transports::*;
 
 mod provider;
@@ -73,6 +74,9 @@ pub mod ens;
 
 mod pending_transaction;
 pub use pending_transaction::PendingTransaction;
+
+mod pending_escalator;
+pub use pending_escalator::EscalatingPending;
 
 mod stream;
 pub use futures_util::StreamExt;
@@ -88,6 +92,8 @@ use serde::{de::DeserializeOwned, Deserialize, Deserializer, Serialize};
 use std::{error::Error, fmt::Debug, future::Future, pin::Pin, str::FromStr};
 
 pub use provider::{FilterKind, Provider, ProviderError};
+
+pub type EscalationPolicy = Box<dyn Fn(U256, usize) -> U256 + Send + Sync>;
 
 // Helper type alias
 #[cfg(target_arch = "wasm32")]
@@ -281,6 +287,44 @@ pub trait Middleware: Sync + Send + Debug {
         block: Option<BlockId>,
     ) -> Result<PendingTransaction<'_, Self::Provider>, Self::Error> {
         self.inner().send_transaction(tx, block).await.map_err(FromErr::from)
+    }
+
+    /// Send a transaction with a simple escalation policy.
+    ///
+    /// `escalation` should be a boxed function that maps `original_gas_price`
+    /// and `number_of_previous_escalations` -> `new_gas_price`
+    ///
+    /// e.g. ```Box::new(|start, escalations| start * 1250.pow(escalations) / 1000.pow(escalations))```
+    ///
+    async fn send_escalating<'a>(
+        &'a self,
+        tx: &TypedTransaction,
+        from: &Address,
+        escalation: EscalationPolicy,
+    ) -> Result<EscalatingPending<'a, Self::Provider>, Self::Error> {
+        let mut original = tx.clone();
+        self.fill_transaction(&mut original, None).await?;
+        let gas_price = original.gas_price().expect("filled");
+        let chain_id = self.get_chainid().await?.low_u64();
+        let reqs: Vec<_> = (0..5)
+            .map(|i| {
+                let new_price = escalation(gas_price, i);
+                let mut r = original.clone();
+                r.set_gas_price(new_price);
+                r
+            })
+            .collect();
+
+        let sign_futs = reqs.into_iter().map(|req| async move {
+            self.sign(req.rlp(chain_id), from).await.map(|sig| req.rlp_signed(chain_id, &sig))
+        });
+
+        // we reverse for convenience. Ensuring that we can always just
+        // `pop()` the next tx off the back later
+        let mut signed = join_all(sign_futs).await.into_iter().collect::<Result<Vec<_>, _>>()?;
+        signed.reverse();
+
+        Ok(EscalatingPending::new(self.provider(), signed))
     }
 
     async fn resolve_name(&self, ens_name: &str) -> Result<Address, Self::Error> {
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -299,7 +299,6 @@ pub trait Middleware: Sync + Send + Debug {
     async fn send_escalating<'a>(
         &'a self,
         tx: &TypedTransaction,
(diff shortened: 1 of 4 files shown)


ethers-providers/src/pending_escalator.rs
```diff
@@ -0,0 +1,198 @@
...
+            }
+            Sleeping(fut) => {
+                if fut.as_mut().poll(cx).is_ready() {
+                    // if timer has elapsed (or this is the first tx)
+                    if this.last.is_none()
+                        || this.last.clone().unwrap().elapsed() > *this.broadcast_interval
+                    {
+                        // then if we have a TX to broadcast, start
+                        // broadcasting it
+                        if let Some(next_to_broadcast) = this.txns.pop() {
```
ltchang2019, line 151: So if we have already broadcast the most escalated tx, we will just wait on that one indefinitely?

ethers-providers/src/pending_escalator.rs
```diff
@@ -0,0 +1,198 @@
...
+
+                // otherwise drain one and check if we have a receipt
+                let mut pollee = futs.pop().expect("checked");
+                match pollee.as_mut().poll(cx) {
+                    //
+                    Poll::Ready(Ok(Some(receipt))) => {
+                        completed!(this, Ok(receipt));
+                    }
+                    // rewake until drained
+                    Poll::Ready(Ok(None)) => cx.waker().wake_by_ref(),
```
ltchang2019, line 181: Is this `cx.waker().wake_by_ref()` necessary? Won't executor automatically put the EscalatingPending future back in queue? Other paths here that return Poll::Pending don't call wake

prestwich: the goal is to explicitly notify the executor that this is ready to be polled again _immediately_. i.e. override the executor's scheduling and jump to the front of the queue as much as we have the ability to, as we know that there's no async operation happening to wait on

ethers-providers/src/pending_escalator.rs
prestwich: yes

ethers-providers/src/lib.rs
```diff
@@ -283,6 +290,42 @@ pub trait Middleware: Sync + Send + Debug {
...
+    /// Send a transaction with a simple escalation policy.
+    ///
+    /// `escalation` should be a boxed function that maps `original_gas_price`
+    /// and `number_of_previous_escalations` -> `new_gas_price`.
+    ///
+    /// e.g. `Box::new(|start, escalations| start * 1250.pow(escalations) / 1000.pow(escalations))`
+    async fn send_escalating<'a>(
+        &'a self,
+        tx: &TypedTransaction,
+        from: &Address,
```
gakonst, line 302: You can use `self.default_sender` instead of `from`

ethers-providers/src/lib.rs
```diff
@@ -283,6 +290,42 @@ pub trait Middleware: Sync + Send + Debug {
...
+        escalation: EscalationPolicy,
```
gakonst, line 303: Maybe this should be a reference?

ethers-providers/src/lib.rs
```diff
@@ -283,6 +290,42 @@ pub trait Middleware: Sync + Send + Debug {
...
+    ) -> Result<EscalatingPending<'a, Self::Provider>, Self::Error> {
+        let mut original = tx.clone();
+        self.fill_transaction(&mut original, None).await?;
+        let gas_price = original.gas_price().expect("filled");
+        let chain_id = self.get_chainid().await?.low_u64();
+        let sign_futs: Vec<_> = (0..5)
```
gakonst, line 309: Should the range be part of `EscalationPolicy`?

ethers-providers/src/lib.rs
```diff
@@ -283,6 +290,42 @@ pub trait Middleware: Sync + Send + Debug {
...
+            .map(|i| {
+                let new_price = escalation(gas_price, i);
+                let mut r = original.clone();
+                r.set_gas_price(new_price);
+                r
+            })
+            .map(|req| async move {
+                self.sign(req.rlp(chain_id), from).await.map(|sig| req.rlp_signed(chain_id, &sig))
```
gakonst, line 317: Is this really bad UX? Should we change it to something better?

gakonst (commented):
some minor qs, love the direction

ethers-providers/src/lib.rs
prestwich: I'm considering making this a trait, as we really want something that generates an arbitrary number txns instead of always 5. we can't do an unbounded number due to the constraints on pre-signing before instantiating the escalating tho

ethers-providers/src/lib.rs
prestwich: `default_sender()` returns `Option<H160>` what should be the behavior in the `None` case?

ethers-providers/src/lib.rs
prestwich: a shortcut function for signing transactions might be good?

ethers-providers/src/lib.rs
prestwich: yes, but haven't gotten there yet

ethers-providers/src/lib.rs
prestwich: I guess `unwrap_or_default`. that'll cause clear errors when we join_all later

ethers-providers/src/pending_escalator.rs
```diff
@@ -0,0 +1,193 @@
...
+        use PendingStates::*;
+
+        let this = self.project();
+
+        match this.state {
+            Initial(fut) => {
+                broadcast_checks!(cx, this, fut);
+            }
+            Sleeping(instant) => {
+                if instant.elapsed() > *this.polling_interval {
```
prestwich, line 139: @gakonst please check this pattern out, as I am unclear why we use `Delay` futures elsewhere. is there some significant behavior difference I'm missing?

mattsse: the futures_timer Delay type uses an additional global timerhandle thread that makes sure the delayed task gets woken up at the right time, however, futures_timer hasn't been updated in a while there's also tokio::time::Sleep now, which I guess works similarly but probably uses some Runtime Handle instead.

looking at the Sleeping state, there might be a case were the Future returns Pending and gets not woken up again (elpased < polling_interval)

ethers-providers/src/pending_escalator.rs
mattsse: this section looks like it could be wrapped in a loop over all futures? So that `Poll::Ready(Ok(None)) =>` merely results in a continue?
Is the order in which they should be polled here important, if not then `Vec::swap_remove` could be useful.

mattsse (commented):
I feel like this could even be two different types.
An additional `Stream` type that processes all transactions.
The `EscalatingPending` Future that polls the underlying stream and returns the first value?

ethers-providers/src/pending_escalator.rs
prestwich: `tokio::time::Sleep` won't work in WASM iirc

I believe all common executors schedule futures for polling in frames regardless of whether the waker has explicitly requested waking

prestwich: didn't finish the previous thought before submitting the tx

given that this could cause issues with uncommon executors, should we audit all our futures to ensure that all polls either request rewaking _or_ poll another future?

ethers-providers/src/pending_escalator.rs
prestwich: I mean it kinda is right now, it's just ensuring that it allows the executor to schedule each step of the loop and interleave other work

prestwich:
test failure is unrelated to these changes

@gakonst @mattsse do you think that this belongs on the main middleware or on some sort of extension trait?

mattsse:
> @gakonst @mattsse do you think that this belongs on the main middleware or on some sort of extension trait?

What, could be nice if this could be somehow integrated with the current `send_transaction -> PendingTx`, so that we can do something like `p.send_transaction(tx).escalate(||...).await` and `p.send_transaction(tx).await.await` still works.

I believe this would be possible with an additional future type (`SendTx` or something?) that would have `Output = PendingTransaction` and a `fn escalate`. But this would mean, `async fn send_transaction` would become `fn send_transaction() -> SendTx`.

this would be a breaking change however.

The easiest way would probably be to add another provider function, as is the case now.

prestwich:
the problem with that is that the signing _must_ be done in the top-level middleware. It _can't_ be done by the provider that the pending transaction has access to. So that feature is blocked by an object-safe `Middleware` trait

mattsse:
ah I see, makes sense, this would then be limited to only the SignerMiddleware...

prestwich:
Not exactly, as the raw provider may sign in many cases (using eth_sign or similar). However, relying on the provider would result in a conflict _if_ any `SignerMiddleware` were used. 

E.g `SignerMiddleware( Provider )`.  The `SignerMiddleware` may be signing with address `0x1234...` where the node controls address `0xabcd...`. This would cause errors and other unexpected behavior if we choose to have signing in the `Provider` held by a `PendingTransaction`

ethers-providers/src/pending_escalator.rs
```diff
@@ -0,0 +1,221 @@
...
+        }
+    }
+
+    pub fn broadcast_interval(mut self, duration: u64) -> Self {
+        self.broadcast_interval = Duration::from_secs(duration);
+        self
+    }
+
+    pub fn polling_interval(mut self, duration: u64) -> Self {
+        self.polling_interval = Duration::from_secs(duration);
```
gakonst, line 81:
```suggestion
        self.polling_interval = Duration::from_millis(duration);
```

ethers-providers/src/pending_escalator.rs
gakonst: Looping over all futures is an anti-pattern I think, the state machine _may_ be slightly easier to write but as @prestwich says it's better to do it on a per-fut basis

gakonst (commented):
looking good, I am OK with placing it as an extra function as-is. I also liked @mattsse's idea of adding an extra type on top of `PendingTransaction` so that we can do `send_transaction(tx).await?.escalator(...).await?` to escalate and `send_transaction(tx).await?.await?` for simple usage

ethers-providers/src/pending_escalator.rs
mattsse: Isn't the `CheckingReceipts` state essentially a `FuturesOrdered` but we only poll one future at a time?

ethers-providers/src/pending_escalator.rs
```diff
@@ -0,0 +1,221 @@
...
+            // placeholder value. We set this again after the initial broadcast
+            // future resolves
+            last: Instant::now(),
+            sent: vec![],
+            state: EscalatorStates::Initial(Box::pin(provider.send_raw_transaction(first))),
+        }
+    }
+
+    pub fn broadcast_interval(mut self, duration: u64) -> Self {
+        self.broadcast_interval = Duration::from_secs(duration);
```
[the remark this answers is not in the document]
prestwich: going to change it to use `Into<Duration>` to match the behavior of `PendingTransaction`

prestwich:
latest commit uses `FuturesUnordered`. I think this is better as @mattsse says. We _don't_ require ordering here, so this functions similarly to select_all, but is easier to manage

prestwich:
I have no other outstanding changes so I'll update the changelog and this is ready for consideration

gakonst (approved):
I like it. Thank you for the hard work, we can iterate on it post-merge. @mattsse any further comments?

mattsse:
> I like it. Thank you for the hard work, we can iterate on it post-merge. @mattsse any further comments?

lgtm

ethers-providers/src/pending_escalator.rs
```diff
@@ -0,0 +1,226 @@
...
+                    // bubble up errors
+                    Poll::Ready(Some(Err(e))) => {
+                        completed!(this, Err(e));
+                    }
+                    Poll::Ready(None) => {
+                        sleep!(cx, this);
+                    }
+                    // check again later
+                    Poll::Pending => {
+                        // stick it pack in the list for polling again later
```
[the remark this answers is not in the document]
prestwich: ah outdated comment. I'll review a those real quick before merge

gakonst:
gg. Should we receive the escalator middleware?

PR #568, opened 2021-11-09
fix: adds TimeLag::new() method

ltchang2019:
<h1> Motivation </h1>

`ethers:middleware::TimeLag` missing `fn new(...) -> Self`

<h1> Solution </h1>

Adds function to instantiate new `TimeLag`


prestwich (approved):
my bad for not adding this in the first place

gakonst (approved):
lgtm

PR #574, opened 2021-11-12
feat: auto-detect solc remappings

gakonst:
Ports over the logic from [here](https://github.com/gakonst/dapptools-rs/blob/master/dapptools/src/utils.rs#L47)

### DIFF
--- a/ethers-solc/src/error.rs
+++ b/ethers-solc/src/error.rs
@@ -23,6 +23,10 @@ pub enum SolcError {
     #[cfg(feature = "svm")]
     #[error(transparent)]
     SvmError(#[from] svm::SolcVmError),
+    #[error("no contracts found under {0}")]
+    NoContracts(String),
+    #[error(transparent)]
+    PatternError(#[from] glob::PatternError),
 }
 
 impl SolcError {
--- a/ethers-solc/src/remappings.rs
+++ b/ethers-solc/src/remappings.rs
@@ -1,4 +1,5 @@
 use crate::{error::SolcError, Result};
+use serde::{Deserialize, Serialize};
 use std::fmt;
 
 const DAPPTOOLS_CONTRACTS_DIR: &str = "src";
@@ -38,6 +39,34 @@ const JS_CONTRACTS_DIR: &str = "contracts";
 pub struct Remapping {
     pub name: String,
     pub path: String,
+}
+
+impl Serialize for Remapping {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::ser::Serializer,
+    {
+        serializer.serialize_str(&self.to_string())
+    }
+}
+
+impl<'de> Deserialize<'de> for Remapping {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::de::Deserializer<'de>,
+    {
+        let remapping = String::deserialize(deserializer)?;
+        let mut split = remapping.split("=");
+        let name = split
+            .next()
+            .ok_or_else(|| serde::de::Error::custom("no remapping prefix found"))?
+            .to_string();
+        let path = split
+            .next()
+            .ok_or_else(|| serde::de::Error::custom("no remapping path found"))?
+            .to_string();
+        Ok(Remapping { name, path })
+    }
 }
 
 // Remappings are printed as `prefix=target`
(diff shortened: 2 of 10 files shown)


ethers-solc/src/remappings.rs
```diff
@@ -0,0 +1,284 @@
...
+                std::fs::create_dir_all(&path).unwrap();
+            }
+        }
+    }
+
+    // helper function for converting path bufs to remapping strings
+    fn to_str(p: std::path::PathBuf) -> String {
+        let mut s = p.into_os_string().into_string().unwrap();
+        s.push('/');
+        s
```
mattsse, line 185:
```suggestion
        format!("{}/", p.display())
```

ethers-solc/src/remappings.rs
```diff
@@ -0,0 +1,284 @@
...
+        } else {
+            format!("{}/{}/**/*.sol", name, source)
+        };
+        let mut dapptools_contracts = glob::glob(&pattern)?;
+        let next = dapptools_contracts.next();
+        if next.is_some() {
+            let path = format!("{}/{}/", name, source);
+            let mut name = name.split('/').last().unwrap().to_string();
+            name.push('/');
+            Ok(Remapping { name, path })
```
mattsse, line 101: since I already pulled in `walkdir`, ref `util::source_files`, I wonder if this can be replaced that. 
We could extract the walkdir iter part to a new function `util::source_files_iter` maybe

ethers-solc/src/remappings.rs
```diff
@@ -0,0 +1,284 @@
...
+        s.push('/');
+        s
+    }
+
+    #[test]
+    fn find_remapping_dapptools() {
+        let tmp_dir = tempdir::TempDir::new("lib").unwrap();
+        let tmp_dir_path = tmp_dir.path();
+        let paths = ["repo1/src/", "repo1/src/contract.sol"];
+        mkdir_or_touch(&tmp_dir_path, &paths[..]);
```
mattsse, line 193:
```suggestion
        mkdir_or_touch(tmp_dir_path, &paths[..]);
```

ethers-solc/src/remappings.rs
```diff
@@ -0,0 +1,284 @@
...
+        }
+    }
+
+    pub fn find_many_str(path: &str) -> Result<Vec<String>> {
+        let remappings = Self::find_many(path)?;
+        Ok(remappings.iter().map(|mapping| format!("{}={}", mapping.name, mapping.path)).collect())
+    }
+
+    /// Gets all the remappings detected
+    pub fn find_many(path: impl AsRef<std::path::Path>) -> Result<Vec<Self>> {
```
mattsse, line 113: it's not instantly clear to me what's going on, but I believe this could also be replaced with a nifty `Walkdir` iterator, the `walkdir::DirEntry` has a `fn depth()` function

ethers-solc/src/remappings.rs
```diff
@@ -0,0 +1,284 @@
...
+impl<'de> Deserialize<'de> for Remapping {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::de::Deserializer<'de>,
+    {
+        let remapping = String::deserialize(deserializer)?;
+        let mut split = remapping.split('=');
+        let name = split
+            .next()
+            .ok_or_else(|| serde::de::Error::custom("no remapping prefix found"))?
```
mattsse, line 63: should this return an error instead?

ethers-solc/src/error.rs
```diff
@@ -23,6 +23,10 @@ pub enum SolcError {
     #[cfg(feature = "svm")]
     #[error(transparent)]
     SvmError(#[from] svm::SolcVmError),
+    #[error("no contracts found under {0}")]
+    NoContracts(String),
```
mattsse, line 27:
```suggestion
    NoContractsFound(String),
```
maybe even PathBuf

mattsse (commented):
some comments/suggestions

ethers-solc/src/remappings.rs
gakonst: As discussed, I tried this and it was not obvious how to do. Let's refactor in a follow-up PR if needed.

ethers-solc/src/remappings.rs
gakonst: already does?

Issue #575, opened 2021-11-12
Build failure on aarch64-unkown-linux-musl cross compile


Recmo:
**Version**

```text
#14 157.5 └── ethers v0.5.4
#14 157.5     ├── ethers-contract v0.5.4
#14 157.5     │   ├── ethers-contract-abigen v0.5.4
#14 157.5     │   │   ├── ethers-core v0.5.5
#14 157.5     │   ├── ethers-contract-derive v0.5.4 (proc-macro)
#14 157.5     │   │   ├── ethers-contract-abigen v0.5.4 (*)
#14 157.5     │   │   ├── ethers-core v0.5.5
#14 157.5     │   ├── ethers-core v0.5.5 (*)
#14 157.5     │   ├── ethers-providers v0.5.5
#14 157.5     │   │   ├── ethers-core v0.5.5 (*)
#14 157.5     ├── ethers-core v0.5.5 (*)
#14 157.5     ├── ethers-middleware v0.5.4
#14 157.5     │   ├── ethers-contract v0.5.4 (*)
#14 157.5     │   ├── ethers-core v0.5.5 (*)
#14 157.5     │   ├── ethers-providers v0.5.5 (*)
#14 157.5     │   ├── ethers-signers v0.5.4
#14 157.5     │   │   ├── ethers-core v0.5.5 (*)
#14 157.5     ├── ethers-providers v0.5.5 (*)
#14 157.5     └── ethers-signers v0.5.4 (*)
```

**Platform**
Docker `linux/arm64` `rust:latest` cross compile to `aarch64-unkown-linux-musl`

`uname -a`: `Linux buildkitsandbox 5.10.47-linuxkit #1 SMP PREEMPT Sat Jul 3 21:50:16 UTC 2021 aarch64 GNU/Linux`

**Description**

<details>
<summary><code>Dockerfile</code></summary>

```Dockerfile
FROM rust
WORKDIR /src

RUN apt-get update &&\
    apt-get install -y texinfo &&\
    apt-get clean && rm -rf /var/lib/apt/lists/* &&\
    cargo install cargo-tree &&\
    rustup target add $(uname -m)-unknown-linux-musl

# Build {x86_64,aarch64}-linux-musl toolchain
# This is required for some build.rs scripts
ARG MUSL_CROSS_VERSION=0.9.9
RUN curl -fL "https://github.com/richfelker/musl-cross-make/archive/v${MUSL_CROSS_VERSION}.tar.gz"\
[...]

PR #580, opened 2021-11-13
feat: make reqwest optional but enabled by default

mattsse:
## Motivation
Provide a way to disable networking for proc-macro, disable request support with `abigen-offline` macro instead.

close. #575 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog


gakonst (approved):
LGTM

PR #581, opened 2021-11-13
Prestwich/escalator fixes

prestwich:
## Motivation

Pending escalation broken by lack of availability of tx signing on middleware

## Solution

Add `Middleware::sign_transaction`. It errors unless there is a `SignerMiddleware` in the stack somewhere. The `from` field has no effect (matching behavior of `SignerMiddleware::sign` but is added to match the `Middleware::sign` API for potential future use.

driveby: while testing escalation locally, hit an edge case with tx braodcast failing due to a previous escalation confirming during broadcast. Added a handler for that in the `Broadcasting` state transition

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-middleware/src/signer.rs
+++ b/ethers-middleware/src/signer.rs
@@ -167,6 +167,14 @@ where
         true
     }
 
+    async fn sign_transaction(
+        &self,
+        tx: &TypedTransaction,
+        _: Address,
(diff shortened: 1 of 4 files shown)


gakonst (approved):
LGTM

PR #587, opened 2021-11-16
fix: remove generic const from TimeLag

ltchang2019:
## Motivation
TimeLag middleware parametrizes the lag amount (number of blocks to lag behind) as generic constant. This prevents devs from being able to instantiate the TimeLag's lag amount with a runtime variable.

## Solution
Replaces the generic const with regular struct field.


prestwich (approved):
Yeh, this is better. I was mostly using const generic here to get a feel for the system.

PR #590, opened 2021-11-17
Fix/providers

ed255:
## Motivation

While reviewing the ethers-providers crate I noticed two small details that could be easily fixed.

The first one is a data race when generating a new id for the http Provider.

The second one is the `JsonRpcClient` trait requiring `R` (the Response) to be `Serialize`, when in fact the Response should only require `Deserialize`.  When writing client only code, we shouldn't need to define serializers for reponse types.

## Solution

For the data race I just copied the new id generation from the Websocket implementation.

For the unnecessary `Serialize` in the `JsonRpcClient` Response, I just removed it.

## PR Checklist

- [x] Added Tests
    - I didn't add tests because the data race is hard to test, and the removal of the `Serialize` in the `JsonRpcClient` response can be already verified to work with the current tests.
- [x] Added Documentation
    - I didn't see the need to add documentation for these changes
- [x] Updated the changelog


prestwich:
related: #217.  Looks like I missed this `load` when cleaning up earlier

gakonst (approved):
LGTM

PR #593, opened 2021-11-18
fix: contract version detection must respect imports

gakonst:
Currently, when detecting the Solc version in a contract, we only parse its own pragma import and use the latest version available for that constraint. 

However, a contract may also import other contracts which may restrict the pragma import.

An example can be seen in thet `test-data` folder where while `Child` would work with any compiler `>=0.7.2`, the rdependency on `Mother` which is `<=0.7.4` and `GrandParent` which is `^0.7.0` should restrict it to 0.7.4.

Before this PR, this restriction was not respected, and it'd try to use `0.8.10` (the currently latest version of the compiler)


gakonst:
closing until we decide how to tackle this

PR #597, opened 2021-11-19
Fix parse units

x3ccd4828:
## Motivation
parse_units and format_units were not behaving as expected when the decimal number was greater than 15 such as ether amounts

## Solution
modified parse_units and format_units to use rug::Float of 128bit such that 18 decimal point floats don't get truncated. format_units now returns a rug::Float so that might break code that is using this function however it's an easy fix for such projects.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Add EIP-712 `sign_typed_data` signer method; add ethers-core type `Eip712`
   trait and derive macro in ethers-derive-eip712
   [#481](https://github.com/gakonst/ethers-rs/pull/481)
+- Fix `format_units` to return a `String` of representing a decimal point float such that the decimal places don't get truncated. [597](https://github.com/gakonst/ethers-rs/pull/597)
 
 ### 0.5.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## ethers-core
 
+- Fix `format_units` to return a `String` of representing a decimal point float
+  such that the decimal places don't get truncated.
+  [597](https://github.com/gakonst/ethers-rs/pull/597)
+
 ### Unreleased
 
 ### 0.6.0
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm, since this is a breaking change, could you make a note in the CHANGELOG and.
a basic example in the function's rustdoc could also be useful, basically copy-paste from test and maybe show how to convert back to U256?

x3ccd4828:
@mattsse I will update the changelog and rust doc when I get back to my computer.

x3ccd4828:
@mattsse it seems the WASM build is broken because of the rug crate https://github.com/gakonst/ethers-rs/runs/4266231384?check_suite_focus=true 

is this a problem?

gakonst (changes requested):
I think the requested change should fix the wasm build

x3ccd4828:
@gakonst @mattsse it seems the default-features = false doesn't help the wasm compile

mattsse:
uh oh, this appears to be an issue with gmp-mpfr-sys that's not wasm compatible? https://gitlab.com/tspiteri/gmp-mpfr-sys/-/blob/master/build.rs 
it requires the force-cross feature apparently, but I'm not sure whether this would work for the wasm32 target, can you try this and adding a wasm test with wasm-bindgen https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/usage.html?

x3ccd4828:
yeah I will look into it

x3ccd4828:
it seems the gmp-mpfr-sys lib doesn't support wasm https://gitlab.com/tspiteri/gmp-mpfr-sys/-/issues/18

should I look for a different lib for float?

gakonst:
Yes I'd prefer if we switched to a different float lib! Also, the function should still return `U256`, because it's expected that somebody uses it with other eth calls, and we want to keep the UX good (i.e. one should not need to manually cast to u256 after using these functions)

x3ccd4828:
@gakonst I will look into a different implementation that supports wasm however the functions should return a float or a string, not a U256 since that is incorrect functionality. That would mean that all values less than 0 would be returned as 0. 

so 0.56749 ETH would be returned as 0 which in my opinion is broken.

the ethers implementation returns a string which is a float number: https://docs.ethers.io/v5/api/utils/display-logic/#utils-formatUnits

x3ccd4828:
@gakonst @mattsse I rebased the branch based on the latest master and I only modified format_units to return a String which preserves the decimal places. This should be a good solution until a better float128 lib is available to support wasm.

x3ccd4828:
I rebased to fix the conflict

gakonst (approved):
Looking good.

PR #599, opened 2021-11-20
feat: add cargo change detection support

mattsse:
## Motivation
Add support to enable cargo change detection in build scripts 
https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath

This is useful since cargo may be unaware of changed contracts and won't rerun the build script if there aren't any changes in the rust files.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/lib.rs
+++ b/ethers-solc/src/lib.rs
@@ -115,6 +115,33 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
         Source::read_all_from(self.paths.sources.as_path())
     }
 
+    /// This emits the cargo [`rerun-if-changed`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath) instruction.
+    /// Which tells Cargo to re-run the build script if a file inside the project's sources
+    /// directory has changed.
+    ///
+    /// Use this if you compile a project in a `build.rs` file.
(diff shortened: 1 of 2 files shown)


ethers-solc/src/lib.rs
```diff
@@ -115,6 +115,33 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
...
+    ///
+    /// fn main() {
+    ///     // configure the project with all its paths, solc, cache etc.
+    ///     let project = Project::builder()
+    ///         .paths(ProjectPathsConfig::hardhat(env!("CARGO_MANIFEST_DIR")).unwrap())
+    ///         .build()
+    ///         .unwrap();
+    ///     let output = project.compile().unwrap();
+    ///     // Tell Cargo that if a source file changes, to rerun this build script.
+    ///     project.rerun_if_sources_changed();
```
gakonst, line 138: Should this be before or after the line above? Or does the order not matter?

gakonst (approved):

ethers-solc/src/lib.rs
mattsse: Does not really matter since the script would panic if `project.compile` fails.

PR #600, opened 2021-11-21
fix: lowercase node client type

gakonst:
Otterscan's Erigon fork returns `erigon` and not `Erigon` so we lowercase the node client return


gakonst:
superseded by #601

univerz:
#601 is not a full revert, just separation of those methods. i expected this PR merged & assumed that (correct) implementation of `client_version` is still useful to have.

PR #606, opened 2021-11-22
add support for polygon and avalanche

casualjim:
## Motivation

I want to use this library to interact with polygon and avalanche too. 

## Solution

I've added chain definitions for the networks and testnets, as well as block explorer API's 
I couldn't find a mutlicall contract on avalanche yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 .vscode
+/.envrc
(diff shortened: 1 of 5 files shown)


gakonst (approved):
Looking good! Might also make sense to update the library's README to reflect that we support all EVM chains, and not just Ethereum/Celo, hinting at usage of the `legacy` flag for non-EIP1559 chains.

PR #613, opened 2021-11-23
fix(codec)!: ambiguity unit8[] and bytes

mattsse:
## Motivation
* There was an ambiguity for how to tokenized `Vec<u8>` since it was used for `bytes` and `unit8[]` alike. 

## Solution
* solidity `bytes` is `ethers::types::Bytes` and
* solidity `unit8[]` is `Vec<u8>`

This is a breaking change but can we fixed easily with `into()` conventions

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog


ethers-core/src/abi/mod.rs
```diff
@@ -117,7 +117,7 @@ macro_rules! impl_abi_type {
 }
 
 impl_abi_type!(
-    Vec<u8> => Bytes,
+    Vec<u8> =>  Array(Box::new(ParamType::Uint(8))),
```
gakonst, line 120: I think this won't work, because `bytes memory` which translates to `Vec<u8>` will now be translate to `Array(Uint(8))` as a `uint8[]`.

Could we make `bytes memory` translate to `ethers::core::Bytes`?

(I think that's probably the bug here, given the tests' output)

gakonst (changes requested):

gakonst (commented):
LGTM

PR #614, opened 2021-11-23
feat: verify checksum before project compilation

roynalnaruto:
## Motivation

We have noticed scenarios where an ongoing operation was terminated, leading to a corrupted `solc` binary. If this binary (with the appropriate name) is considered OK and used to compile contracts, we end up with errors that are hard to debug.

## Solution

The proposed solution by @odyslam and @gakonst was to:
- [x] Verify the SHA256 checksum of the `solc` binary against the checksum [found remotely](https://binaries.soliditylang.org/macosx-amd64/list.json). In case of mismatch, re-install that version of `solc` before proceeding.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/compile.rs
+++ b/ethers-solc/src/compile.rs
@@ -166,8 +166,7 @@ impl Solc {
         // load the local / remote versions
         let versions = svm::installed_versions().unwrap_or_default();
         let local_versions = Self::find_matching_installation(&versions, &sol_version);
-        let (_releases, sorted_releases) = RELEASES.clone();
-        let remote_versions = Self::find_matching_installation(&sorted_releases, &sol_version);
+        let remote_versions = Self::find_matching_installation(&RELEASES.1, &sol_version);
 
         // if there's a better upstream version than the one we have, install it
         Ok(match (local_versions, remote_versions) {
@@ -244,8 +243,7 @@ impl Solc {
         hasher.update(&content);
         let checksum_calc = &hasher.finalize()[..];
 
-        let (all_releases, _sorted_releases) = RELEASES.clone();
-        let checksum_found = all_releases.get_checksum(&version).expect("checksum not found");
+        let checksum_found = &RELEASES.0.get_checksum(&version).expect("checksum not found");
 
         if checksum_calc == checksum_found {
             Ok(())
--- a/ethers-solc/src/compile.rs
+++ b/ethers-solc/src/compile.rs
@@ -46,16 +46,20 @@ static LOCK: Lazy<Mutex<()>> = Lazy::new(|| Mutex::new(()));
 #[cfg(all(feature = "svm", feature = "async"))]
 /// A list of upstream Solc releases, used to check which version
 /// we should download.
-pub static RELEASES: Lazy<Vec<Version>> = Lazy::new(|| {
+pub static RELEASES: Lazy<(svm::Releases, Vec<Version>)> = Lazy::new(|| {
     // Try to download the releases, if it fails default to empty
     match tokio::runtime::Runtime::new()
         .expect("could not create tokio rt to get remote releases")
         // TODO: Can we make this future timeout at a small time amount so that
         // we do not degrade startup performance if the consumer has a weak network?
-        .block_on(svm::all_versions())
+        .block_on(svm::all_releases(svm::platform()))
     {
-        Ok(inner) => inner,
-        Err(_) => Vec::new(),
+        Ok(releases) => {
+            let mut sorted_releases = releases.releases.keys().cloned().collect::<Vec<Version>>();
+            sorted_releases.sort();
+            (releases, sorted_releases)
+        },
+        Err(_) => (svm::Releases::default(), Vec::new()),
     }
 });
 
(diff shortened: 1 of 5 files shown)


ethers-solc/src/compile.rs
```diff
@@ -225,6 +225,30 @@ impl Solc {
...
+        let mut version_path = svm::version_path(version.to_string().as_str());
+        version_path.push(format!("solc-{}", version.to_string().as_str()));
+        let content = std::fs::read(version_path)?;
+
+        use sha2::Digest;
+        let mut hasher = sha2::Sha256::new();
+        hasher.update(&content);
+        let checksum_calc = &hasher.finalize()[..];
+
+        let all_releases = tokio::runtime::Runtime::new().unwrap().block_on(svm::all_releases(svm::platform()))?;
```
gakonst, line 242: Should probably use [`RELEASES`](https://github.com/gakonst/ethers-rs/blob/8e09715125f4aad77f43172fde02a83d2d91af2f/ethers-solc/src/compile.rs#L49) instead?

roynalnaruto: I was trying to get rid of this duplication. Actually `RELEASES` fetches only a sorted `Vec<Version>` (basically the `svm::all_versions`).

What I used here is `svm::all_releases` which also provides the build info (checksum info).

gakonst: In any case, it'd be nice if we cached the response in a static or in `self` so that we don't need to query it each time?

mattsse: yeah, I think a global Lazy would be appropriate here, within the lazy init we can use this pattern here to get the runtime https://github.com/gakonst/foundry/blob/432e91196138f63289be40424599869f4d60f208/evm-adapters/src/blocking_provider.rs#L17-L20

ethers-solc/src/lib.rs
```diff
@@ -194,15 +197,22 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
...
         let mut compiled =
             ProjectCompileOutput::with_ignored_errors(self.ignored_error_codes.clone());
 
         // run the compilation step for each version
         for (solc, sources) in sources_by_version {
+            // verify that this solc version's checksum matches the checksum found remotely
+            let version = solc_versions.get(&solc.solc).unwrap();
+            while let Err(_e) = solc.verify_checksum() {
+                Solc::blocking_install(version)?;
+            }
```
mattsse, line 214: what's the reasoning for the loop here, is this just to retry?

ethers-solc/src/compile.rs
roynalnaruto: Yup, gonna refactor that part. Sounds good!

ethers-solc/src/lib.rs
roynalnaruto: Yea, its just to retry. The `blocking_install` would fail if the checksum didn't match on download. So it might actually be unnecessary. Would you suggest removing it?

I actually feel, in future if a bug creeps in, it will be extremely hard to debug this (as we would go into an infinite loop)

ethers-solc/src/compile.rs
```diff
@@ -162,7 +166,8 @@ impl Solc {
         // load the local / remote versions
         let versions = svm::installed_versions().unwrap_or_default();
         let local_versions = Self::find_matching_installation(&versions, &sol_version);
-        let remote_versions = Self::find_matching_installation(&RELEASES, &sol_version);
+        let (_releases, sorted_releases) = RELEASES.clone();
+        let remote_versions = Self::find_matching_installation(&sorted_releases, &sol_version);
```
gakonst, line 170: Do we need the `clone`? Since it's by reference, can't we do `&RELEASES.1` or something instead of `&sorted_releases`?

ethers-solc/src/compile.rs
```diff
@@ -225,6 +230,30 @@ impl Solc {
...
+        version_path.push(format!("solc-{}", version.to_string().as_str()));
+        let content = std::fs::read(version_path)?;
+
+        use sha2::Digest;
+        let mut hasher = sha2::Sha256::new();
+        hasher.update(&content);
+        let checksum_calc = &hasher.finalize()[..];
+
+        let (all_releases, _sorted_releases) = RELEASES.clone();
+        let checksum_found = all_releases.get_checksum(&version).expect("checksum not found");
```
gakonst, line 248: Same question here about `RELEASES.0.get_checksum`

gakonst (commented):
Looking good, pending using new release of svm.

ethers-solc/src/compile.rs
roynalnaruto: Yea that sounds more reasonable.

ethers-solc/src/lib.rs
```diff
@@ -194,15 +197,23 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
...
 
         let mut compiled =
             ProjectCompileOutput::with_ignored_errors(self.ignored_error_codes.clone());
 
         // run the compilation step for each version
         for (solc, sources) in sources_by_version {
+            // verify that this solc version's checksum matches the checksum found remotely. If
+            // not, re-install the same version.
+            let version = solc_versions.get(&solc.solc).unwrap();
+            if let Err(_e) = solc.verify_checksum() {
```
gakonst, line 213: For a later PR, we should add a tracing log here that it found a checksum mismatch

gakonst (approved):
Looking good.

gakonst (approved):
woohooo

PR #621, opened 2021-11-25
feat(etherscan, middleware): implement gas endpoints and use in oracle middleware

shekhirin:
## Motivation

https://github.com/gakonst/ethers-rs/issues/485

## PR Checklist

- [x] Add free tier `gastracker/gas_estimate` and `gastracker/gas_oracle` methods for corresponding endpoints
- [x] Use `Etherscan::gas_oracle` for oracle middleware
- [x] Run at most 1 etherscan test per 250ms and make all etherscan tests serial instead of parallel so we don't get rate limited
- [x] Add separate etherscan API token for Celo tests so we don't get rate limited


gakonst (approved):
nice work.

PR #623, opened 2021-11-26
fix: pass partial artifact cache to project compiler output

korboismoe:
## Motivation

If the cache contains some artifacts, but not all, and there are changed files, then the project compiler output does not contain the cached artifacts.

This means when calling `find` or `into_artifacts`, the cached artifacts would not be present in `artifacts`. 

For example, if you run a compile, then create a new contract and run compile again, the `ProjectCompileOutput` 
will not contain the cached artifacts and you have read the cache manually in order to get them. The provided
test covers this scenario. 

This also results in some strange behavior when you run compile multiple times, as the previous cache is overwritten
with a new partial cache covering only the new files.

This originated from using foundry and seeing the above issue when running a newly created test and having it flip flop
between the old and new tests.
If I misunderstood how this works, or this is the intended behavior of the library and if cache reading should be done manually, then feel free to close this PR.

## Solution

If any cached artifacts exists, they are now returned as part of the project compiler output. Any missing files are removed from the cache.

## PR Checklist

- [x] Added Tests
- [x] Updated the changelog

### DIFF
--- a/ethers-solc/src/lib.rs
+++ b/ethers-solc/src/lib.rs
@@ -242,13 +242,13 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
         // If there's a cache set, filter to only re-compile the files which were changed
         let (sources, cached_artifacts) = if self.cached && self.paths.cache.exists() {
             let mut cache = SolFilesCache::read(&self.paths.cache)?;
+            cache.remove_missing_files();
             let changed_files = cache.get_changed_or_missing_artifacts_files::<Artifacts>(
                 sources,
                 Some(&self.solc_config),
                 &self.paths.artifacts,
             );
 
-            cache.remove_missing_files();
             let cached_artifacts = cache
                 .read_artifacts::<Artifacts>(&self.paths.artifacts)
                 .unwrap_or_else(|_| BTreeMap::default());
--- a/ethers-solc/src/lib.rs
+++ b/ethers-solc/src/lib.rs
@@ -249,9 +249,11 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
                 &self.paths.artifacts,
             );
 
-            let cached_artifacts = cache
-                .read_artifacts::<Artifacts>(&self.paths.artifacts)
-                .unwrap_or_else(|_| BTreeMap::default());
+            let cached_artifacts = if self.paths.artifacts.exists() {
+                cache.read_artifacts::<Artifacts>(&self.paths.artifacts)?
+            } else {
+                BTreeMap::default()
+            };
             // if nothing changed and all artifacts still exist
             if changed_files.is_empty() {
                 return Ok(ProjectCompileOutput::from_unchanged(cached_artifacts))
--- a/ethers-solc/src/lib.rs
+++ b/ethers-solc/src/lib.rs
@@ -1,4 +1,4 @@
-#![doc = include_str!("../README.md")]
+#![doc = include_str ! ("../README.md")]
 
 pub mod artifacts;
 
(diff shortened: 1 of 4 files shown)


ethers-solc/src/lib.rs
```diff
@@ -237,22 +240,26 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
...
-            let cache = SolFilesCache::read(&self.paths.cache)?;
+        let (sources, cached_artifacts) = if self.cached && self.paths.cache.exists() {
+            let mut cache = SolFilesCache::read(&self.paths.cache)?;
             let changed_files = cache.get_changed_or_missing_artifacts_files::<Artifacts>(
                 sources,
                 Some(&self.solc_config),
                 &self.paths.artifacts,
             );
 
+            cache.remove_missing_files();
```
mattsse, line 251: nice catch! But we probably want to call this before `get_changed_or_missing_artifacts_files`, right?

mattsse:
> If the cache contains some artifacts, but not all, and there are changed files, then the project compiler output does not contain the cached artifacts.

That sounds right to me, that means, in that case, we need to merge the compiler output with the cached artifacts.

ethers-solc/tests/project.rs
```diff
@@ -75,3 +78,67 @@ fn can_compile_dapp_sample() {
...
+    std::fs::copy(new_file, root.join("src/NewContract.sol")).unwrap();
+    let compiled = project.compile().unwrap();
+    assert!(compiled.find("Dapp").is_some());
+    assert!(compiled.find("NewContract").is_some());
+    assert!(!compiled.is_unchanged());
+
+    // deleted artifact is not taken from the cache
+    std::fs::remove_file(&project.paths.sources.join("Dapp.sol")).unwrap();
+    let compiled = project.compile().unwrap();
+    assert!(compiled.find("Dapp").is_none());
```
mattsse, line 129: this verifies that we clear deleted files from the cache

ethers-solc/tests/project.rs
```diff
@@ -75,3 +78,67 @@ fn can_compile_dapp_sample() {
...
+    let compiled = project.compile().unwrap();
+    assert!(compiled.find("Dapp").is_some());
+    assert!(!compiled.is_unchanged());
+
+    // new file is compiled even with partial cache
+    std::fs::copy(new_file, root.join("src/NewContract.sol")).unwrap();
+    let compiled = project.compile().unwrap();
+    assert!(compiled.find("Dapp").is_some());
+    assert!(compiled.find("NewContract").is_some());
+    assert!(!compiled.is_unchanged());
```
mattsse, line 124: This hits `from_compiler_output_and_cache` since we have a cached artifact but need to compile the NewContract.
lgtm

ethers-solc/src/lib.rs
```diff
@@ -237,22 +240,26 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
...
+            let cached_artifacts = cache
+                .read_artifacts::<Artifacts>(&self.paths.artifacts)
+                .unwrap_or_else(|_| BTreeMap::default());
```
mattsse, line 254: This is unlikely to fail since we checked that those files exists, but could still fail due to some race conditions, like the files got wiped after we checked that they exist...

I feel we should propagate the error here because if that would fail, we end up with a successful output that lacks some artifacts.

mattsse (changes requested):
great find, thank you!
this looks reasonable to me and the root cause you identified is the reason for the issue.

some nits.

ethers-solc/src/lib.rs
korboismoe: The main failure case (outside of a race) for the fallback was when the artifacts folder is deleted, but the cache metadata still exists. 
I'll change it to propagate the error, and fall back to empty cache if there is no artifacts folder. This way if some of the files have been wiped, it'll return the error. There's still a chance that the folder gets deleted at the same time still and then there'd be missing artifacts.

gakonst (approved):
Great find!

PR #625, opened 2021-11-26
fix: cache svm rs releases

gakonst:
ref https://github.com/roynalnaruto/svm-rs/pull/7


gakonst:
Closing, the bottleneck was in the `hasher.update` function in checksum verification. Separate PR to follow with a fix.

PR #627, opened 2021-11-26
fix: remove redundant cache read

gakonst:
This seems to be redundant with the code [right below](https://github.com/gakonst/ethers-rs/blob/master/ethers-solc/src/lib.rs#L252-L264)


mattsse (approved):
nice catch, this should be redundant now.

PR #629, opened 2021-11-26
fix(solc): do not overwrite existing cache

gakonst:
previously if 1 file changed, it'd overwrite the existing cache on disk with just
that 1 changed file. this change reads the cache from disk and merges it with the new checksums

ref: https://github.com/gakonst/ethers-rs/pull/623

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1190,6 +1190,7 @@ dependencies = [
  "tempdir",
  "thiserror",
  "tokio",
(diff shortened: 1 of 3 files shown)


korboismoe:
Something minor to note here is that deleted files will remain in the cache metadata file.

gakonst:
Wonder what's the best way to approach this to have a robust cache file

PR #630, opened 2021-11-27
fix(solc): remove changed artifacts from the cache

korboismoe:
## Motivation

When reading from the cache, there may be changed artifacts that need to be recompiled. 
Currently only missing files are removed from the cache metadata struct, but changed files remain.
Since we're now returning the cached artifacts as part of #623 on partial changes and writing new
files to the cache as part of #629, we also need to remove changed files from the cache. 

## Solution

After getting the changed sources, remove them from the cache metadata (`SolFilesCache`) so
they are no longer returned.

## PR Checklist

- [x] Added Tests

### DIFF
--- a/ethers-solc/src/cache.rs
+++ b/ethers-solc/src/cache.rs
@@ -58,6 +58,10 @@ impl SolFilesCache {
         self.files.retain(|file, _| Path::new(file).exists())
     }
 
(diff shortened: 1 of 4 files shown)


gakonst (approved):
LGTM

mattsse (approved):
great find!
lgtm

PR #633, opened 2021-11-29
fix: source parsing for local paths

mattsse:
## Motivation
followup on #631 that prevents path resolution while parsing the path

## Solution
don't canonicalize local paths like `./testdata/console.json` in `Source::parse`, resolving paths for `Source::Local` happens at a later stage at `Source::get`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
we'll get it right eventually

PR #640, opened 2021-12-02
Add dev-rpc middleware

wolflo:
Adds a dev-rpc feature flag to ethers-providers. This gives you a `DevRpcMiddleware` that exposes `evm_snapshot` and `evm_revert` for managing snapshot state.

There are [more](https://github.com/trufflesuite/ganache-cli-archive#custom-methods) methods that may be useful, but there is a bit of a mess to untangle with conflicts between documentation, hardhat, and ganache.

A lighter-weight alternative may be to just expose `provider.request()` for raw requests under a dev-rpc flag.

- [X] Added Tests
- [ ] Added Documentation TODO
- [ ] Updated the changelog TODO

### DIFF
--- a/ethers-providers/src/provider.rs
+++ b/ethers-providers/src/provider.rs
@@ -1094,6 +1094,8 @@ pub mod dev_rpc {
         }
     }
     #[cfg(test)]
+    // Celo blocks can not get parsed when used with Ganache
+    #[cfg(not(feature = "celo"))]
     mod tests {
         use super::*;
         use crate::{Http, Provider};
--- a/ethers-providers/src/provider.rs
+++ b/ethers-providers/src/provider.rs
@@ -1087,9 +1087,10 @@ pub mod dev_rpc {
                 .request::<[U256; 1], bool>("evm_revert", [id])
                 .await
                 .map_err(DevRpcMiddlewareError::ProviderError)?;
-            match ok {
-                true => Ok(()),
-                false => Err(DevRpcMiddlewareError::NoSnapshotError),
+            if ok {
+                Ok(())
+            } else {
+                Err(DevRpcMiddlewareError::NoSnapshotError)
             }
         }
     }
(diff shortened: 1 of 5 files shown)


wolflo:
Not sure what the test failure is, will take a look today. Seems to pass when run with `--features dev-rpc` but fail when run with `--all-features`.

gakonst (approved):
Nice work! Will merge once linted, if you want to do it @wolflo, otherwise i'll try later

ethers-providers/src/provider.rs
```diff
@@ -1021,6 +1021,145 @@ impl TryFrom<String> for Provider<HttpProvider> {
...
+        pub async fn revert_to_snapshot(&self, id: U256) -> Result<(), DevRpcMiddlewareError<M>> {
+            let ok = self
+                .provider()
+                .request::<[U256; 1], bool>("evm_revert", [id])
+                .await
+                .map_err(DevRpcMiddlewareError::ProviderError)?;
+            match ok {
+                true => Ok(()),
+                false => Err(DevRpcMiddlewareError::NoSnapshotError),
+            }
```
mattsse, line 1093: an `if {} else {}`would be more idiomatic here

ethers-providers/src/provider.rs
```diff
@@ -1021,6 +1021,146 @@ impl TryFrom<String> for Provider<HttpProvider> {
...
+    #[derive(Error, Debug)]
+    pub enum DevRpcMiddlewareError<M: Middleware> {
+        #[error("{0}")]
+        MiddlewareError(M::Error),
+
+        #[error("{0}")]
+        ProviderError(ProviderError),
+
+        #[error("Could not revert to snapshot")]
+        NoSnapshot,
```
wolflo, line 1045: Maybe I misunderstood this lint, but it seems to not jive with the rest of the Error naming convention for ethers.

PR #641, opened 2021-12-03
fix(provider::ws): ignore the websocket response when the request has been cancelled

guanqun:
In cases when the request is out, but cancelled locally before it gets back, then later on, the replied request should be ignored, otherwise, it hits an error and aborts the whole websocket connection.


mattsse (approved):
nice catch!

guanqun:
force pushed to fix a fmt issue, didn't notice that it was using a nightly fmt.

PS: Actually the IPC has similar issue.

PR #650, opened 2021-12-04
feat(abigen): support overloaded functions with different casing

### LINKED ISSUE #571: abigen is intolerant of duplicate names with different casing
**Version**
https://github.com/gakonst/ethers-rs#906c504f

**Platform**
linux

**Description**
Abigen is intolerant of identifiers that have the same name but differ in casing.

For example the names `index` and `INDEX` both appear in the sOHM contract: 0x04f2694c8fcee23e8fd0dfea1d4f5bb8c352111f

I tried this code:

```rust
Abigen::new(
    "StakedOHM",
    "https://etherscan.io/address/0x04f2694c8fcee23e8fd0dfea1d4f5bb8c352111f",
  )
  .unwrap()
  .generate()
  .unwrap()
  .write_to_file("src/contracts/staked_ohm.rs")
  .unwrap();
```

I expected to see this happen: the generated code compiles

Instead, this happened: 

```
error[E0201]: duplicate definitions with name `index`:
   --> src/contracts/staked_ohm.rs:153:5
    |
51  | /     pub fn index(&self) -> ethers::contract::builders::ContractCall<M, ethers::core::types::U256> {
52  | |       self
53  | |         .0
54  | |         .method_hash([45, 247, 92, 177], ())
55  | |         .expect("method not found (this should never happen)")
56  | |     }
    | |_____- previous definition of `index` here
...
153 | /     pub fn index(&self) -> ethers::contract::builders::ContractCall<M, ethers::core::types::U256> {
154 | |       self
155 | |         .0
156 | |         .method_hash([41, 134, 192, 229], ())
157 | |         .expect("method not found (this should never happen)")
158 | |     }
    | |_____^ duplicate definition
```

mattsse:
## Motivation
Close #571

## Solution
look for overloaded functions with different casing when deriving aliases

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst:
Instead of casing and suffixing _0/_1 etc., should we instead just maintain the same casing as the source function?

gakonst (approved):
🔥🔥

PR #652, opened 2021-12-06
feat(solc): add support for compiling solc in parallel

mattsse:
## Motivation
If the svm feature is used, contracts are grouped by their required solidity version and compiled separately. Instead of processing this sequentially, we can spawn multiple processes concurrently using the `async_compile` functions which is faster.

## Solution

* Add `Solc::compile_many -> CompileMany` which effectively works like a `FuturesUnordered` capped at a certain number, (`num_cpus` in our case).
* add a `job` parameter in `Project` which sets this threshold, (default, num_cpu)
* when calling `svm_compile` we check if we can use `compile_many` and use that instead

I was able to unify some stuff and refactor some things, but there is probably still some code duplication in those to branches (jobs == 1 || jobs > 1) when it comes to artifact and cache handling.

needs some further testing with forge.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
lgtm - pending further performance improvements on a per-file level

PR #655, opened 2021-12-06
Fix nonce too low for different providers

kekonen:
This is a complimentary PR to the previous one https://github.com/gakonst/ethers-rs/pull/643
This one solves the same issue but for different providers.
The implemented test is based on: https://github.com/smartcontractkit/chainlink/blob/develop/core/services/eth/errors_test.go#L24-L50

### DIFF
--- a/ethers-providers/src/pending_escalator.rs
+++ b/ethers-providers/src/pending_escalator.rs
@@ -126,6 +126,15 @@ macro_rules! completed {
     };
 }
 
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

gakonst (approved):
Ack. I think should probably move to tighter provider errors instead of manually parsing the string everywhere, but this lgtm

PR #656, opened 2021-12-06
feat(solc): add support for library linking

mattsse:
## Motivation
Add support for lib linking

## Solution
Turn `Bytecode`'s `object` into an enum and add support to link and resolve

following https://docs.soliditylang.org/en/develop/using-the-compiler.html#library-linking
and https://github.com/ethereum/solc-js

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog


gakonst (approved):
LGTM

PR #662, opened 2021-12-09
add buffer for file reading

laizy:
https://github.com/gakonst/foundry/issues/198

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse (approved):
nice catch!

I guess this is due to how serde reads from reader, the serde docs also use Bufreader.
lgtm

gakonst (approved):
Very nice catch.

PR #663, opened 2021-12-09
feat: trezor support

joshieDo:
## Motivation

Ref: #1
Adding trezor signing support, with the end goal of adding it to [foundry/cast](https://github.com/gakonst/foundry/tree/master/cast).

## Solution

Forked a year old rust client (https://github.com/romanz/rust-trezor-api -> https://github.com/joshieDo/rust-trezor-api) and added the ethereum interface it lacked, along with updating the libusb (+rusb) to be compatible with ethers-rs.  I'm not much experienced in Rust, so an additional pair of eyes in my fork is most welcome.

On the `ethers-signers` implementation, I tried to follow the existing Ledger structure as much as possible.

- [x] get_address
- [x] sign_message
- [x] sign_tx
- [x] sign_eip1559
- [ ] sign_eip712

## ~~WIP~~

~~Fixing: Passphrase is being requested on every request~~

## Missing EIP712

I need to understand  EIP712 better from both sides, At first glance, it seems tricky to pass the intended data from `ethers-rs` to the `trezor client`. I was hoping that could be done in a future issue/PR. What do you think?

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1697,31 +1697,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ebdb29d2ea9ed0083cd8cece49bbd968021bd99b0849edb4a9a7ee0fdf6a4e0"
 
 [[package]]
-name = "hid"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "74ec4bb2e3f01d3838ea2bbbe3919d98f921c7aeed14a5491df89ac3dc4839f9"
-dependencies = [
- "hidapi-sys",
- "libc",
-]
-
-[[package]]
 name = "hidapi"
 version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8d6f5e247bc66f3255d755e96d9d43f6b191f4e182072b811d55584ff58c510f"
-dependencies = [
- "cc",
- "libc",
- "pkg-config",
-]
-
-[[package]]
-name = "hidapi-sys"
-version = "0.1.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dd8a9410aec7ca9f4571ff40c7b1813a28503c2a664a028921fc973073dcd4bf"
 dependencies = [
  "cc",
  "libc",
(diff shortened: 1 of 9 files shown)


ethers-signers/Cargo.toml
```diff
@@ -28,6 +28,7 @@ yubihsm = { version = "0.39.0", features = ["secp256k1", "http", "usb"], optiona
 futures-util = "0.3.18"
 futures-executor = "0.3.18"
 semver = "1.0.4"
+trezor-client = { package="trezor", git = "https://github.com/joshiedo/rust-trezor-api", rev = "f0c9775", optional = true }
```
mattsse, line 31: this would prevent us from publishing a new version of ethers, 
what's the state of that crate, any chance you can publish this beforehand?

mattsse (commented):
This is great.
can't speak to the maturity of the trezor client lib, but before some nits

there are a quite a few unwraps, which may be fine but to be sure I'd rather see a basic helper function like `split(&tx) -> Result<(nonce,gasprice,...)>` that gets all values in a safe way.

Any plans for publishing the trezor on crates?

ethers-signers/Cargo.toml
joshieDo: done: [trezor-client](https://crates.io/crates/trezor-client)

joshieDo:
> This is great. can't speak to the maturity of the trezor client lib, but have some nits:
> 
> there are a quite a few unwraps, which may be fine but to be sure I'd rather see a basic helper function like `split(&tx) -> Result<(nonce,gasprice,...)>` that gets all values in a safe way.

cleaned it up (also on the trezor client lib side)

Cargo.lock
```diff
@@ -215,6 +221,38 @@ dependencies = [
...
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cd5e7935f613ba170459072926f01dc5ddb8aa22382dc4badf44bbb55e2d243d"
+dependencies = [
+ "bitcoin-bech32",
+ "bitcoin_hashes",
+ "byteorder",
+ "hex 0.3.2",
+ "rand 0.3.23",
+ "secp256k1",
+]
```
gakonst, line 236: It seems like we're pulling in a few dependencies which should not be required? Maybe we can feature gate the Bitcion/Ethereum APIs in the trezor library?

gakonst (changes requested):
LGTM - but let's maybe try to reduce the deps introduced by the trezor lib?

Cargo.lock
joshieDo: makes sense, done

gakonst (approved):
LGTM! Thank you.

PR #664, opened 2021-12-09
feat(abigen): use structs for outputs

mattsse:
## Motivation
Closes #661 

## Solution
substitute output tuples that are structs with rust structs, same as inputs

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

PR #665, opened 2021-12-10
fix: naming and default impl

roynalnaruto:
## Motivation

Adding some helpers for EvmVersion. Those will be useful when configuring the `SolcConfig` before compiling a `Project`. This is the `ethers-rs` PR to fix https://github.com/gakonst/foundry/issues/202

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts.rs
+++ b/ethers-solc/src/artifacts.rs
@@ -185,7 +185,7 @@ impl Default for Settings {
             optimizer: Default::default(),
             metadata: None,
             output_selection: Self::default_output_selection(),
-            evm_version: Some(EvmVersion::Istanbul),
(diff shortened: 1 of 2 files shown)


mattsse (commented):
lgtm,
but it seems like the tests using ganache are failing now,
maybe we have to set to instanbul explicitly now?

roynalnaruto:
@mattsse All tests are passing for me. Not sure why those fail in the CI

mattsse:
this looks like a solc issu, emb just a flake.
we run with 0.6.0 currently in CI and the contracts should be supported, they passed in #664 🤔

gakonst:
I'm :+1: to bump Solc in tests

PR #668, opened 2021-12-10
Serialize `eth_feeHistory` block count as `QUANTITY`

nlordell:
## Motivation

Retrieving fee history with, say, a `u64` would serialize the the `block_count` as a number twice, instead of first trying as a `QUANTITY` which is my understanding of what it is meant to do (i.e. hex-encoded string - e.g. `0x42`). This PR aims to fix that.

## Solution

Instead of serializing the `block_count` that is passed in, it first converts to a `U256` (which serializes as a `QUANTITY`).

## PR Checklist

- [ ] Added Tests (No - but should be covered by existing tests?)
- [ ] Added Documentation (No - behaviour shouldn't have been changed)
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
   [597](https://github.com/gakonst/ethers-rs/pull/597)
 - Implement hex display format for `ethers::core::Bytes`
   [#624](https://github.com/gakonst/ethers-rs/pull/624).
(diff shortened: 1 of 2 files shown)


gakonst (commented):
Good catch!

nlordell:
There was an issue!

PR #674, opened 2021-12-10
chore: allow unused http_get

mattsse:
## Motivation
allow unused for http_get to silence CI

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/util.rs
+++ b/ethers-contract/ethers-contract-abigen/src/util.rs
@@ -84,7 +84,7 @@ where
 }
 
 /// Perform an HTTP GET request and return the contents of the response.
-#[allow(unused)]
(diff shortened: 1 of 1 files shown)


ethers-contract/ethers-contract-abigen/src/util.rs
```diff
@@ -84,6 +84,7 @@ where
 }
 
 /// Perform an HTTP GET request and return the contents of the response.
+#[allow(unused)]
```
gakonst, line 87: i think this is probably the right fix, instead of silencing the warning?

```suggestion
#[cfg(not(target_arch = "wasm32"))]
```

gakonst (changes requested):

PR #675, opened 2021-12-11
format_units return was truncating the decimal places 

x3ccd4828:
## Motivation
The recent change to format_units was truncating the leading zeros of the decimal places causing the following output as an example: 1001234567891234567 eth became 1.1234567891234567 when formatted. 

## Solution
Added zero padding to the decimal numbers when printed.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
Good catch!

PR #677, opened 2021-12-11
feat(solc): add hardhat artifact support

### LINKED ISSUE #676: ethers-solc: support previous hardhat builds
* ethers-solc and hardhat use different artifact layout but same cache format, so if a user has built with hardhat and then again with forge, we fail to decode the artifact because the cache tells us nothing has changed

I believe we can simply change our format identifier and check if it's the hardhat one or ethers-solc one

**Additional context**
https://github.com/gakonst/foundry/issues/209

mattsse:
## Motivation
Add support for hardhat style artifacts
Closes #676

## Solution
* Add artifact bindings + handler
* Add additional Artifacthandler `MinimalCombinedArtifactsHardhatFallback` that tries to deserialize hardhat artifacts if the previous attempt failed. Alternatively this behaviour could be moved to a bool flag in the existing `MinimalCombinedArtifacts` handler and enabled by default?

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog


gakonst (approved):
I like it - merging optimistically and we can iterate on it

PR #678, opened 2021-12-11
feat(solc): add workspace utils

mattsse:
## Motivation
Add utils for `ethers-solc` to easily create temp workspaces we can also use in foundry.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/project_util.rs
+++ b/ethers-solc/src/project_util.rs
@@ -14,7 +14,7 @@ use tempdir::TempDir;
 
 pub struct TempProject<T: ArtifactOutput> {
     /// temporary workspace root
-    root: TempDir,
+    _root: TempDir,
     /// actual project workspace with the `root` tempdir as its root
     inner: Project<T>,
 }
@@ -22,7 +22,7 @@ pub struct TempProject<T: ArtifactOutput> {
 impl<T: ArtifactOutput> TempProject<T> {
     /// Makes sure all resources are created
     fn create_new(root: TempDir, inner: Project<T>) -> std::result::Result<Self, SolcIoError> {
-        let project = Self { root, inner };
+        let project = Self { _root: root, inner };
         project.paths().create_all()?;
         Ok(project)
     }
@@ -56,36 +56,9 @@ impl<T: ArtifactOutput> TempProject<T> {
         self.project().paths.root.as_path()
     }
 
-    /// Returns the handle to the tempdir and the project
-    ///
-    /// NOTE: the `TempDir` object deletes its directory on drop, also removing all the project's
-    /// content
-    pub fn split(self) -> (Project<T>, TempDir) {
-        (self.inner, self.root)
-    }
-
-    /// Copies a single file into the given dir
-    pub fn copy_file(&self, source: impl AsRef<Path>, target_dir: impl AsRef<Path>) -> Result<()> {
-        let source = source.as_ref();
-        let target = target_dir.as_ref().join(
-            source
-                .file_name()
-                .ok_or_else(|| SolcError::msg(format!("No file name for {}", source.display())))?,
-        );
-
-        fs_extra::file::copy(source, target, &file_copy_options())?;
-        Ok(())
-    }
-
-    /// Copies all content of the source dir into the target dir
-    pub fn copy_dir(&self, source: impl AsRef<Path>, target_dir: impl AsRef<Path>) -> Result<()> {
-        fs_extra::dir::copy(source, target_dir, &dir_copy_options())?;
-        Ok(())
-    }
-
     /// Copies a single file into the projects source
     pub fn copy_source(&self, source: impl AsRef<Path>) -> Result<()> {
-        self.copy_file(source, &self.paths().sources)
+        copy_file(source, &self.paths().sources)
     }
 
     pub fn copy_sources<I, S>(&self, sources: I) -> Result<()>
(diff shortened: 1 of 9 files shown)


ethers-solc/src/project_util.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// Returns the handle to the tempdir and the project
+    ///
+    /// NOTE: the `TempDir` object deletes its directory on drop, also removing all the project's
+    /// content
+    pub fn split(self) -> (Project<T>, TempDir) {
+        (self.inner, self.root)
+    }
+
+    /// Copies a single file into the given dir
+    pub fn copy_file(&self, source: impl AsRef<Path>, target_dir: impl AsRef<Path>) -> Result<()> {
```
gakonst, line 68: Think this shouldn't be needed?

```suggestion
    pub fn copy_file(source: impl AsRef<Path>, target_dir: impl AsRef<Path>) -> Result<()> {
```

ethers-solc/src/project_util.rs
```diff
@@ -0,0 +1,191 @@
...
+        self.project().paths.root.as_path()
+    }
+
+    /// Returns the handle to the tempdir and the project
+    ///
+    /// NOTE: the `TempDir` object deletes its directory on drop, also removing all the project's
+    /// content
+    pub fn split(self) -> (Project<T>, TempDir) {
+        (self.inner, self.root)
+    }
```
gakonst, line 65: This seems unused? Is it needed?

ethers-solc/src/project_util.rs
```diff
@@ -0,0 +1,191 @@
...
+                .file_name()
+                .ok_or_else(|| SolcError::msg(format!("No file name for {}", source.display())))?,
+        );
+
+        fs_extra::file::copy(source, target, &file_copy_options())?;
+        Ok(())
+    }
+
+    /// Copies all content of the source dir into the target dir
+    pub fn copy_dir(&self, source: impl AsRef<Path>, target_dir: impl AsRef<Path>) -> Result<()> {
```
gakonst, line 81:
```suggestion
    pub fn copy_dir(source: impl AsRef<Path>, target_dir: impl AsRef<Path>) -> Result<()> {
```

ethers-solc/src/project_util.rs
```diff
@@ -0,0 +1,191 @@
...
+        let cache = tmp_dir.path().join("cache");
+        let cache = cache.join(SOLIDITY_FILES_CACHE_FILENAME);
+
+        let paths = ProjectPathsConfig::builder()
+            .cache(cache)
+            .sources(root.join("contracts"))
+            .artifacts(root.join("artifacts"))
+            .lib(root.join("node_modules"))
+            .root(root)
+            .build()?;
```
gakonst, line 139: Can't you use [`ProjectPathsConfig::hardhat`](https://github.com/gakonst/ethers-rs/blob/5dec757493f15be016074e9355f7154d8e84aaae/ethers-solc/src/config.rs#L43) instead?

ethers-solc/src/project_util.rs
```diff
@@ -0,0 +1,191 @@
...
+
+impl TempProject<MinimalCombinedArtifacts> {
+    /// Creates an empty new dapptools style workspace in a new temporary dir
+    pub fn dapptools() -> Result<Self> {
+        let tmp_dir = TempDir::new("tmp_dapp").map_err(|err| SolcError::io(err, "temp_dapp"))?;
+        let root = tmp_dir.path().to_path_buf();
+        let cache = tmp_dir.path().join("cache");
+        let cache = cache.join(SOLIDITY_FILES_CACHE_FILENAME);
+
+        let paths = ProjectPathsConfig::builder()
```
gakonst, line 154: Same comment as above but with the dapptools function

ethers-solc/tests/project.rs
```diff
@@ -47,22 +39,10 @@ fn can_compile_hardhat_sample() {
...
-    let paths = ProjectPathsConfig::builder()
-        .cache(cache)
-        .sources(root.join("src"))
-        .artifacts(artifacts)
-        .lib(root.join("lib"))
-        .root(root)
-        .build()
-        .unwrap();
+    let paths = ProjectPathsConfig::builder().sources(root.join("src")).lib(root.join("lib"));
+    let project = TempProject::<MinimalCombinedArtifacts>::new(paths).unwrap();
```
gakonst, line 44: Maybe this could be done with `TempProject::<_>::dapptools()`?

gakonst (changes requested):
LGTM - small nits / questions

ethers-solc/src/project_util.rs
mattsse: yeah, used that before adding more utility functions, removed now

ethers-solc/src/project_util.rs
mattsse: moved to standalone function, same as copy_file

ethers-solc/src/project_util.rs
mattsse: nice, this simplifies it quite a bit

ethers-solc/tests/project.rs
mattsse: hmm we're essentially setting the project root to `./test-data/..` instead of creating a totally new, empty temp project that `TempProject::<_>::dapptools()` would do.
in these cases we'd need to use `TempProject::new` and pass in the preconfigured ProjectPaths,

gakonst (approved):
great!

PR #679, opened 2021-12-11
fix(solc): exclude sha2 asm feature for msvc

mattsse:
## Motivation

MSVC is not supported for sha2 asm https://github.com/RustCrypto/asm-hashes/issues/17

## Solution

We disable asm if the target is msvc 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (approved):

rhacker:
I have solved this by change to gnu-windows just in case anyone is looking for a quickfix.

PR #680, opened 2021-12-12
refactor(solc): add new io error with path info

mattsse:
## Motivation
If there was an error due to a `File not found` io::Error then there is no info about the path, because `io::File` and alike consists of file descriptors and lacks file name information.

The general solution to that is to add additional context with anyhow::Context for example.

## Solution
Add a `SolcIoError` that also holds the path info when doing general io like reading from files

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
LGTM - hopefully the debugging experience is now a little better

PR #681, opened 2021-12-12
feat(providers) Extract DevMiddleware trait

wolflo:
## Motivation

Extracts a `DevMiddleware` trait, implementing it for `DevRpcMiddleware`. I needed this trait locally, but I think the trait will probably change once we add an [evm-adapters](https://github.com/gakonst/foundry/blob/3635682343318d7e7584b74157793562c9e62c81/evm-adapters/src/sputnik/evm.rs#L78-L81) middleware.

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Updated the changelog -- TODO

### DIFF
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -34,7 +34,7 @@ pub use provider::{FilterKind, Provider, ProviderError};
 
 // feature-enabled support for dev-rpc methods
 #[cfg(feature = "dev-rpc")]
-pub use provider::dev_rpc::DevRpcMiddleware;
+pub use provider::dev_rpc::{DevMiddleware, DevRpcMiddleware};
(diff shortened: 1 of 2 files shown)


gakonst:
Could you provide some more context on why this would be needed? Shouldn't the existing `DevRpcMiddleware` do the trick?

wolflo:
That's probably true. The reason I needed it locally is a bit convoluted, but not a reason to merge it.

I think eventually we'll want to pull it into a trait because evm-adapters will need to snapshot and reset in a different way than a test rpc connection will. But, might be better to wait until this is more clear.

gakonst:
Yeah I'd rather we defer this until it's really needed, if you're OK with that.

PR #682, opened 2021-12-12
ci: test on windows targets

mattsse:
## Motivation
Build on windows targets in CI.
But only builds, for tests, need to figure out how to install ganache and geth on windows

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,7 +229,6 @@ jobs:
     env:
       CFG_RELEASE_CHANNEL: ${{ matrix.cfg_release_channel }}
     strategy:
-      max-parallel: 2
       fail-fast: false
       matrix:
         target: [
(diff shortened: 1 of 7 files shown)


gakonst (approved):
good call - would've caught the Windows incompatibility on svm-rs earlier

.github/workflows/ci.yml
```diff
@@ -222,3 +222,53 @@ jobs:
         run: |
             export PATH=$HOME/bin:$PATH
             ./scripts/examples.sh
+
+  windows-build:
```
gakonst, line 226: Might be worth putting this in a different workflow file?

mattsse: windows.yml?

mattsse: but I'm a little hesitant to move this out, perhaps we can leverage all the geth, ganach stuff and unify everything in a matrix build with all vars etc, like I've attempted here https://github.com/gakonst/foundry/pull/218

gakonst:
Blocked on https://github.com/roynalnaruto/svm-rs/pull/9 getting merged / new release

gakonst (approved):

PR #686, opened 2021-12-13
fix: handling empty tx parameters on trezor-client

joshieDo:
## Motivation
I wrongly assumed that a transaction would come with the necessary params (even if 0, or empty), and would only be None, if they were not even necessary.
When adding trezor to foundry, i realized that's not the case.

## Solution

This PR makes sure the none values are converted into `vec![]` or `0` when necessary.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-signers/src/trezor/app.rs
+++ b/ethers-signers/src/trezor/app.rs
@@ -263,6 +263,20 @@ mod tests {
 
     #[tokio::test]
     #[ignore]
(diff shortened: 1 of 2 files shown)


gakonst (approved):
approving - i don't have a trezor rn to test this myself but trust that the test passes :D

PR #695, opened 2021-12-15
fix: contract creation on trezor

joshieDo:
## Motivation
To create a contract, it's required to pass the `to` address  as an empty string.

## Solution
* remove defaulting to the 0x0 address, if no address is passed
* handle `to` as an empty string

It should show on the device:
```
0 Wei ETH
new contract?"
```

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-signers/src/trezor/types.rs
+++ b/ethers-signers/src/trezor/types.rs
@@ -4,7 +4,7 @@
 use std::fmt;
 use thiserror::Error;
 
-use ethers_core::types::{transaction::eip2718::TypedTransaction, NameOrAddress, H160, U256};
+use ethers_core::types::{transaction::eip2718::TypedTransaction, NameOrAddress, U256};
 use trezor_client::client::AccessListItem as Trezor_AccessListItem;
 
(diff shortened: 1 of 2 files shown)


ethers-signers/src/trezor/types.rs
```diff
@@ -72,9 +72,20 @@ impl TrezorTransaction {
...
-            NameOrAddress::Address(value) => format!("0x{}", hex::encode(value)),
+        let to: String = match tx.to() {
+            Some(v) => match v {
+                NameOrAddress::Name(value) => {
+                    // Contract Creation
+                    if value.is_empty() {
+                        "".to_string()
+                    } else {
+                        unimplemented!()
+                    }
```
gakonst, line 83: think we should just be erroring here if a user passes an ENS name instead of silently letting it through in the empty case or panicking if it's not empty?

gakonst (changes requested):
1 small change

gakonst (approved):

PR #699, opened 2021-12-16
add deploy_tokens to ContractFactory

joshieDo:
## Motivation
It's more of a proposal, which i believe would ease things up when deploying from a ContractFactory.

Right now, `deploy()` seems to assume that the `constructor_args` passed are of the exact type that the constructor expects. So if one passes `deploy((1, "string".to_string()))`   to a constructor that accepts `constructor(uint256 a, string memory b)`, it will fail because it will tokenize `1` as `Int` instead of `Uint`. 

This way, we maintain the method, but add a new one which receives an already ready `Vec<Token>`. Other libs have more freedom on doing their own parsing, and makes things easier if one requires to parse arguments from string. (eg. https://github.com/gakonst/foundry/blob/master/utils/src/lib.rs#L119-L125)

Another option would be  to bring over the code from the above link, taking `Vec<String` as `constructor_args`, and parsing it internally.

 

## Solution
Adds `pub fn deploy_tokens(params: Vec<Token>)` 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-contract/src/factory.rs
+++ b/ethers-contract/src/factory.rs
@@ -1,7 +1,7 @@
 use crate::{Contract, ContractError};
 
 use ethers_core::{
-    abi::{Abi, Tokenize},
+    abi::{Abi, Token, Tokenize},
     types::{transaction::eip2718::TypedTransaction, BlockNumber, Bytes, TransactionRequest},
 };
 use ethers_providers::Middleware;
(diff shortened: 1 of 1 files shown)


gakonst (approved):
LGTM

PR #701, opened 2021-12-17
fix: parse_units was generating an error on some values 

x3ccd4828:
## Motivation

parse_units was generating an error on some values because of extra decimal places when multiplying by 10^decimals 

## Solution

added round before converting to U256 to prevent error

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
good catch - ty

PR #704, opened 2021-12-18
feat(abigen): add EthAbiCodec proc macro

mattsse:
## Motivation
Add AbiEncode + AbiDecode implementation for custom structs.
Currently, those get derived for EthCall only.

## Solution
Add `EthAbiCodec` as a complementary proc macro to `EthAbiType` that derives encode/decode.
The reason why this needs to be separate and not included in `EthAbiType` is because in EthCall we have an additional leading 4byte selector, so `EthCall` needs to implement the codec traits but requires all the tokenized support that `EthAbiType` provides.
So we simply add an additional `EthAbiCodec` derive when we generate AbiEncoderV2 structs within abigen!

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-contract/tests/common/derive.rs
+++ b/ethers-contract/tests/common/derive.rs
@@ -1,6 +1,8 @@
-use ethers_contract::{abigen, EthAbiType, EthCall, EthDisplay, EthEvent, EthLogDecode};
+use ethers_contract::{
+    abigen, EthAbiCodec, EthAbiType, EthCall, EthDisplay, EthEvent, EthLogDecode,
+};
 use ethers_core::{
-    abi::{RawLog, Tokenizable},
+    abi::{AbiDecode, AbiEncode, RawLog, Tokenizable},
     types::{Address, H160, H256, I256, U128, U256},
 };
 
@@ -477,4 +479,19 @@ fn can_derive_for_enum() {
 
     let token = ActionChoices::GoLeft.into_token();
     assert_eq!(ActionChoices::GoLeft, ActionChoices::from_token(token).unwrap());
+}
+
+#[test]
+fn can_derive_abi_codec() {
+    #[derive(Debug, Clone, PartialEq, EthAbiType, EthAbiCodec)]
+    pub struct SomeType {
+        inner: Address,
+        msg: String,
+    }
+
+    let val = SomeType { inner: Default::default(), msg: "hello".to_string() };
+
+    let encoded = val.clone().encode();
+    let other = SomeType::decode(&encoded).unwrap();
+    assert_eq!(val, other);
 }
--- a/ethers-contract/tests/common/derive.rs
+++ b/ethers-contract/tests/common/derive.rs
@@ -495,3 +495,68 @@ fn can_derive_abi_codec() {
     let other = SomeType::decode(&encoded).unwrap();
     assert_eq!(val, other);
 }
+
+#[test]
+fn can_derive_abi_codec_single_field() {
+    #[derive(Debug, Clone, PartialEq, EthAbiType, EthAbiCodec)]
+    pub struct SomeType {
+        inner: Vec<U256>,
+    }
+
+    let val = SomeType { inner: Default::default() };
+
+    let encoded = val.clone().encode();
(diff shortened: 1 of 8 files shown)


mattsse:
Fixed clippy lints on #705

sveitser:
Awesome. It works well on the struct I was testing with.

I noticed something that seems a bit strange to me. In order to `abi.decode` in the contract I am wrapping the struct in a one-element tuple then tokenize and encode.

If I use the new decode method on that I do get and extra `32` in front  (compared to doing `into_tokens` followed by `encode`).

Tried to illustrate this below
```rust
    abigen!(
        Example,
        r#"[
        struct Foo { uint256[] a; }
        function readFoo(bytes b)
        ]"#,
        event_derives(serde::Deserialize, serde::Serialize);
    );

    fn disp(b: &[u8]) {
        hex::encode(b)
            .chars()
            .collect::<Vec<char>>()
            .chunks(64)
            .for_each(|x| println!("{}", x.iter().collect::<String>()));
        println!("");
    }

    let encoded = foo.clone().encode();
    disp(&encoded);
    // 0000000000000000000000000000000000000000000000000000000000000020
    // 0000000000000000000000000000000000000000000000000000000000000001
    // 00000000000000000000000000000000000000000000000000000000000000ff

    // same as encode above
    let from_tokens = abi::encode(&Tokenize::into_tokens(foo.clone()));
    disp(&from_tokens);
    // 0000000000000000000000000000000000000000000000000000000000000020
    // 0000000000000000000000000000000000000000000000000000000000000001
    // 00000000000000000000000000000000000000000000000000000000000000ff

    let encoded_wrapped = (foo.clone(),).encode();
    disp(&encoded_wrapped);
    // 0000000000000000000000000000000000000000000000000000000000000020  X
    // 0000000000000000000000000000000000000000000000000000000000000020  X
    // 0000000000000000000000000000000000000000000000000000000000000020
    // 0000000000000000000000000000000000000000000000000000000000000001
    // 00000000000000000000000000000000000000000000000000000000000000ff

    // this works with abi.decode(b, (Foo)) in a contract
    let from_tokens_wrapped = abi::encode(&Tokenize::into_tokens((foo.clone(),)));
    disp(&from_tokens_wrapped);
    // 0000000000000000000000000000000000000000000000000000000000000020  X
    // 0000000000000000000000000000000000000000000000000000000000000020
    // 0000000000000000000000000000000000000000000000000000000000000001
    // 00000000000000000000000000000000000000000000000000000000000000ff
```

mattsse:
nice catch!
I think I fixed tuple codec now, could you try again please @sveitser?

sveitser:
Yes. I'm getting matching outputs now for the tuple :partying_face: 

```
    let encoded_wrapped = (foo.clone(),).encode();
    disp(&encoded_wrapped);
    // 0000000000000000000000000000000000000000000000000000000000000020  X
    // 0000000000000000000000000000000000000000000000000000000000000020
    // 0000000000000000000000000000000000000000000000000000000000000001
    // 00000000000000000000000000000000000000000000000000000000000000ff
```

ethers-contract/tests/common/derive.rs
```diff
@@ -478,3 +480,83 @@ fn can_derive_for_enum() {
...
+
+    let encoded_tuple = (Vec::<U256>::default(), Address::default()).encode();
+
+    assert_eq!(encoded_tuple, encoded);
+    let decoded_tuple = SomeType::decode(&encoded_tuple).unwrap();
+    assert_eq!(decoded_tuple, decoded);
+
+    let tuple = (val,);
+    let encoded = tuple.clone().encode();
+    let decoded = <(SomeType,)>::decode(&encoded).unwrap();
```
gakonst, line 553: this tuple syntax seems like a footgun? can we get rid of the `x,)` requirement?

mattsse: it's a bit weird but this required rust's syntax for single-element tuple, otherwise it would just be a parenthesized expression

image
https://doc.rust-lang.org/1.30.0/book/second-edition/appendix-02-operators.html

gakonst (approved):

PR #706, opened 2021-12-18
feat(solc): revamped remapping auto detect

mattsse:
## Motivation
Better, faster remappings autodetection with more hardhat layouts in mind

## Solution
optimize lib folder scanning, unify nested hardhat remappings.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
not sure i understand 100% the algo, but if the unit tests + new one pass I'm :+1:

PR #709, opened 2021-12-19
chore(solc): use svm global_version if it exists and no SOLC_PATH is set

mattsse:
## Motivation
consider `~/svm/.global_version` in `Solc::default`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/compile.rs
+++ b/ethers-solc/src/compile.rs
@@ -88,8 +88,11 @@ impl Default for Solc {
         }
         #[cfg(not(target_arch = "wasm32"))]
         {
-            if let Some(solc) = Solc::svm_global_version().filter(|p| p.exists()) {
-                return Solc::new(solc)
+            if let Some(solc) = Solc::svm_global_version()
+                .and_then(|vers| Solc::find_svm_installed_version(&vers.to_string()).ok())
+                .flatten()
+            {
+                return solc
             }
         }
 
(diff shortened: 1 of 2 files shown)


ethers-solc/src/compile.rs
```diff
@@ -78,7 +83,17 @@ pub struct Solc {
...
 impl Default for Solc {
     fn default() -> Self {
-        std::env::var("SOLC_PATH").map(Solc::new).unwrap_or_else(|_| Solc::new(SOLC))
+        if let Ok(solc) = std::env::var("SOLC_PATH") {
+            return Solc::new(solc)
+        }
+        #[cfg(not(target_arch = "wasm32"))]
+        {
+            if let Some(solc) = Solc::svm_global_version().filter(|p| p.exists()) {
+                return Solc::new(solc)
```
gakonst, line 92: .global-version returns only the version, not the path to the binary with that version. So we need an extra resolution from version to path

gakonst (changes requested):

ethers-solc/src/compile.rs
mattsse: oh you're right, somehow I thought this was a copy

gakonst (approved):
LGTM, we could have done `if let Ok(...)` instead of `.ok()` and doing `if let Some(...)`?

PR #711, opened 2021-12-19
feat(ethers): add default features

gakonst:


gakonst:
I'd rather keep everything off by default, and can't bother thinking about why wasm fails again lol

PR #715, opened 2021-12-19
chore: remove libudev-dev dependency  [WIP]

joshieDo:
## Motivation
Removes libudev-dev dependency.

## Solution
Replaces `hidapi` for `hidapi-rusb` on both the [ledger](https://github.com/summa-tx/bitcoins-rs/pull/90) (PENDING) and [trezor](https://github.com/joshieDo/rust-trezor-api/pull/2). It makes the `hidapi` backend to be `libusb`, which gets built by `rusb`.

NOTE: Had to bump `yubihsm` version because of incompatibilities on `rusb` versions. It added quite a few dependencies on the lock file...

**The ledger dependency on the git repository is temporary.**

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,8 +90,6 @@ jobs:
             export PATH=$HOME/bin:$PATH
             geth version
 
-      - name: Install libusb (for Ledger)
-        run: sudo apt update && sudo apt install pkg-config libudev-dev
       - name: Install stable toolchain
         uses: actions-rs/toolchain@v1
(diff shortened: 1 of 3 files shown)


gakonst:
Blocked on #617, so that we bump yubihsm to its latest version which uses the right rusb version

gakonst:
included in https://github.com/gakonst/ethers-rs/pull/617

PR #719, opened 2021-12-20
fix(solc): handle more remapping edge cases

mattsse:
## Motivation
Better remapping auto detection for both hardhat and dapp

## Solution
There are quite a few edge cases with certain assumptions based on the dir structure.
Finding remappings works as follows:

1. scan all sub dirs (every dir in a lib like (lib/*, node_modules/*)) for candidates
2. simplify all candidates by looking at their most common ancestor and their surrounding dirs, this has various edge cases, especially in hardhat layouts with `contracts`. This tries its best to anticipate to what dir the remapping should actually point

`dependency/contracts/s.sol` -> `dependency/contracts`
`dependency/contracts/nested/s.sol` -> `dependency/contracts`
`dependency/nested/contracts/s.sol` -> `dependency`

there is probably an easier way to determine this, but for most common cases this should work even though this looks a bit horrible...

Ref https://github.com/gakonst/foundry/pull/273

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -838,6 +838,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "ctor"
+version = "0.1.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ccc0a48a9b826acdf4028595adc9db92caea352f7af011a3034acd172a52a0aa"
+dependencies = [
+ "quote",
+ "syn",
+]
+
+[[package]]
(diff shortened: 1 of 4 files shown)


gakonst (approved):
Let's go with it and I think we refactor/cleanup as we encounter more edge cases. Surely there should be a "clean" way of determining the proper remappings structure.

PR #723, opened 2021-12-21
mark TransactionFut as Send

georgewhewell:
## Motivation

`transactions_unordered()` is not `Send`

## Solution

mark `TransactionFut` as `Send`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-providers/src/stream.rs
+++ b/ethers-providers/src/stream.rs
@@ -162,7 +162,10 @@ impl From<GetTransactionError> for ProviderError {
     }
 }
 
+#[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
lg, but this won't work in wasm.
here's an existing workaround for that

https://github.com/gakonst/ethers-rs/blob/dcf20022c66921779059bbfc58f5d378d20b84be/ethers-middleware/src/gas_escalator/mod.rs#L19-L22

gakonst (approved):
great! if you're building on ethers and not already in the chatroom / want to talk more about what you're doing: https://t.me/ethers_rs

PR #724, opened 2021-12-22
feat(abigen): add MultiAbigen to generate multiple contract bindings

mattsse:
## Motivation
rust-analyzer and intellij still struggle to expand the `abigen!` macro and fail to provide useful intellisense, this makes it harder to get started with `abigen!`and usually requires to expand the macro by hand to see what functions are available.

## Solution
The new `MultiAbigen` type can be used to create bindings for several contract at once and generate a new module like this: https://github.com/Gearbox-protocol/liquidation-bot/blob/master/src/bindings/mod.rs

This can be used to generate a module like this once, keep that in repository so that intellisense is available and ensure the CI will fail if the generated code is out of date

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@
 
 ## ethers-contract-abigen
 
+- Add `MultiAbigen` to generate a series of contract bindings that can be kept in the repo
+  [#724](https://github.com/gakonst/ethers-rs/pull/724).
 - Add provided `event_derives` to call and event enums as well
   [#721](https://github.com/gakonst/ethers-rs/pull/721).
 - Implement snowtrace and polygonscan on par with the etherscan integration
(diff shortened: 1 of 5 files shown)


gakonst (approved):
Nice - wonder if we should create a binary which exposes the functionality that one can easily install to re-generate their bindings? Feels like this is the high level utility we needed to make such a CLI useful.

mattsse:
> Nice - wonder if we should create a binary which exposes the functionality that one can easily install to re-generate their bindings? Feels like this is the high level utility we needed to make such a CLI useful.

I don't know if a cli is appropriate for this feature only, but adding such a cli should be pretty easy. 
The intended setup would be to run this as in a test to ensure everything is up to date.

What could be useful as well is a workspace template for https://github.com/cargo-generate/cargo-generate

PR #725, opened 2021-12-22
feat: add EventStream::select to combine multiple event streams

mattsse:
## Motivation
Combine multiple `EventStream` with different items

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-contract/src/stream.rs
+++ b/ethers-contract/src/stream.rs
@@ -1,6 +1,9 @@
 use crate::LogMeta;
 use ethers_core::types::{Log, U256};
-use futures_util::stream::{Stream, StreamExt};
+use futures_util::{
+    future::Either,
+    stream::{Stream, StreamExt},
+};
 use pin_project::pin_project;
 use std::{
     pin::Pin,
@@ -23,6 +26,7 @@ pub struct EventStream<'a, T, R, E> {
 }
 
 impl<'a, T, R, E> EventStream<'a, T, R, E> {
+    /// Turns this stream of events into a stream that also yields the event's metadata
     pub fn with_meta(self) -> EventStreamMeta<'a, T, R, E> {
         EventStreamMeta(self)
     }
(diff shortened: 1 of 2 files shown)


ethers-contract/src/stream.rs
```diff
@@ -49,9 +53,140 @@ where
...
+where
+    T: Stream<Item = Either<Result<L, LE>, Result<R, RE>>> + 'a,
+    L: 'a,
+    LE: 'a,
+    R: 'a,
+    RE: 'a,
+{
+    /// Turns a stream of Results to a stream of `Result::ok` for both arms
+    pub fn ok(self) -> Pin<Box<dyn Stream<Item = Either<L, R>> + 'a>> {
+        Box::pin(self.filter_map(|e| async move {
```
gakonst, line 145: Do we want the `filter_map` here? Won't this silently eat all errors if any are observed?

ethers-contract/src/stream.rs
```diff
@@ -49,9 +53,140 @@ where
...
+    /// while let Some(either) = events.next().await {
+    ///     match either {
+    ///         Either::Left(approval) => { let Approval{token_owner,spender,tokens} = approval; }
+    ///         Either::Right(transfer) => { let Transfer{from,to,tokens} = transfer; }
+    ///     }
+    /// }
+    ///
+    /// # }
+    /// ```
+    pub fn select<St>(self, st: St) -> SelectEvent<SelectEither<'a, Result<R, E>, St::Item>>
```
gakonst, line 119: Wonder if there's a world where we do this with a macro instead of a function (like the futures `select!`), so that we are not limited to only selecting 2 events?

ethers-contract/src/stream.rs
mattsse: yes this skips all erroneous values, but is probably useful if you're only interested in the `ok` values.

ethers-contract/src/stream.rs
mattsse: driving multiple streams to completion with `select!` is a bit verbose and the core problem is that we need to return a type that represents all the different events.
you can do this with `EventStream::select` 

```rust
let ev3 = contract.event::<OtherEvent>();

let stream = ev3.stream().await().unwrap().select(
   ev1.stream().await.unwrap().select(ev2.stream().await.unwrap())
);
```
which will return an `Either<OtherEvent, Either<Approval, Transfer>>`

But there'd be also the possibility to create your own event type and use the `Contract::event_with_filter` function

```rust
 #[derive(Debug, Clone, PartialEq, Eq)]
    pub enum Events {
        Approval(Approval),
        Transfer(Transfer),
        Other(Other),
    }
  impl EthLogDecode for Events {...}

```

gakonst: Ah, right! OK if that pattern works then should be fine.

gakonst (approved):

PR #733, opened 2021-12-24
feat(abigen): search json recursively

mattsse:
## Motivation
scan folder recursively to make it compatible with `ethers-solc` artifacts output

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

PR #736, opened 2021-12-25
MultiAbigen recursively read json files under given directory

thasarito:
## Motivation

Easier integration when working `abi` generated from `forge build` which `json` files are stored under the folders. 

## Solution

Use `walkdir` to recursively read files (following pattern from this module https://github.com/gakonst/ethers-rs/blob/master/ethers-solc/src/utils.rs#L54).

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1111,6 +1111,7 @@ dependencies = [
  "syn",
  "tempfile",
  "url",
+ "walkdir",
 ]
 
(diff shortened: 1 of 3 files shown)


thasarito:
Just seen that this is duplicate of #733

PR #747, opened 2021-12-28
feat/trezor: cache session on filesystem

joshieDo:
## Motivation

Sessions are used to "keep" the  passphrase on the device in-between different calls. And while it was being reused inside the same `TrezorEthereum` object, once we instantiated a new one, we would lose the session.

For example, different `forge` or `cast` calls would always request the user to insert their passphrase again and again. With this change, it won't.

Relevant information from  [sessions @ docs.trezor.io](https://docs.trezor.io/trezor-firmware/common/communication/sessions.html) :
* `Sessions only exist in RAM and are lost when Trezor is disconnected.`
* `New session is started by calling Initialize with no arguments. The response is a Features message, which contains a 32-byte session_id. All subsequent commands happen within the given session.`
* `To resume a previous session (after creating a new one), call Initialize with a stored session_id as an argument. Attempt to resume an unknown session ID will transparently allocate a new session ID.`

## Solution
Read/write `session_id` to `./cache/trezor.session`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-signers/src/trezor/app.rs
+++ b/ethers-signers/src/trezor/app.rs
@@ -34,6 +34,9 @@ pub struct TrezorEthereum {
 
 const FIRMWARE_MIN_VERSION: &str = ">=2.4.2";
 
+// https://docs.trezor.io/trezor-firmware/common/communication/sessions.html
+const SESSION_ID_LENGTH: usize = 32;
+
 impl TrezorEthereum {
     pub async fn new(derivation: DerivationType, chain_id: u64) -> Result<Self, TrezorError> {
         let mut blank = Self {
@@ -62,17 +65,15 @@ impl TrezorEthereum {
     }
 
     fn get_cached_session() -> Result<Option<Vec<u8>>, TrezorError> {
-        let mut session = [0; 32];
+        let mut session = [0; SESSION_ID_LENGTH];
 
         let path = env::current_dir()
             .map_err(|e| TrezorError::CacheError(e.to_string()))?
             .join("cache")
             .join("trezor.session");
 
-        if let Ok(file) = fs::File::open(path) {
-            let mut file = std::io::BufReader::new(file);
-            file.read(&mut session).map_err(|e| TrezorError::CacheError(e.to_string()))?;
-
+        if let Ok(mut file) = fs::File::open(path) {
+            file.read_exact(&mut session).map_err(|e| TrezorError::CacheError(e.to_string()))?;
             Ok(Some(session.to_vec()))
         } else {
             Ok(None)
@@ -85,9 +86,9 @@ impl TrezorEthereum {
         fs::create_dir_all(&path).map_err(|e| TrezorError::CacheError(e.to_string()))?;
         path = path.join("trezor.session");
 
-        let file = fs::File::create(path).map_err(|e| TrezorError::CacheError(e.to_string()))?;
-        let mut file = std::io::BufWriter::new(file);
-        file.write(&session_id);
+        let mut file =
+            fs::File::create(path).map_err(|e| TrezorError::CacheError(e.to_string()))?;
+        file.write_all(&session_id).map_err(|e| TrezorError::CacheError(e.to_string()))?;
 
         self.session_id = session_id;
         Ok(())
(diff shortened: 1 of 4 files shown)


ethers-signers/src/trezor/app.rs
```diff
@@ -56,9 +61,41 @@ impl TrezorEthereum {
         Ok(())
     }
 
+    fn get_cached_session() -> Result<Option<Vec<u8>>, TrezorError> {
```
mattsse, line 64: is the session only a 32byte identifier?
need definitely some docs on this, or links to trezor api

ethers-signers/src/trezor/app.rs
```diff
@@ -56,9 +61,41 @@ impl TrezorEthereum {
...
+        let mut session = [0; 32];
+
+        let path = env::current_dir()
+            .map_err(|e| TrezorError::CacheError(e.to_string()))?
+            .join("cache")
+            .join("trezor.session");
+
+        if let Ok(file) = fs::File::open(path) {
```
mattsse, line 72: should we return a `None` here if the file does not exist?
and an error if reading the file failed?

ethers-signers/src/trezor/app.rs
```diff
@@ -56,9 +61,41 @@ impl TrezorEthereum {
...
+            let mut file = std::io::BufReader::new(file);
+            file.read(&mut session).map_err(|e| TrezorError::CacheError(e.to_string()))?;
```
mattsse, line 74: don't think a BufReader is necessary here for 32bytes?
also read does not guarantee to fill the buf entirely.
there is https://doc.rust-lang.org/stable/std/io/trait.Read.html#method.read_exact instead, or alternatively File::read_to_end

ethers-signers/src/trezor/app.rs
```diff
@@ -56,9 +61,41 @@ impl TrezorEthereum {
...
+    }
+
+    fn save_session(&mut self, session_id: Vec<u8>) -> Result<(), TrezorError> {
+        let mut path =
+            env::current_dir().map_err(|e| TrezorError::CacheError(e.to_string()))?.join("cache");
+        fs::create_dir_all(&path).map_err(|e| TrezorError::CacheError(e.to_string()))?;
+        path = path.join("trezor.session");
+
+        let file = fs::File::create(path).map_err(|e| TrezorError::CacheError(e.to_string()))?;
+        let mut file = std::io::BufWriter::new(file);
```
mattsse, line 89: same, if the session id is only 32byte, probably not worth it, `fs::write` should do just fine I believe

mattsse (changes requested):

ethers-signers/src/trezor/app.rs
joshieDo: > is the session only a 32byte identifier?

yes. copied the relevant paragraphs into the PR, and added a reference on the code.

ethers-signers/src/trezor/app.rs
joshieDo: done

although in the beginning, i was thinking of issuing a warning, without halting the execution

ethers-signers/src/trezor/app.rs
```diff
@@ -56,9 +64,39 @@ impl TrezorEthereum {
         Ok(())
     }
 
+    fn get_cached_session() -> Result<Option<Vec<u8>>, TrezorError> {
+        let mut session = [0; SESSION_ID_LENGTH];
+
+        let path = env::current_dir()
+            .map_err(|e| TrezorError::CacheError(e.to_string()))?
+            .join("cache")
+            .join("trezor.session");
```
gakonst, line 73: Maybe we should pass a `cache_dir` to trezor when initializing it? Or have a global `~/.ethers-rs/trezor/cache/trezor.session` file?

joshieDo: added all options
1. accepts `cache_dir` as a parameter
2. if none, then tries `home_dir`
3. if none, then tries `current_dir`
4. if err, exit

gakonst (approved):

PR #750, opened 2021-12-29
feat(solc): add dependency graph implementation

### LINKED ISSUE #739: Improve library/import resolution
Currently, resolving libraries is very primitive and basically only supports lookups in the configured lib folder without applying remappings.

Essentially all imports outside of the src directory rely on remappings so solc can apply them.
This means, they're not part of the cache since we don't discover them properly. Therefor changes in  lib contracts can be skipped if caching is enabled

Likely the case here https://github.com/gakonst/foundry/issues/299
also definitely related #727

Short term:
- [ ] resolve remappings in `resolve_libraries`.
- [ ] check all `imports` contracts for changes when checking a `CacheEntry`

Long term, add a proper dependency graph implementation.

mattsse:
## Motivation
Close #739

## Solution
Add a dependency graph implementation, see resolve.rs for docs.

### Changes
* add a `Graph` implementation that also resolves all imported dependencies that live outside of the project's configured source dir. This applies the configured remappings during lookup
* Also add those dependencies to the cache file, so changes in a contract/library in the `lib` folder are detected and trigger a rebuild.
* parallelize as much as possible (reading and parsing of source) files. parallelizing import resolving is not trivial because it can result in duplicated lookups (Lib1 imports Lib3, Lib2 imports Lib3)

ToDos

- [x] version checks, find minimum viable solc version applicable for all dependencies
- [x] integrate in `Project::compile`

Followup
- [ ] add pretty printing like `cargo tree`

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-solc/src/lib.rs
+++ b/ethers-solc/src/lib.rs
@@ -26,6 +26,7 @@ use crate::{artifacts::Source, cache::SolFilesCache};
 
 pub mod error;
 pub mod utils;
+pub mod resolve;
 
 use crate::{
     artifacts::Sources,
--- a/ethers-solc/src/lib.rs
+++ b/ethers-solc/src/lib.rs
@@ -227,6 +227,8 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
         if !self.allowed_lib_paths.0.is_empty() {
             solc = solc.arg("--allow-paths").arg(self.allowed_lib_paths.to_string());
         }
+
+        let sources = Graph::resolve_sources(&self.paths, sources)?.into_sources();
         self.compile_with_version(&solc, sources)
     }
 
@@ -351,9 +353,6 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
         solc: &Solc,
         sources: Sources,
     ) -> Result<ProjectCompileOutput<Artifacts>> {
-        // let resolved = Graph::resolve_sources(&self.paths, sources)?;
-        // let sources = resolved.into_sources();
-
         let (sources, paths, cached_artifacts) = match self.preprocess_sources(sources)? {
             PreprocessedJob::Unchanged(artifacts) => {
                 return Ok(ProjectCompileOutput::from_unchanged(artifacts))
--- a/ethers-solc/src/lib.rs
+++ b/ethers-solc/src/lib.rs
@@ -204,7 +204,8 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
         Ok(libs)
     }
 
-    /// Attempts to compile the contracts found at the configured location.
+    /// Attempts to compile the contracts found at the configured source location, see
+    /// `ProjectPathsConfig::sources`.
     ///
     /// NOTE: this does not check if the contracts were successfully compiled, see
     /// `CompilerOutput::has_error` instead.
@@ -295,6 +296,7 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
         Ok(compiled)
     }
 
+    /// Compiles all sources with their intended `Solc` version sequentially.
     #[cfg(all(feature = "svm", feature = "async"))]
     fn compile_sources(
         &self,
@@ -314,6 +316,9 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
         Ok(compiled)
     }
 
+    /// Compiles all sources with their intended `Solc` version in parallel.
+    ///
+    /// This runs `Self::solc_jobs` parallel `solc` jobs at most.
     #[cfg(all(feature = "svm", feature = "async"))]
     fn compile_many(
         &self,
--- a/ethers-solc/src/lib.rs
+++ b/ethers-solc/src/lib.rs
@@ -7,6 +7,7 @@ use std::collections::btree_map::Entry;
 
 pub mod cache;
 pub mod hh;
+mod resolver;
 pub use hh::{HardhatArtifact, HardhatArtifacts};
 
 mod compile;
@@ -26,7 +27,6 @@ use crate::{artifacts::Source, cache::SolFilesCache};
 
 pub mod error;
 pub mod utils;
-pub mod resolve;
 
 use crate::{
     artifacts::Sources,
@@ -154,8 +154,7 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
     /// Returns all sources found under the project's configured sources path
     #[tracing::instrument(skip_all, fields(name = "sources"))]
     pub fn sources(&self) -> Result<Sources> {
-        tracing::trace!("reading all sources from \"{}\"", self.paths.sources.display());
-        Ok(Source::read_all_from(&self.paths.sources)?)
+        self.paths.read_sources()
     }
 
     /// This emits the cargo [`rerun-if-changed`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath) instruction.
@@ -935,7 +934,7 @@ mod tests {
     fn test_build_many_libs() {
         use super::*;
 
-        let root = dunce::canonicalize("./test-data/test-contract-libs").unwrap();
+        let root = utils::canonicalize("./test-data/test-contract-libs").unwrap();
 
         let paths = ProjectPathsConfig::builder()
             .root(&root)
@@ -962,7 +961,7 @@ mod tests {
     fn test_build_remappings() {
         use super::*;
 
-        let root = dunce::canonicalize("./test-data/test-contract-remappings").unwrap();
+        let root = utils::canonicalize("./test-data/test-contract-remappings").unwrap();
         let paths = ProjectPathsConfig::builder()
             .root(&root)
             .sources(root.join("src"))
--- a/ethers-solc/src/lib.rs
+++ b/ethers-solc/src/lib.rs
@@ -214,7 +214,7 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
     /// solc versions across files.
     #[tracing::instrument(skip_all, name = "compile")]
     pub fn compile(&self) -> Result<ProjectCompileOutput<Artifacts>> {
-        let sources = self.sources()?;
+        let sources = self.paths.read_input_files()?;
         tracing::trace!("found {} sources to compile: {:?}", sources.len(), sources.keys());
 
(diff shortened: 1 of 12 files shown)


gakonst (commented):
seems like we can merge the parallel reading improvements independently of the dep graph code?

mattsse:
@gakonst no idea why abigen tests are failing...
seems unrelated?

ethers-solc/src/lib.rs
```diff
@@ -394,6 +351,9 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
         solc: &Solc,
         sources: Sources,
     ) -> Result<ProjectCompileOutput<Artifacts>> {
+        // let resolved = Graph::resolve_sources(&self.paths, sources)?;
+        // let sources = resolved.into_sources();
```
gakonst, line 355: should this still be commented out?

ethers-solc/src/resolver.rs
```diff
@@ -0,0 +1,518 @@
...
+    /// The indices of `edges` correspond to the `nodes`. That is, `edges[0]`
+    /// is the set of outgoing edges for `nodes[0]`.
+    edges: Vec<Vec<usize>>,
+    /// index maps for a solidity file to an index, for fast lookup.
+    indices: HashMap<PathBuf, usize>,
+    /// with how many input files we started with, corresponds to `let input_files =
+    /// nodes[..num_input_files]`.
+    num_input_files: usize,
+    /// the root of the project this graph represents
+    #[allow(unused)]
```
gakonst, line 55: remove?

ethers-solc/src/resolver.rs
```diff
@@ -0,0 +1,518 @@
...
+    root: PathBuf,
+}
+
+impl Graph {
+    /// Returns a list of nodes the given node index points to for the given kind.
+    pub fn imported_nodes(&self, from: usize) -> &[usize] {
+        &self.edges[from]
+    }
```
gakonst, line 63: what do we get from exposing this as a method?

ethers-solc/src/resolver.rs
```diff
@@ -0,0 +1,518 @@
...
+                .map(|p| Path::new(p).to_path_buf())
+                .collect()
+        }
+    };
+    let version_req = if let Some(ref v) = version { Solc::version_req(v).ok() } else { None };
+    SolData { version_req, version, imports }
+}
+
+#[cfg(test)]
+mod tests {
```
gakonst, line 466: let's add a test to see that we cover the cache issue with libs

ethers-solc/src/resolver.rs
```diff
@@ -0,0 +1,518 @@
...
+        // locations
+        while let Some((path, node)) = unresolved.pop_front() {
+            let mut resolved_imports = Vec::with_capacity(node.data.imports.len());
+
+            // parent directory of the current file
+            let node_dir = path.parent();
+            if node_dir.is_none() {
+                continue
+            }
+            let node_dir = node_dir.unwrap();
```
gakonst, line 118: maybe more idiomatic 
```suggestion
            let node_dir = match path.parent() { Some(inner) => inner, None => continue };
```

ethers-solc/src/resolver.rs
```diff
@@ -0,0 +1,518 @@
...
+
+            for import in node.data.imports.iter() {
+                let component = import.components().next();
+                if component.is_none() {
+                    continue
+                }
+                let component = component.unwrap();
```
gakonst, line 125: same comment re more idiomatic

ethers-solc/src/resolver.rs
```diff
@@ -0,0 +1,518 @@
...
+                    // file set
+                    match utils::canonicalize(node_dir.join(import)) {
+                        Ok(target) => {
+                            // the file at least exists,
+                            if let Some(idx) = index.get(&target).cloned() {
+                                resolved_imports.push(idx);
+                            } else {
+                                // imported file is not part of the input files
+                                let node = read_node(&target)?;
+                                unresolved.push_back((target.clone(), node));
```
gakonst, line 137: I wonder if there can be any infinite loop caused by this pop_front/push_back, not a blocker but smth we could be aware of

ethers-solc/src/resolver.rs
```diff
@@ -0,0 +1,518 @@
...
+                                let idx = index.len();
+                                index.insert(target.clone(), idx);
+                                resolved_imports.push(idx);
+                            }
+                        }
+                        Err(err) => {
+                            tracing::trace!("failed to resolve relative import \"{:?}\"", err);
```
gakonst, line 144: Should we silently ignore this error or should we throw?

ethers-solc/src/resolver.rs
```diff
@@ -0,0 +1,518 @@
...
+                        if let Some(idx) = index.get(&lib).cloned() {
+                            resolved_imports.push(idx);
+                        } else {
+                            // imported file is not part of the input files
+                            let node = read_node(&lib)?;
+                            unresolved.push_back((lib.clone(), node));
+                            let idx = index.len();
+                            index.insert(lib.clone(), idx);
+                            resolved_imports.push(idx);
+                        }
```
gakonst, line 159: should this logic be deduplicated, given it's the same in both the if/else branch, with the only difference being whether `path.resolve_library_import` is called vs `canonicalize(node_dir.join(import))`

ethers-solc/src/resolver.rs
```diff
@@ -0,0 +1,518 @@
...
+        Self::resolve_sources(paths, paths.read_input_files()?)
+    }
+}
+
+#[cfg(all(feature = "svm", feature = "async"))]
+impl Graph {
+    /// Returns all input files together with their appropriate version.
+    ///
+    /// First we determine the compatible version for each input file (sources and test, see
+    /// `Self::resolve`) Then we also add any other dependencies
```
gakonst, line 186:
```suggestion
    /// `Self::resolve`) Then we also add any other dependencies.
```

(or there was some unfinished sentence?)

ethers-solc/src/resolver.rs
```diff
@@ -0,0 +1,518 @@
...
+    pub fn into_sources_by_version(self, offline: bool) -> Result<VersionedSources> {
+        /// insert the imports of the given node into the sources map
+        /// There can be following graph:
+        /// `A(<=0.8.10) imports C(>0.4.0)` and `B(0.8.11) imports C(>0.4.0)`
+        /// where `C` is a library import, in which case we assign `C` only to the first input file.
+        /// However, it's not required to include them in the solc `CompilerInput` as they would get
+        /// picked up by solc otherwise, but we add them so we can create a corresponding
+        /// cache entry for them as well. This can be optimized however
```
gakonst, line 194: We're basically double compiling some files?

ethers-solc/src/resolver.rs
```diff
@@ -0,0 +1,518 @@
...
+            candidates.retain(|v| req.matches(v.as_ref()));
+        }
+        for dep in self.imported_nodes(idx) {
+            let dep = *dep;
+            // check for circular deps which would result in endless recursion SO here
+            // a circular dependency exists, if there was already a `dependency imports current
+            // node` relationship in the traversed path
+            if traversed.contains(&(dep, idx)) {
+                let mut msg = String::new();
+                self.format_imports_list(dep, &mut msg).unwrap();
```
gakonst, line 274: The main reason for `format_imports_list` is for pretty error printing? Let's maybe add that as a note?

gakonst (approved):
This is really great. A few nits, but nothing blocking

ethers-solc/src/resolver.rs
mattsse: the root is used to simplify the file paths for error messages which are feature gated so this simply prevents the unused warning when svm is not enabled

ethers-solc/src/lib.rs
mattsse: moved this to `fn compile`

ethers-solc/src/resolver.rs
gakonst: nit: can add a `cfg_attr(...)` then or some cfg flag for the field?

no need if big trouble

ethers-solc/src/resolver.rs
mattsse: can be used to iterate over all imports of a file.
this will be useful when I add tree printing.
but can be private for now

ethers-solc/src/resolver.rs
mattsse: I'd stick with the unused, because this will be used when adding a function to print the tree, to simplify all filepaths

ethers-solc/src/resolver.rs
mattsse: this is very much exhaustive because we only add a new entry to the list for files we haven't read yet, so even if there are circular deps in the code, there is no issue here because we check if we've read the file already

ethers-solc/src/resolver.rs
mattsse: this can happen if the import path is wrong, points to a file that does not exist.
so the expectation here would be that this will fail in solc regardless but with a better error message.

ethers-solc/src/resolver.rs
mattsse: good call, moved it to an anonymous function

ethers-solc/src/resolver.rs
mattsse: I think that's possible in scenarios like:
`A(<0.8.10) imports C (>0.4.0)`
`B(=0.8.11) imports C (>0.4.0)`

where A and B are input contracts and C is lib import.
In which case we can't compile A and B together but both require C.
But we only have a single instance of C in our dependency graph.
So I think it's fine to only include it in 1 `Solc -> Sources` set, like `0.8.11 -> (B,C)`, because solc will resolve C for A itself if not provided in the standard json input.

the reason why I decided against cloning here is, that these `Sources` can cause some overhead when we process the cache entries.
I wanted to optimize that first.

gakonst (approved):
took another look and looks good, no additional comment to my previous ones

ethers-solc/src/resolver.rs
mattsse: added test that checks changes in a lib are detected

gakonst:
Let's see if this is the solution to all of our problems!

PR #755, opened 2022-01-01
feat: save compiler version + runs

gakonst:
Saves the compiler version + paths used under `$ARTIFACTS/solidity-versions.json`.

TODO:
- [x] Save the compiler runs.

We need this to do etherscan verification.


gakonst:
Gonna close as stale, but something to keep in mind @mattsse @rkrasiuk. The problem is that when we compile in Foundry if we don't save the compiler version that an artifact was compiled with, we cannot know what version to tell etherscan to use.

PR #756, opened 2022-01-03
feat: add FromStr impl for Chain

odyslam:
Enables https://github.com/gakonst/foundry/pull/357

Ideally, I would prefer that the default case of the `match` statement is sth like `Chain::UnknownChain`, but I added mainnet for compatibility.

Should we add that to the enum?


gakonst (approved):
LGTM. Used `thiserror::Error` for the error generation and made the function return an error when string not found

Issue #751, opened 2021-12-29
Consider making ethers-solc RELEASES offline first


mattsse:
rn the `RELEASES` static does make a http request to retrieve all versions.

alternatively, we can hard code this as a list or even static array

Upside is that we can get rid of the request overhead
downside is manually updating this list with every new release

gakonst:
I am OK with us hardcoding the values, and we can add a test that runs on CI every day which checks if it's out of sync? 

The error is handled right now by the way: https://github.com/gakonst/ethers-rs/blob/debd1dd153bf7f4bfbfecdebb39b7afa426cdd47/ethers-solc/src/compile.rs#L53-L65.

Think there might be some other calls from svm-rs which we do which end up making network requests.

gakonst:
Closed in #759

PR #759, opened 2022-01-05
fix(solc): add timeout and error detection for releases lookup

mattsse:
## Motivation
It seems like `https://binaries.soliditylang.org` is down, so `Releases` requests fail after some time with a DNS lookup error. which highlighted an issue:

a) takes quite a bit of time
b) all checksums fail bc no checksums to compare against

Ref #751

## Solution
* Add timeout to Releases lookup, 3secs is probably reasonable
* only check checksums of we could lookup releases successfully

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3419,11 +3419,10 @@ checksum = "cda74da7e1a664f795bb1f8a87ec406fb89a02522cf6e50620d016add6dbbf5c"
 
 [[package]]
 name = "tokio"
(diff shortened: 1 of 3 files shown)


gakonst (approved):
LGTM pending the failing test

gakonst:
god bless savior PR

PR #765, opened 2022-01-06
feat(solc): add more contract utility types and functions

mattsse:
## Motivation
Add various helper types for dealing with contract's bytecodes

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts.rs
+++ b/ethers-solc/src/artifacts.rs
@@ -6,6 +6,7 @@ use md5::Digest;
 use semver::Version;
 use std::{
     collections::BTreeMap,
+    convert::TryFrom,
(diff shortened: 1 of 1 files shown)


gakonst (approved):
LGTM - feels like we should do a pass at the entire package once we're a bit stabler on Foundry and see if we have duplicate logic / if we can simplify / take the package on a diet as it's starting to get quite big I think

mattsse:
yeah, the artifacts.rs needs another pass even if it's just moving types to a separate file, it's coming in at 2k sloc rn which is a bit much.
should at least extract all contract related types

PR #769, opened 2022-01-07
ethers-addressbook crate

roynalnaruto:
## Motivation

The `ethers-addressbook` crate adds the addresses of frequently used contracts for various networks (mainnet, ropsten, etc.). We expose a function `contract`:
```rust
pub fn contract<S: Into<String>>(symbol: S) -> Option<Contract>;
```

A `Contract` struct is simply a hash map, mapping a network name to that contract's address on that network.
```rust
use ethers_core::types::{Address, Chain};

struct Contract {
    addresses: HashMap<Chain, Address>,
}
```

The goal is to use this crate as dependency in Foundry to allow users to `cast send` transactions targeting contracts such as `DAI`, `USDC` or other commonly used ones as per the below UX:
```
$ cast send dai transfer vitalik.eth 100
```

Future PRs are expected to be limited to simply adding more contracts and/or their addresses for more networks (chain IDs) in the `ethers-addressbook/src/contracts/contracts.json` JSON file.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
LGTM, interested in any thoughts/comments on this from @brockelmore @prestwich @mattsse. We can start hardcoding more addresses, e.g. the ones from MEV Inspect https://github.com/flashbots/mev-inspect-rs/blob/master/src/addresses.rs, or figure out some nice way (maybe a build.rs?) to further populate the addressbook with tokenlist information for token name / ticker / addr

mds1:
Just came across this feature, and really like this idea and how it fits into [`cast`](https://github.com/gakonst/foundry). Two thoughts:
1. Use chain ID instead of networks names. This will remove any chance of ambiguity but also make it easier to seed the address book from a [tokenlist](https://tokenlists.org/)
2. TrueBlocks already maintains a mapping of contract address to names, which you can see by installing it and running `chifra names`. I don't know the details about how that list was generated, where it lives, or how to update it, but it might be a good idea to leverage that and have a global Ethereum address book instead of building up a brand new one from scratch here. cc'ing @tjayrush from TrueBlocks

tjayrush:
Thanks, Matt. Yeah. We've been collecting names and addresses since about 2016, so there are currently ~11,300 names in our database. It's here: https://github.com/TrueBlocks/trueblocks-core/blob/master/src/other/install/names/names.tab

No guarantee that it's accurate (except for the entries that have `on-chain` as their source (~4,750) which we pulled from the token's own on-chain data.

We always wanted this to be public -- we don't feel bad about collecting names if people are going to post them on Twitter for example, so would be totally open to building something that works.

If you look at the rest of TrueBlocks, you'll see we publish a period manifest of the index we build to IPFS. This lets us run this index database without any infrastructure costs. Our code reads the IPFS hash of the manifest from a smart contract and pulls the entire index database from IPFS with no servers.

We're going to be adding the `names` database to that manifest soon, so you're very welcome to read that data as we do.

Here's a short video explaining that process: https://unchainedindex.io/

PR #770, opened 2022-01-07
Add WETH to address book

onbjerg:
Arguably the most used token ever and it's not in the address book! Not sure if we should name it WETH or WETH9 - thoughts?


gakonst (approved):
Let's go with weth for now

PR #772, opened 2022-01-07
Check for warnings after compilation

onbjerg:
## Motivation

Currently `forge` suppresses warnings when building projects (https://github.com/gakonst/foundry/issues/304). This adds relevant methods that we can use to detect warnings there.

### DIFF
--- a/ethers-solc/src/artifacts.rs
+++ b/ethers-solc/src/artifacts.rs
@@ -701,9 +701,11 @@ impl<'a> OutputDiagnostics<'a> {
 
 impl<'a> fmt::Display for OutputDiagnostics<'a> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if self.has_warning() {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

onbjerg:
I assumed that this portion spat out both warnings and errors? :thinking: Maybe this assumption is incorrect https://github.com/gakonst/ethers-rs/blob/0ddcb4c84cc3c4ec2ba65fcce280bd40caaa481b/ethers-solc/src/artifacts.rs#L709-L718

I definitely want it to emit warnings as well in this PR if this is not the case

mattsse:
oh my bad, yeah warnings are also stored in `errors`, doh :)

onbjerg:
Committed a minor fix, think it would technically be possible for it to say "Compiler run successful with warnings" even though you also have high severity errors if you had both errors and warnings

gakonst (approved):

PR #774, opened 2022-01-07
feat(solc): flatten

rkrasiuk:
## Motivation

Implementation of flatten feature discussed in [foundry #312](https://github.com/gakonst/foundry/issues/312)

## Solution

Use the dependency graph to recursively resolve the file imports into a single string.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog


ethers-solc/src/lib.rs
```diff
@@ -481,6 +481,49 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
         }
         Ok(())
     }
+
+    /// Flatten all file imports into a single string
+    pub fn flatten(&self, target: &PathBuf) -> Result<String> {
```
mattsse, line 486: since the graph only requires the `ProjectPathsConfig`, I'd rather move this to `ProjectPathsConfig::flatten`

mattsse (changes requested):
while this approach looks like it could work, I have some suggestions how to improve this a bit.

we're already parsing the solidity file and detecting import statements:

https://github.com/gakonst/ethers-rs/blob/1742630d5f09e6dc0e88bbc6b0ba37cfb7666f07/ethers-solc/src/resolver.rs#L429-L434

all these import variants include a `Loc` value which has the start and end of the whole statement.
So ideally we want to use this data so we can later simply replace that slice `[start..end]` with the content of the file that's imported here.

however since `solang` only targets >0.7.0, it can happen that need to use the regex fall back, so ideally the `utils::find_import_paths` will also return them, which should be possible since the `regex::Match` type has `start` and `end` https://docs.rs/regex/latest/regex/struct.Match.html

If we roll a `struct Import {path, start, end}` for the `SolData` type we should be able to flatten the entire graph.

This depth-first operation where you start with the targeted file and then recursively loop through all imports and their imports... and replace every import statement with content. That's probably a bit tricky since we need to keep track of updated locations because after we replaced [start..end] with the first import the start and end locations of the next import are no longer valid

ethers-solc/src/config.rs
```diff
@@ -146,6 +146,69 @@ impl ProjectPathsConfig {
...
+                let mut i = 0;
+                while i < content.len() {
+                    if let Some(ref import) = curr_import {
+                        let (start, end) = import.loc();
+                        if i == start {
+                            let import_path =
+                                utils::resolve_import_component(import.path(), target_dir, &self)?;
+                            let import_content = (flattener.f)(flattener, &import_path)?;
+                            let import_content =
+                                utils::RE_SOL_PRAGMA_VERSION.replace_all(&import_content, "");
```
rkrasiuk, line 186: @mattsse the location for the pragma version directive is also not exposed, do we want to add another PR to solang?

mattsse: yeah sgtm!

ethers-solc/src/config.rs
```diff
@@ -146,6 +146,67 @@ impl ProjectPathsConfig {
...
     }
+
+    /// Flatten all file imports into a single string
+    pub fn flatten(&self, target: &Path) -> Result<String> {
+        tracing::trace!("flattening file");
+        let graph = Graph::resolve(self)?;
+
+        struct Flattener<'a> {
+            f: &'a dyn Fn(&Flattener, &Path) -> Result<String>,
+        }
```
mattsse, line 157: seems like this could be a standalone separate function?

ethers-solc/src/config.rs
```diff
@@ -146,6 +146,67 @@ impl ProjectPathsConfig {
...
+
+                let mut imports = target_node.imports().clone();
+                imports.sort_by(|a, b| b.loc().0.cmp(&a.loc().0));
+
+                let content = target_node.content().bytes().collect::<Vec<_>>();
+                let mut curr_import = imports.pop();
+                let mut extended = vec![];
+
+                let mut i = 0;
+                while i < content.len() {
```
mattsse, line 179: there should also be `[u8].advance` to shift,
seems like that's what's going on in this loop?

mattsse (changes requested):
great progress!
some suggestions and comments

ethers-solc/src/config.rs
```diff
@@ -171,6 +171,61 @@ impl ProjectPathsConfig {
...
+        let target_dir = target.parent().ok_or_else(|| {
+            SolcError::msg(format!("failed to get parent directory for \"{:?}\"", target.display()))
+        })?;
+        let target_index = graph.files().get(target).ok_or_else(|| {
+            SolcError::msg(format!("cannot resolve file at \"{:?}\"", target.display()))
+        })?;
+        let target_node = graph.node(*target_index);
+
+        let mut imports = target_node.imports().clone();
+        imports.sort_by(|a, b| a.loc().0.cmp(&b.loc().0));
```
mattsse, line 198: this should also work
```suggestion
         imports.sort_by_key(|x| x.loc().0);
```

ethers-solc/src/config.rs
```diff
@@ -171,6 +171,61 @@ impl ProjectPathsConfig {
...
+
+        let mut content = target_node.content().bytes().collect::<Vec<_>>();
```
mattsse, line 200:
```suggestion
        let mut content = target_node.content().as_bytes().to_vec();
```

ethers-solc/src/config.rs
```diff
@@ -171,6 +171,61 @@ impl ProjectPathsConfig {
...
+            }
+        }
+
+        for import in imports.iter() {
+            let import_path = self.resolve_import(target_dir, import.path())?;
+            let import_content = self.flatten_node(&import_path, graph, true)?;
+            let import_content = import_content.trim().as_bytes().to_owned();
+            let import_content_len = import_content.len() as isize;
+            let (start, end) =
+                import.loc_by_offset(offset).expect("failed to determine import location");
```
mattsse, line 217: I wonder if there's a way to make offset `usize`?
perhaps flip it and do `offset +=`?

The offset should keep track of the adjusted length.

Oh, I can see now why this can be negative.

But the cast to size in `loc_by_offset` which requires a Result is a bit redundant if we unwrap it.

can we do loc_by_offset directly here, because we're also casting here again
`let import_offset = import_content_len - ((end - start) as isize);`

there should also be `isize::saturated_sub()` and `saturated_add`

mattsse (changes requested):
few nits,

overall this is looking good,
I'd like to see couple more tests, for the dapp/hardhat examples
and perhaps you could try manually try this with some real repos and see if that works

mattsse (approved):
lgtm!

gakonst (approved):
test failure is unrelated, merging and we can iterate as we close the feedback loop with usage in foundry.

is there any follow up that needs to be done which we know we need? probably need to get the solidity compiler version saved in the artifacts (https://github.com/gakonst/ethers-rs/pull/755)

PR #775, opened 2022-01-08
Ignore SPDX and contract size for tests

onbjerg:
## Motivation

Following discussion in https://github.com/gakonst/foundry/pull/399 we want to be able to ignore some errors for tests and dependencies, but not for source.

## Solution

This PR just ignores 2 errors for both tests and dependencies by default: the SPDX and contract size warnings.

### DIFF
--- a/ethers-solc/src/artifacts.rs
+++ b/ethers-solc/src/artifacts.rs
@@ -537,18 +537,17 @@ impl CompilerOutput {
     }
 
     /// Whether the output contains a compiler warning
-    pub fn has_warning<'a>(&self, ignored_error_codes: &'a [u64]) -> bool {
+    pub fn has_warning(&self, ignored_error_codes: &[u64]) -> bool {
         self.errors.iter().any(|err| {
-            let is_ignored = err.error_code.as_ref().map_or(false, |code| {
-                !ignored_error_codes.contains(code)
-            });
+            let is_ignored =
+                err.error_code.as_ref().map_or(false, |code| !ignored_error_codes.contains(code));
 
             err.severity.is_warning() && !is_ignored
         })
     }
 
     pub fn diagnostics<'a>(&'a self, ignored_error_codes: &'a [u64]) -> OutputDiagnostics {
-        OutputDiagnostics { compiler_output: &self, ignored_error_codes }
+        OutputDiagnostics { compiler_output: self, ignored_error_codes }
     }
 
     /// Finds the _first_ contract with the given name
@@ -701,20 +700,17 @@ impl<'a> OutputDiagnostics<'a> {
 
     /// Returns true if there is at least one warning
     pub fn has_warning(&self) -> bool {
-        self.compiler_output.has_warning(&self.ignored_error_codes)
+        self.compiler_output.has_warning(self.ignored_error_codes)
     }
 
     fn is_test<T: AsRef<str>>(&self, contract_path: T) -> bool {
         if contract_path.as_ref().ends_with(".t.sol") {
-            return true;
+            return true
         }
 
-        self.compiler_output.find(&contract_path)
-            .map_or(false, |contract| {
-                contract.abi.map_or(false, |abi| {
-                    abi.functions.contains_key("IS_TEST")
-                })
-            })
+        self.compiler_output.find(&contract_path).map_or(false, |contract| {
+            contract.abi.map_or(false, |abi| abi.functions.contains_key("IS_TEST"))
+        })
     }
 }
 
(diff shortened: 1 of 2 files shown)


ethers-solc/src/artifacts.rs
```diff
@@ -708,13 +727,22 @@ impl<'a> fmt::Display for OutputDiagnostics<'a> {
...
-                if !self.ignored_error_codes.contains(&error_code) {
-                    writeln!(f, "\n{}", err)?;
+        for err in &self.compiler_output.errors {
+            let is_ignored = err.error_code.as_ref().map_or(false, |code| {
+                if let Some(source_location) = &err.source_location {
+                    // we ignore spdx and contract size warnings in test
+                    // files. if we are looking at one of these warnings
+                    // from a test file we skip
+                    if self.is_test(&source_location.file)
+                        && (*code == 1878 || *code == 5574) {
```
onbjerg, line 737: Not super keen on this, but I am not entirely sure what direction to go in with fully configurable error codes based on types either in terms of DX or configuration. Is this fine for now or does anyone have any pointers on how to make this more elegant?

ethers-solc/src/artifacts.rs
```diff
@@ -537,12 +537,18 @@ impl CompilerOutput {
...
-    pub fn has_warning(&self) -> bool {
-        self.errors.iter().any(|err| err.severity.is_warning())
+    pub fn has_warning<'a>(&self, ignored_error_codes: &'a [u64]) -> bool {
+        self.errors.iter().any(|err| {
+            let is_ignored = err.error_code.as_ref().map_or(false, |code| {
+                !ignored_error_codes.contains(code)
+            });
+
+            err.severity.is_warning() && !is_ignored
+        })
```
onbjerg, line 547: Realized we had to filter for errors here too

ethers-solc/src/artifacts.rs
```diff
@@ -683,19 +689,32 @@ impl OutputContracts {
...
+        self.compiler_output.has_error()
     }
 
     /// Returns true if there is at least one warning
     pub fn has_warning(&self) -> bool {
-        self.errors.iter().any(|err| err.severity.is_warning())
+        self.compiler_output.has_warning(&self.ignored_error_codes)
+    }
+
+    fn is_test<T: AsRef<str>>(&self, contract_path: T) -> bool {
```
onbjerg, line 707: Not sure how robust this is and I couldn't really figure out a good way to do it for deps, should the test be "if it sits in node_modules or lib then it is a dependency"?

ethers-solc/src/artifacts.rs
```diff
@@ -537,12 +537,17 @@ impl CompilerOutput {
     }
 
     /// Whether the output contains a compiler warning
-    pub fn has_warning(&self) -> bool {
-        self.errors.iter().any(|err| err.severity.is_warning())
+    pub fn has_warning(&self, ignored_error_codes: &[u64]) -> bool {
+        self.errors.iter().any(|err| {
+            let is_ignored =
+                err.error_code.as_ref().map_or(false, |code| !ignored_error_codes.contains(code));
```
mattsse, line 543: I'd be more efficient to compute this only if `err.severity.is_warning() == true`

ethers-solc/src/artifacts.rs
mattsse: hmm, one idea I have is to leave this completely up to the user,

rn we're using it like this:

https://github.com/gakonst/ethers-rs/blob/371afb9dff2041e8ec8d6d808ce5bf18699b04da/ethers-solc/src/lib.rs#L864

perhaps we can add something like 

```rust
 filter: dyn ContractFilter

trait ContractFilter {
   fn matches(&self, contract: &Contract) -> bool;
}

impl ContractFilter for () {
  fn matches(&self) -> boo {true}
}
```

then implement this for `Fn(&str) -> bool`

so we can add a `CompilerOutput::diagnostics_with_filter(error_codes, dyn ContractFilter)`

onbjerg: we could also do that for filtering of warnings in general, so instead of a contract filter etc. we just give the user

- `has_warnings`
- `has_errors`
- a way to iterate the errors and filter them themselves

unless there is a specific reason we have this in ethers-solc as opposed to foundry?

mattsse: sg, so we should map each error to its contract and ignore an error if:

* ignored error code
* contract should be ignored

gakonst (approved):
LGTM

PR #776, opened 2022-01-08
make it work with latest solang

gelfand:


ethers-solc/src/resolver.rs
```diff
@@ -33,7 +33,7 @@ use std::{
 
 use rayon::prelude::*;
 use semver::VersionReq;
-use solang_parser::pt::{Import, SourceUnitPart};
```
mattsse, line 36: don't think that's needed

ethers-solc/src/resolver.rs
```diff
@@ -282,7 +282,7 @@ impl Graph {
                     utils::source_name(&self.nodes[idx].path, &self.root).display(),
                     utils::source_name(&self.nodes[dep].path, &self.root).display()
                 );
-                continue
+                continue;
```
mattsse, line 285: please use nightly formatting `cargo +nightly fmt`

mattsse (changes requested):
thanks,
two nits :)

ethers-solc/src/resolver.rs
gelfand: my bad (

gakonst (approved):

PR #777, opened 2022-01-08
Implement `is_empty` for `AllowedLibPaths`

brockelmore:
## Motivation
Some foundry tooling needs to check if allowed lib paths is empty

## Solution
Implement the method

### DIFF
--- a/ethers-solc/src/config.rs
+++ b/ethers-solc/src/config.rs
@@ -550,6 +550,12 @@ impl ArtifactOutput for MinimalCombinedArtifactsHardhatFallback {
 #[derive(Clone, Debug, Default)]
 pub struct AllowedLibPaths(pub(crate) Vec<PathBuf>);
 
+impl AllowedLibPaths {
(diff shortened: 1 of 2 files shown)


ethers-solc/src/config.rs
```diff
@@ -550,6 +550,12 @@ impl ArtifactOutput for MinimalCombinedArtifactsHardhatFallback {
 #[derive(Clone, Debug, Default)]
 pub struct AllowedLibPaths(pub(crate) Vec<PathBuf>);
 
+impl AllowedLibPaths {
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
```
mattsse, line 556: iirc there's a clippy lint that checks that `fn is_empty` and `fn len` come in pairs.
alternatively, we can `impl AsRef` and/or `AsMut`, or simply make it pub

mattsse (approved):
lgtm

PR #782, opened 2022-01-10
feat(chain): enum values + TryFrom<u64>

tarrencev:
Upstream changes to support feedback at https://github.com/gakonst/foundry/pull/412

cc @mattsse

### DIFF
--- a/ethers-core/src/types/chain.rs
+++ b/ethers-core/src/types/chain.rs
@@ -1,6 +1,7 @@
 use serde::Deserialize;
 use thiserror::Error;
 
+use core::convert::TryFrom;
 use std::{fmt, str::FromStr};
 
 use crate::types::U256;
@@ -8,7 +9,9 @@ use crate::types::U256;
 #[derive(Debug, Clone, Error)]
 #[error("Failed to parse chain: {0}")]
 pub struct ParseChainError(String);
+pub struct TryFromUnknownChain;
 
+#[repr(u64)]
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash, Deserialize)]
 #[serde(rename_all = "snake_case")]
 pub enum Chain {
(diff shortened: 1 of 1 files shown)


ethers-core/src/types/chain.rs
```diff
@@ -1,33 +1,36 @@
 use serde::Deserialize;
 use thiserror::Error;
 
+use core::convert::TryFrom;
```
mattsse, line 4: I believe, as of 2021 this should be imported by default?

ethers-core/src/types/chain.rs
```diff
@@ -71,6 +57,32 @@ impl From<Chain> for u64 {
     }
 }
 
+impl TryFrom<u64> for Chain {
+    type Error = TryFromUnknownChain;
```
mattsse, line 61: would be ok to just make this a `()`,
which makes it essentially an Option

mattsse (changes requested):

ethers-core/src/types/chain.rs
tarrencev: Hmm ok my IDE was complaining but maybe it is outdated

ethers-core/src/types/chain.rs
tarrencev: I'm not sure I understand, do you mind elaborating?

mattsse: the error type is not really useful here since it's obvious, 
so we could simply make this `type Error = ();`

ethers-core/src/types/chain.rs
```diff
@@ -71,6 +57,32 @@ impl From<Chain> for u64 {
     }
 }
 
+impl TryFrom<u64> for Chain {
+    type Error = ();
+
+    fn try_from(chainid: u64) -> Result<Chain, Self::Error> {
```
tarrencev, line 63: we could use a crate like https://crates.io/crates/num_enum to avoid duplicating the mapping here

gakonst: i think it's fine as-is, if we see the maintenance of this becoming a problem we can consider

Issue #727, opened 2021-12-23
solc: Emit all artifacts during cached build


mattsse:
Currently the `Artifacts` iterator may omit artifacts of lib contracts during a cached run and only yield those that correspond to the files found in the configured sources folder.

PR #783, opened 2022-01-11
fix(solc): invalid cached artifacts

mattsse:
## Motivation
https://github.com/gakonst/foundry/issues/316

the way we checked if we need to recompile a contract was too optimistic and didn't check for their usage properly so in https://github.com/gakonst/foundry/issues/316 we only recompiled the base contract and not the contract that inherits.
This was fixed but should be improved, generally the whole compile <-> cache pipeline is ripe for a v2 rewrite that uses the new dependency path.

I also discovered and fixed the issue that's likely related to #727, we used different keys for the `into_artifacts` iterator, while for cached artifacts we used `<file>:<name>` we only used the name for recompiled assets. but #727 needs an additional test as follow-up.

## Solution
resolve all imports and check if they were changed.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/cache.rs
+++ b/ethers-solc/src/cache.rs
@@ -3,11 +3,11 @@ use crate::{
     artifacts::{Contracts, Sources},
     config::SolcConfig,
     error::{Result, SolcError},
-    utils, ArtifactOutput,
+    utils, ArtifactOutput, ProjectPathsConfig,
 };
 use serde::{Deserialize, Serialize};
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::{BTreeMap, HashMap, HashSet},
     fs::{self, File},
     path::{Path, PathBuf},
     time::{Duration, UNIX_EPOCH},
@@ -133,17 +133,15 @@ impl SolFilesCache {
         &'a self,
         sources: Sources,
         config: Option<&'a SolcConfig>,
-        artifacts_root: &Path,
+        paths: &ProjectPathsConfig,
     ) -> Sources {
+        // all file hashes
+        let content_hashes: HashMap<_, _> =
+            sources.iter().map(|(file, source)| (file.clone(), source.content_hash())).collect();
         sources
             .into_iter()
-            .filter(move |(file, source)| {
-                self.has_changed_or_missing_artifact::<T>(
-                    file,
-                    source.content_hash().as_bytes(),
-                    config,
-                    artifacts_root,
-                )
+            .filter(move |(file, _)| {
+                self.has_changed_or_missing_artifact::<T>(file, &content_hashes, config, paths)
             })
             .collect()
     }
@@ -153,10 +151,11 @@ impl SolFilesCache {
     pub fn has_changed_or_missing_artifact<T: ArtifactOutput>(
         &self,
         file: &Path,
-        hash: &[u8],
+        hashes: &HashMap<PathBuf, String>,
         config: Option<&SolcConfig>,
-        artifacts_root: &Path,
+        paths: &ProjectPathsConfig,
     ) -> bool {
+        let hash = hashes.get(file).unwrap().as_bytes();
         if let Some(entry) = self.files.get(file) {
             if entry.content_hash.as_bytes() != hash {
                 tracing::trace!("changed content hash for cached artifact \"{}\"", file.display());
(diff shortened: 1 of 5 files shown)


ethers-solc/tests/project.rs
```diff
@@ -137,6 +137,90 @@ fn can_compile_dapp_detect_changes_in_libs() {
     assert!(!compiled.is_unchanged());
 }
 
+#[test]
+fn can_compile_dapp_detect_changes_in_sources() {
+    let project = TempProject::<MinimalCombinedArtifacts>::dapptools().unwrap();
+
+    let src = project
+        .add_source(
```
gakonst, line 145: this technique makes test building so much easier

ethers-solc/src/lib.rs
```diff
@@ -844,16 +870,22 @@ impl<T: ArtifactOutput + 'static> ProjectCompileOutput<T> {
     /// ```
     pub fn into_artifacts(mut self) -> Box<dyn Iterator<Item = (String, T::Artifact)>> {
         let artifacts = self.artifacts.into_iter().filter_map(|(path, art)| {
-            T::contract_name(&path)
-                .map(|name| (format!("{:?}:{}", path.file_name().unwrap(), name), art))
+            T::contract_name(&path).map(|name| {
+                (format!("{}:{}", path.file_name().unwrap().to_string_lossy(), name), art)
+            })
```
gakonst, line 875: Was there a bug here and you do `to_string_lossy`? I guess the issue would be that due to the `{:?}` the key becomes a quoted string, which ends up having a different key from `{}:{}`

ethers-solc/src/lib.rs
```diff
@@ -844,16 +870,22 @@ impl<T: ArtifactOutput + 'static> ProjectCompileOutput<T> {
...
+            self.compiler_output.take()
+        {
+            Box::new(artifacts.chain(T::output_to_artifacts(output).into_values().flatten().map(
+                |(name, artifact)| {
+                    (format!("{}:{}", T::output_file_name(&name).display(), name), artifact)
+                },
+            )))
+        } else {
+            Box::new(artifacts)
+        };
```
gakonst, line 888: Ah good catch
> we used different keys for the into_artifacts iterator, while for cached artifacts we used <file>:<name> we only used the name for recompiled assets

ethers-solc/tests/project.rs
```diff
@@ -137,6 +137,90 @@ fn can_compile_dapp_detect_changes_in_libs() {
...
+    let compiled = project.compile().unwrap();
+    assert!(compiled.find("DssSpellTest").is_some());
+    assert!(compiled.find("DssSpellTestBase").is_some());
+    // ensure change is detected
+    assert!(!compiled.is_unchanged());
+    // and all recompiled artifacts are different
+    for (p, artifact) in compiled.into_artifacts() {
+        let other = artifacts.remove(&p).unwrap();
+        assert_ne!(artifact, other);
+    }
```
gakonst, line 221: :+1:

ethers-solc/src/cache.rs
```diff
@@ -153,10 +151,11 @@ impl SolFilesCache {
     pub fn has_changed_or_missing_artifact<T: ArtifactOutput>(
         &self,
         file: &Path,
-        hash: &[u8],
+        hashes: &HashMap<PathBuf, String>,
         config: Option<&SolcConfig>,
-        artifacts_root: &Path,
+        paths: &ProjectPathsConfig,
     ) -> bool {
+        let hash = hashes.get(file).unwrap().as_bytes();
```
gakonst, line 158: This `unwrap` should be safe if used with `get_changed_or_missing_artifacts_files`, but could panic if somebody did not construct the map properly, should we handle the `None` case, or should we restrict the visibility of this function so it can never be called in an unsafe way?

gakonst (approved):

ethers-solc/src/cache.rs
mattsse: good catch, yes there's not guarantee, will revise

ethers-solc/src/lib.rs
mattsse: this added quotes for the file path like `\"file\":name`

PR #789, opened 2022-01-13
fix(abi): check for uint8 params in human readable types

mattsse:
## Motivation
Based on https://github.com/rust-ethereum/ethabi/issues/254 unknown identifiers are now treated as uint8, this broke how the `AbiParser` resolves structs in a human readable ABI. because from pov of ethabi's `Reader`:

```solidity
        struct Bar { uint256 x; }
        function foo(Bar memory x)
```

type `Bar` in function `foo` is now a `unit8`.

## Solution
if `ethabi::Reader` returns an `Uint8` check that the input string also includes a `uint8`, otherwise treat it as struct/tuple.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/abi/human_readable.rs
+++ b/ethers-core/src/abi/human_readable.rs
@@ -365,7 +365,11 @@ impl AbiParser {
     /// See https://github.com/rust-ethereum/ethabi/issues/254
     ///
     /// Therefore, we need to double-check if the `ethabi::Reader` parsed an `uint8`, and ignore the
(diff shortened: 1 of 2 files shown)


ethers-core/src/abi/human_readable.rs
```diff
@@ -463,6 +511,30 @@ fn detect_state_mutability(s: &str) -> StateMutability {
...
+        false
+    }
+}
+
+/// Returns true if the `ParamType` contains an `uint8`
+pub fn contains_uint8(kind: &ParamType) -> bool {
+    match kind {
+        ParamType::Uint(8) => true,
+        ParamType::Array(arr) => contains_uint8(&*arr),
+        ParamType::FixedArray(arr, _) => contains_uint8(&*arr),
```
gakonst, line 532: Think this is better naming? The inner param is not an array itself, it's the inner `ParamType`. It'd be an array if they're nested I guess.

```suggestion
        ParamType::Array(kind) => contains_uint8(&*kind),
        ParamType::FixedArray(kind, _) => contains_uint8(&*kind),
```

gakonst (approved):
lgtm subject to the unit test fix + maybe addressing my small nit - good catch.

gakonst (approved):

PR #792, opened 2022-01-14
Implement trace_callMany

gelfand:
## Motivation

Implement [trace_callMany](https://openethereum.github.io/JSONRPC-trace-module#trace_callmany)

## Solution

Actually do this. I will also need to implement `ErigonInstance` so I can add test for this.
Now it actually works with my local node
```sh
◆ ethers-fork git:(master) ✗ ❯❯❯ cargo test -p ethers-providers --lib -- --nocapture provider::tests::test_trace_call_many
   Compiling ethers-providers v0.6.0 (/Users/eugene/Developer/rust/ethers-fork/ethers-providers)
    Finished test [unoptimized + debuginfo] target(s) in 3.52s
     Running unittests (target/debug/deps/ethers_providers-6eb7cba4c6e73967)

running 1 test
[ethers-providers/src/provider.rs:1485] traces = [
    BlockTrace {
        output: Bytes(
            b"",
        ),
        trace: Some(
            [],
        ),
        vm_trace: None,
        state_diff: Some(
            StateDiff(
                {
                    0x0000000000000000000000000000000000000000: AccountDiff {
                        balance: Changed(
                            ChangedType {
                                from: 11326881149672971579810,
                                to: 1553734685634137288388564520290,
                            },
                        ),
                        nonce: Changed(
                            ChangedType {
                                from: 0,
                                to: 1,
                            },
                        ),
                        code: Same,
                        storage: {},
                    },
                    0x0000000000000000000000000000000000000001: AccountDiff {
                        balance: Changed(
                            ChangedType {
                                from: 13629332194864958173,
                                to: 13639332194864958173,
                            },
                        ),
                        nonce: Same,
                        code: Same,
                        storage: {},
                    },
                },
            ),
        ),
        transaction_hash: None,
    },
    BlockTrace {
        output: Bytes(
            b"\xe3\xb0\xc4B\x98\xfc\x1c\x14\x9a\xfb\xf4\xc8\x99o\xb9$'\xaeA\xe4d\x9b\x93L\xa4\x95\x99\x1bxR\xb8U",
        ),
        trace: Some(
            [],
        ),
        vm_trace: None,
        state_diff: Some(
            StateDiff(
                {
                    0x0000000000000000000000000000000000000001: AccountDiff {
                        balance: Changed(
                            ChangedType {
                                from: 13639332194864958173,
                                to: 1553734674320895966171835647553,
                            },
                        ),
                        nonce: Changed(
                            ChangedType {
                                from: 0,
                                to: 1,
[...]

PR #797, opened 2022-01-16
test for trace_callMany

gelfand:
continuation of #792

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,19 +46,14 @@ jobs:
           geth version
       - name: Install Erigon
         run: |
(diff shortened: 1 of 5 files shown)


gakonst (changes requested):
Looks good, you also need to install Erigon in CI! Will merge after  that

gelfand:
okay, I will prebuilt binaries and upload them to my server, and then just gonna pull them

gakonst:
It might be possible the binary isn't getting detected, or that you're downloading a binary that is not matching the platform of the gha runner?

gelfand:
> It might be possible the binary isn't getting detected, or that you're downloading a binary that is not matching the platform of the gha runner?

I built it for Linux x86-64, tested on another Linux machine even without Go being installed and it was fine.
I'm gonna give a try again to test this binary on freshly installed linux.

PR #800, opened 2022-01-17
fix(solc): traverse nodes iteratively

mattsse:
## Motivation
Nodes were traversed recursively which could Hit a SO on large repos, like https://github.com/fei-protocol/tribe-turbo

## Solution
Use an iterator `NodesIter` instead that yields all nodes instead

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/resolver.rs
+++ b/ethers-solc/src/resolver.rs
@@ -27,7 +27,7 @@
 //! which is defined on a per source file basis.
 
 use std::{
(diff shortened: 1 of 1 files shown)


gakonst (approved):
lgtm - it does seem like these SO's are hard to unit test, so guess the best way is to add such repos in the integration tests in upstream foundry, or add integration tests for ethers-solc itself

PR #808, opened 2022-01-18
Move solc install lock into install functions

onbjerg:
We call `Solc::install` and `Solc::blocking_install` elsewhere in `ethers-solc`, and I think that might be causing the issue in https://github.com/gakonst/foundry/issues/442

Moved the lock into those functions instead


gakonst (approved):
pray

gakonst:
This seemed to cause a deadlock, so reset--hard on master to before it.

Issue #791, opened 2022-01-14
Change artifact naming conventions


lattejed:
This was prompted by the forge issue here: https://github.com/gakonst/foundry/issues/392

Two overlapping issues:

1. Some tests are being executed but dropped by forge due to naming conventions (that are decided upstream here). This can be fixed w/in forge, however:

2. Contract names are under-specified here, where duplicate names across different files are presented ambiguously as filenames / paths are not retained. 

Example, this is how dapptools presents test names:

`<path>/<filename>:<name>`

forge:

`<filename>:<name>`

Example test output:

dapptools:
```
Running 3 tests for src/test/Greeter.t.sol:Greet
[PASS] testCannotGm() (gas: 4540)
...

Running 3 tests for src/test/Greeter_2.t.sol:Greet
[PASS] testCannotGm() (gas: 4540)
...
```

forge:
```
Running 3 tests for Greet.json:Greet
[PASS] testCannotGm() (gas: 6819)
```
^ both `Greeter` and `Greeter_2` are run by forge, but one is dropped from presentation due to how result are collected internally.

Using dapptools-style output is unambiguous. I'm proposing either changing naming conventions here to match dapptools. This would be a breaking change.

It may also make sense to keep things as-is but expose path, etc. so they can be consumed downstream. 

--

As @mattsse has pointed out, currently artifacts from different compiler versions are not handled correctly. This is a caching, not a presentation issue, but can be addressed at the same time. 

He also suggested having some / all of this configurable, tying in:

https://github.com/gakonst/foundry/blob/f9007fa8c238ffbb1365b26460a949ead303c046/config/README.md

@mattsse I'm not sure if you're thinking making output naming configurable or caching or both?

Any / all input welcome. I'll work on this myself.

lattejed:
@gakonst @mattsse to what degree is the format of `SolFilesCache` sacred? I understand that it's an internal foundry format, but it's also currently (afaik) 1:1 with hardhat's format. 

Dealing with artifact naming collisions wrt different version pragmas, we necessarily end up with artifacts arranged sth like:

```
out/
|-- 0.8.10xxx/src/
|---------------- A.sol/A.json
|---------------- C.sol/C.json
|-- 0.8.11xxx/src/
|---------------- B.sol/B.json
|---------------- C.sol/C.json
```

But with only one entry for `C.sol` in `SolFilesCache` since the entries are keyed to source paths -- none of the entries have a direct reference to the artifact(s).

[...]

PR #802, opened 2022-01-17
refactor(solc): rewrite compiler passes and cache change detection

mattsse:
## Motivation
Close #784
Close #727

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

PR #809, opened 2022-01-18
Change artifact naming conventions

lattejed:
Will address #791 

- [x] Fix downstream issue of ambiguous contract naming
- [x] Handle collision of artifacts compiled with different version
- [x] Add convenience methods so the cache can be queried for artifact compile info (optim. runs, comp. version)
- [ ] ~~Support foundry config file~~

Currently this changes `ProjectCompileOutput::into_artifacts` to return dapptools-style contract names. Example:

`Greet.json:Greet`

becomes:

`src/test/Greeter.t.sol:Greet`

This fixes [this issue](https://github.com/gakonst/foundry/issues/392) w/o any material changes on the foundry side.

Internally this does two main things:

1. Changes how output contract paths are created
2. Preserves relative paths within the artifacts directory

#\2 is to be able to re-map artifact paths to source paths. During compilation, source paths are discarded and since artifacts don't carry metadata, there's no clean way to carry them around. Keeping relative paths in the artifacts dir solves this.

@mattsse let me know if you think #\2 is problematic. The flat structure of the artifacts dir seems unnecessary to me, but I may be missing something.

We could also add compiler versions to paths to avoid the related collision issue.

WIP

### DIFF
--- a/ethers-solc/src/config.rs
+++ b/ethers-solc/src/config.rs
@@ -500,28 +500,6 @@ pub trait ArtifactOutput {
             .unwrap_or_else(|| Self::output_file_name(name))
     }
 
-    // TODO: Rename this
-    // this should output `src/path[/version]/contract.sol/contract.json`
-    // returns `contract.json` by default
-    fn output_file2(
-        source_file: impl AsRef<Path>,
-        version: Option<impl AsRef<str>>,
-        name: impl AsRef<str>,
-        root: impl AsRef<Path>,
-    ) -> PathBuf {
-        let mut path = PathBuf::new();
-        if let Some(src_path) =
-            source_file.as_ref().strip_prefix(root).ok().and_then(|p| p.parent())
-        {
-            path.push(src_path);
-        }
-        if let Some(version) = version {
-            path.push(version.as_ref());
-        }
-        path.push(Self::output_file_name(name));
-        path
-    }
-
     /// The inverse of `contract_file_name`
     ///
     /// Expected to return the solidity contract's name derived from the file path
@@ -578,7 +556,11 @@ pub trait ArtifactOutput {
             .collect()
     }
 
-    /// TODO: doc
+    /// This returns a maybe versioned artifact path, e.g., either
+    /// `Greeter.sol/Greeter.json` or `Greeter.sol/Greeter.version-string-xxxx.json`
+    /// depending on whether or not `version` is passed in. This also retuns
+    /// the same with the relative source path preserved.
+    /// Defaults to `Greeter.json` if there are any errors.
     fn versioned_artifact_path(
         source_path: impl AsRef<Path>,
         version: Option<impl AsRef<str>>,
@@ -601,7 +583,10 @@ pub trait ArtifactOutput {
         (rel_path.join(art_path.clone()), contract_name.join(art_path))
     }
 
-    /// TODO: doc
+    /// This returns a set of maybe versioned artifact path, e.g., either
+    /// `Greeter.sol/Greeter.json` or `Greeter.sol/Greeter.version-string-xxxx.json`
+    /// depending on whether or not multiple versions exist for an artifact.
+    /// Defaults to `Greeter.json` if there are any errors.
     fn versioned_artifact_paths(
         outputs: Vec<(&CompilerOutput, String)>,
         source_dir: impl AsRef<Path>,
--- a/ethers-solc/src/config.rs
+++ b/ethers-solc/src/config.rs
@@ -493,6 +493,30 @@ pub trait ArtifactOutput {
             .unwrap_or_else(|| Self::output_file_name(name))
     }
 
+    // TODO: Rename this
+    // this should output `src/path[/version]/contract.sol/contract.json`
+    // returns `contract.json` by default
+    fn output_file2(
+        source_file: impl AsRef<Path>,
+        version: Option<impl AsRef<str>>,
+        name: impl AsRef<str>,
+        root: impl AsRef<Path>
+    ) -> PathBuf {
+        let mut path = PathBuf::new();
+        if let Some(src_path) = source_file
+            .as_ref()
+            .strip_prefix(root)
+            .ok()
+            .and_then(|p| p.parent()) {
+            path.push(src_path);
+        }
+        if let Some(version) = version {
+            path.push(version.as_ref());
+        }
+        path.push(Self::output_file_name(name));
+        path
+    }
+
     /// The inverse of `contract_file_name`
     ///
     /// Expected to return the solidity contract's name derived from the file path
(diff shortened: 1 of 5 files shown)


lattejed:
FYI, I'm currently seeing two test failures locally:

`can_abiencoderv2_output`
`can_handle_underscore_functions`

I think it's a version mismatch for `ganache` as the internal ganache read loop isn't getting expected input and hangs

lattejed:
@gakonst I'm not sure how much overlap there is here with @mattsse 's rewrite, but if we ship this now it will support https://github.com/gakonst/ethers-rs/pull/828 and fix https://github.com/gakonst/foundry/issues/392

This breaks some tests downstream. Those appear to be minor changes and I'll look at them next.

lattejed:
Actually I've got a couple more commits incoming on this.

lattejed:
Ok, that should be it

ethers-solc/src/config.rs
```diff
@@ -468,7 +470,12 @@ pub trait ArtifactOutput {
     type Artifact: Artifact + DeserializeOwned;
 
     /// Handle the compiler output.
-    fn on_output(output: &CompilerOutput, layout: &ProjectPathsConfig) -> Result<()>;
+    fn on_output(
+        output: &CompilerOutput,
+        version: String,
```
rkrasiuk, line 475:
```suggestion
        version: &str,
```
prefer references over taking ownership. then in any `BTreeMap<_>::get` call down the road you can just use `.to_owned`

ethers-solc/src/config.rs
```diff
@@ -547,6 +555,74 @@ pub trait ArtifactOutput {
...
+        let art_path = version
+            .map(|ver| Self::output_file_name(format!("{}.{}", name.as_ref(), ver.as_ref())))
+            .unwrap_or_else(|| Self::output_file_name(name));
+        let rel_path = source_path
+            .as_ref()
+            .strip_prefix(source_dir.as_ref())
+            .map(PathBuf::from)
+            .unwrap_or_else(|_| art_path.clone());
+        (rel_path.join(art_path.clone()), contract_name.join(art_path))
+    }
```
rkrasiuk, line 584: nit: saves you from redundant `as_ref` calls 
```suggestion
    fn versioned_artifact_path(
        source_path: impl AsRef<Path>,
        version: Option<&str>,
        name: &str,
        source_dir: impl AsRef<Path>,
    ) -> (PathBuf, PathBuf) {
        let contract_name = source_path
            .as_ref()
            .file_name()
            .map(PathBuf::from)
            .unwrap_or_else(|| PathBuf::from(name));
        let art_path = version
            .map(|ver| Self::output_file_name(format!("{}.{}", name, ver)))
            .unwrap_or_else(|| Self::output_file_name(name));
        let rel_path = source_path
            .as_ref()
            .strip_prefix(source_dir)
            .map(PathBuf::from)
            .unwrap_or_else(|_| art_path.clone());
        (rel_path.join(&art_path), contract_name.join(&art_path))
    }
```

ethers-solc/src/lib.rs
```diff
@@ -332,20 +344,49 @@ impl<Artifacts: ArtifactOutput> Project<Artifacts> {
...
-            let output = res?;
+        let all_output = outputs
+            .outputs()
+            .filter_map(|(res, solc, input)| res.as_ref().ok().map(|output| (output, solc, input)))
+            .map(|(output, solc, input)| solc.version().map(|ver| (output, ver.to_string(), input)))
+            .collect::<Result<Vec<_>>>()?;
+
+        let outputs_vers = all_output
+            .iter()
+            .cloned()
```
rkrasiuk, line 355: doesn't seem like `cloned` is necessary here
```suggestion
```

ethers-solc/src/lib.rs
```diff
@@ -885,24 +964,29 @@ impl<T: ArtifactOutput + 'static> ProjectCompileOutput<T> {
...
-        } else {
-            Box::new(artifacts)
-        };
-        artifacts
+        let cache = self.cache_path.and_then(|path| SolFilesCache::read(path).ok());
+        let mut out_artifacts = Vec::new();
+        for (path, art) in self.artifacts {
+            if let Some(name) = T::contract_name(&path) {
+                if let Some(src_name) =
+                    cache.clone().and_then(|cache| cache.source_name_for_artifact(&name))
```
rkrasiuk, line 972:
```suggestion
                    cache.as_ref().and_then(|cache| cache.source_name_for_artifact(&name))
```

ethers-solc/src/lib.rs
```diff
@@ -885,24 +964,29 @@ impl<T: ArtifactOutput + 'static> ProjectCompileOutput<T> {
...
+                {
+                    out_artifacts.push((format!("{}:{}", src_name.display(), name), art))
+                }
+            }
+        }
+        if let Some(output) = self.compiler_output.take() {
+            for (file, artifacts) in T::output_to_artifacts(output) {
+                for (name, artifact) in artifacts {
+                    if let Some(src_name) =
+                        cache.clone().and_then(|cache| cache.source_name_for_source(&file))
```
rkrasiuk, line 982:
```suggestion
                        cache.as_ref().and_then(|cache| cache.source_name_for_source(&file))
```

rkrasiuk (commented):
good job overall! one test seems to be failing for me `cache::tests::can_parse_solidity_files_cache` though, it says `artifactPaths` field is missing

mattsse (commented):
this is going in the right direction,
like the inclusion of Version in the output handler.
what we need to do is also including this info in the CacheEntry  as discussed

mattsse:
> Preserves relative paths within the artifacts directory

I very much in favor of this, paths should always be relative to the project's root

mattsse:
yeah, but this on a per file basis, which means we currently have a conflict if we compile a file with two different versions so there is no way to go from the cacheentry to the format here:
https://github.com/gakonst/ethers-rs/issues/791#issuecomment-1017573205

lattejed:
Artifacts are mapped to the version they were compiled with in `artifact_paths`:

`output path`: `version string`

lattejed:
To elaborate a bit more here:

Artifacts are compiled with a specific compiler version, and paths are versioned, so there's a 1:1 mapping from artifact path -> compiler version. The `SolcConfig` is the input to the compiler and is the same for each artifact version, even if the compiler was different. Everything else in `CacheEntry` is related to the source file, which of course is singular.

The only thing that seems iffy to me is that `SolcConfig` / compiler input takes a normalized compiler version so (technically?) there could be a case where the compiler version differed enough between compiles (for different includes) that the normalized version changed, but I'd guess that won't be an issue in practice. Maybe I'm wrong there.

I don't have a strong opinion about how to format the cache file, but afaik what I've put together here includes all necessary info. If you do want to change it let me know.

mattsse:
ah gotcha, I can follow now.
good progress, this is helpful, tinkering with it rn

lattejed:
FYI, there's an issue here with `into_artifacts` when the compiler is being run in 'test mode'. It's not caught by tests in ethers but is caught by tests in foundry. Needs to be fixed of course but ethers should also have some tests added to cover it.

gakonst:
@lattejed @mattsse how can we keep up momentum in this PR?

lattejed:
I understood that ***@***.***(https://github.com/mattsse) was going to cherry pick this for his rewrite. If you want to use this it needs to be changed to fix some downstream tests. (It has some assumptions which make it fail when the output is tested.) If you want me to fix that let me know.

mattsse:
yeah, thanks again for this @lattejed !
I cherry-picked some parts for #802 

so we can close this

gakonst:
excellent - closing and let's move to #802

PR #811, opened 2022-01-19
try and fix thread 'main' panicked 

ChainsightAnalytics:
at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.synchronous tasks.', C:\Users\Ben\.cargo\registry\src\github.com-1ecc6299db9ec823\tokio-1.15.0\src\runtime\enter.rs:39:9

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/compile.rs
+++ b/ethers-solc/src/compile.rs
@@ -5,6 +5,7 @@ use crate::{
 };
 use semver::{Version, VersionReq};
 use serde::{de::DeserializeOwned, Serialize};
+
 use std::{
     fmt,
     fmt::Formatter,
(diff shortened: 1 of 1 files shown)


gakonst:
@ChainsightAnalytics this needs some fixes to compile, could you take another pass perhaps?

ChainsightAnalytics:
Updated, should compile now!

gakonst:
@ChainsightAnalytics one last change required to format properly!

PR #820, opened 2022-01-20
Remove entire cache folder, not just contents

onbjerg:
Should fix https://github.com/gakonst/foundry/issues/322 (?)


gakonst (approved):

mds1:
Hm, I'm not sure we want to remove the whole `cache` folder. For people migrating from Hardhat to Foundry, this will also delete the hardhat cache that lives in `cache/hardhat-network-fork`. Seems it'd be better UX to be more selective and only delete cache files/folders from foundry or ethers-rs

mattsse:
this also broke the `clean` function because `cache_path` points to the file, not the dir, so `std::fs::remove_dir_all(self.cache_path())` always fails.

gakonst:
ugh i rushed to merge this, let's revert

PR #822, opened 2022-01-20
chore(solc): also rm cache dir if empty

mattsse:
## Motivation
followup #820
Ref https://github.com/gakonst/foundry/issues/322#issuecomment-1017590516

## Solution
if the cache file was the only file in the dir, remove the parent dir (`./cache/`) as well

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
lgtm - note that if we intend changing the caching format we'll need to adjust this logic too

Issue #684, opened 2021-12-13
subscribe_blocks() is not working


stephenctw:
**Version**
0.6.2

**Platform**
WSL2 (Linux LAPTOP-163CP44A 4.19.104-microsoft-standard #1 SMP Wed Feb 19 06:37:35 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux)

**Description**
ethers::provider::subscribe_blocks() not responding, the wss endpoint is valid that subscribe_pending_txs() is working properly.

I tried this code:

```
use ethers::prelude::*;
use ethers_providers::Ws;

use std::time::Duration;

#[tokio::main]
async fn main() -> std::result::Result<(), Box<dyn std::error::Error>> {
    let ws_endpoint = "wss://goerli.infura.io/ws/v3/<PROJECT_ID>";

    let ws = Ws::connect(ws_endpoint).await?;

    println!("endpoint ready?: {}", ws.ready());

    let provider = Provider::new(ws).interval(Duration::from_millis(2000));

    let mut stream = provider.subscribe_blocks().await?.take(5);
    // let mut stream = provider.subscribe_pending_txs().await?.take(5);

    println!("Stream created! Entering loop...");
    while let Some(block) = stream.next().await {
        dbg!(block);
    }

    Ok(())
}
```

gakonst:
Given that we're not hitting this bug in the [test](https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/provider.rs#L1350-L1359) I wonder if this has to do with some network delay / bug in our websocket implementation?

stephenctw:
I can confirm `subscribe_blocks()` is working fine with local hardhat node. `subscribe_pending_txs()` works with goerli + infura.

gakonst:
This is quite weird. I tested with Alchemy and it's working, but I am not getting any results from Infura. (Same code as yours, just replaced Infura URL with Alchemy)

lerencao:
> I can confirm `subscribe_blocks()` is working fine with local hardhat node. `subscribe_pending_txs()` works with goerli + infura.

Is your serde_json  `arbitrary_precision` enabled ?
I met the same problem, and found out it related to serde_json.

stephenctw:
> > I can confirm `subscribe_blocks()` is working fine with local hardhat node. `subscribe_pending_txs()` works with goerli + infura.
> 
[...]

PR #827, opened 2022-01-24
fix(ws): propagate deserialization errors upstream + infura quickfix

erwanor:
## Motivation

Deserializing a poorly formatted `Incoming` message over websocket fails silently. This is why #684 was mysteriously failing when using a `wss` provider from Infura. Infura streams back messages with an all capitalized `JSONRPC` field which serde chokes over when deserializing into a `Notification`. 

## Solution

Propagate deserialization errors through `ClientError`.

I think it's fair to support Infura idiosyncrasies considering they're such a prominent provider. With that said, the stable solution is probably using something like `#[serde(case_insensitive)]` (see https://github.com/serde-rs/serde/pull/1902) although this should come with an overhead. In the meantime, this fixes the issue for Infura wss users. 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-providers/src/transports/common.rs
+++ b/ethers-providers/src/transports/common.rs
@@ -39,6 +39,7 @@ pub struct Request<'a, T> {
 #[derive(Serialize, Deserialize, Debug)]
 /// A JSON-RPC Notifcation
 pub struct Notification<R> {
+    #[serde(alias = "JSONRPC")]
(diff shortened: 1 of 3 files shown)


gakonst (approved):
this is quite crazy, good find, thank you!

PR #828, opened 2022-01-25
fix/etherscan verification

rkrasiuk:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-etherscan/src/contract.rs
+++ b/ethers-etherscan/src/contract.rs
@@ -262,11 +262,12 @@ impl Client {
 
 #[cfg(test)]
 mod tests {
-    use std::time::Duration;
+    use std::{path::PathBuf, time::Duration};
 
     use serial_test::serial;
 
     use ethers_core::types::Chain;
+    use ethers_solc::{MinimalCombinedArtifacts, Project, ProjectPathsConfig};
 
     use crate::{contract::VerifyContract, tests::run_at_least_duration, Client};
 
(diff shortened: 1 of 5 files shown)


ethers-etherscan/src/contract.rs
```diff
@@ -302,26 +303,32 @@ mod tests {
...
+            let contract = project.flatten(&root.join("UniswapExchange.sol")).expect("failed to flatten contract");
 
             let client = Client::new_from_env(Chain::Mainnet).unwrap();
 
             let contract =
-                VerifyContract::new(address, contract.to_string(), compiler_version.to_string())
+                VerifyContract::new(address, contract, compiler_version.to_string())
                     .constructor_arguments(Some(constructor_args))
                     .optimization(true)
                     .runs(200);
```
rkrasiuk, line 329: the parameters for this test will be changed out for the ones exposed by #809

mattsse:
what could be nice to have is some roundtrip test:
1. get contract source code for a specific address
2. get metadata (hopefully compiler version and constructor arguments are included)
3. call verify

gakonst:
@rkrasiuk FYI #809 got rolled in #802 so if you'd want to unblock yourself from it you could use that as the base for this PR. Otherwise, will ping again when 802 is merged and we can then move forward with it?

rkrasiuk:
@gakonst what do you think about the following - I add a roundtrip test and we merge as it is. then, in foundry, we require to provide compiler version and num of optimizations for now. then, after #802 is merged, we just improve on top of this PR

gakonst:
agreed - sgtm

rkrasiuk:
@gakonst perfect, will take care of this tmrw morning

rkrasiuk:
@mattsse so, the metadata is all there in the response. the problem with the roundtrip test is that etherscan would error out on trying to verify the contract that was already verified. i guess we can redeploy the contract each time on goerli for verification, but we'd have to set up some dummy wallet with goerli eth

mattsse:
oh I see, I don't think that's worth it.
looking good!

gakonst (approved):
Looks good! Let's maybe look into how we can expand flattening to support multiple contracts In the same file?

PR #829, opened 2022-01-25
Add authorization for http and websocket

th4s:
## Summary
This pull request adds basic and bearer authentication for http and non-wasm websockets transport.

## Motivation
Some people operate their node behind a reverse-proxy which uses some form of authentication. This pull request makes it possible to use ethers-rs  in this case.

## Solution

The solution is to inject auth headers into http requests and the websocket handshake request.

## PR Checklist

- [ ] Added Tests (Is that necessary in this, case? If yes, how would you do that?
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-providers/src/transports/common.rs
+++ b/ethers-providers/src/transports/common.rs
@@ -2,7 +2,7 @@
 use ethers_core::types::U256;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
-use std::{fmt, string::ToString};
+use std::fmt;
 use thiserror::Error;
 
 #[derive(Serialize, Deserialize, Debug, Clone, Error)]
@@ -99,17 +99,25 @@ impl ResponseData<serde_json::Value> {
 /// Use to inject username and password or an auth token into requests
 #[derive(Clone, Debug)]
 pub enum Authorization {
-    Basic(String, String),
+    Basic(String),
     Bearer(String),
 }
 
 impl Authorization {
-    pub fn basic(username: impl ToString, password: impl ToString) -> Self {
-        Self::Basic(username.to_string(), password.to_string())
+    pub fn basic(username: impl Into<String>, password: impl Into<String>) -> Self {
+        let auth_secret = base64::encode(username.into() + ":" + &password.into());
+        Self::Basic(auth_secret)
     }
 
-    pub fn bearer(token: impl ToString) -> Self {
-        Self::Bearer(token.to_string())
+    pub fn bearer(token: impl Into<String>) -> Self {
+        Self::Bearer(token.into())
+    }
+
+    pub(crate) fn into_auth_string(self) -> String {
+        match self {
+            Authorization::Basic(auth_secret) => format!("Basic {}", auth_secret),
+            Authorization::Bearer(token) => format!("Bearer {}", token),
+        }
     }
 }
 
--- a/ethers-providers/src/transports/common.rs
+++ b/ethers-providers/src/transports/common.rs
@@ -112,11 +112,13 @@ impl Authorization {
     pub fn bearer(token: impl Into<String>) -> Self {
         Self::Bearer(token.into())
     }
+}
 
-    pub(crate) fn into_auth_string(self) -> String {
+impl fmt::Display for Authorization {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Authorization::Basic(auth_secret) => format!("Basic {}", auth_secret),
-            Authorization::Bearer(token) => format!("Bearer {}", token),
+            Authorization::Basic(auth_secret) => write!(f, "Basic {}", auth_secret),
+            Authorization::Bearer(token) => write!(f, "Bearer {}", token),
         }
     }
 }
--- a/ethers-providers/src/transports/common.rs
+++ b/ethers-providers/src/transports/common.rs
@@ -2,7 +2,7 @@
 use ethers_core::types::U256;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
-use std::fmt;
+use std::{fmt, string::ToString};
 use thiserror::Error;
 
 #[derive(Serialize, Deserialize, Debug, Clone, Error)]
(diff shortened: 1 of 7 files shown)


ethers-providers/src/transports/http.rs
```diff
@@ -69,10 +70,19 @@ impl JsonRpcClient for Provider {
...
-        let res = self.client.post(self.url.as_ref()).json(&payload).send().await?;
+        let mut client = self.client.post(self.url.as_ref()).json(&payload);
+        if let Some(auth) = &self.authorization {
+            client = match auth {
+                Authorization::Basic(username, password) => {
+                    client.basic_auth(username, Some(password))
+                }
+                Authorization::Bearer(token) => client.bearer_auth(token),
+            };
+        }
```
mattsse, line 83: instead of doing this, we should rather set the default headers for the client
https://docs.rs/reqwest/latest/reqwest/blocking/struct.ClientBuilder.html#method.default_headers

ethers-providers/src/transports/http.rs
```diff
@@ -94,7 +104,24 @@ impl Provider {
...
+    /// use ethers_providers::{Authorization, Http};
+    /// use url::Url;
+    ///
+    /// let url = Url::parse("http://localhost:8545").unwrap();
+    /// let provider = Http::new_with_auth(url, Authorization::Basic("admin".into(), "good_password".into()));
+    /// ```
+    pub fn new_with_auth(url: impl Into<Url>, auth: Authorization) -> Self {
+        let mut provider = Self::new(url);
+        provider.authorization = Some(auth);
+        provider
```
mattsse, line 124: same point,
instead of rolling our own `Auth` types, let's add an additional constructor that accepts a `reqwest::Client`, this way a user can configure any default headers

ethers-providers/src/transports/ws.rs
```diff
@@ -140,6 +144,26 @@ impl Ws {
         Ok(Self::new(ws))
     }
 
+    /// Initializes a new WebSocket Client with authentication
+    #[cfg(not(target_arch = "wasm32"))]
+    pub async fn connect_with_auth(
```
mattsse, line 149: this makes more sense here as adding auth is a bit more verbose

ethers-providers/src/transports/ws.rs
```diff
@@ -140,6 +144,26 @@ impl Ws {
...
+        uri: impl AsRef<str> + Unpin,
+        auth: Authorization,
+    ) -> Result<Self, ClientError> {
+        let mut request: HttpRequest<()> =
+            HttpRequest::builder().method("GET").uri(Uri::from_str(uri.as_ref())?).body(())?;
+        let auth_header = match auth {
+            Authorization::Basic(username, password) => {
+                String::from("Basic ") + base64::encode(username + ":" + &password).as_str()
```
mattsse, line 157: use format style string
`format!("Basic {}",base64::encode(username + ":" + &password).as_str())`

ethers-providers/src/transports/ws.rs
```diff
@@ -140,6 +144,26 @@ impl Ws {
...
+            }
+            Authorization::Bearer(token) => String::from("Bearer ") + &token,
```
mattsse, line 159:
```suggestion
            Authorization::Bearer(token) => format!("Bearer {}", &token),
```

ethers-providers/src/transports/http.rs
mattsse: however if we use the `Authorization` type for the ws auth, then this is fine.

still would like to see an additional function that accepts a `reqwest::Client`

ethers-providers/src/transports/common.rs
```diff
@@ -94,6 +94,15 @@ impl ResponseData<serde_json::Value> {
     }
 }
 
+/// Basic or bearer authentication in http or websocket transport
+///
+/// Use to inject username and password or an auth token into requests
+#[derive(Clone, Debug)]
+pub enum Authorization {
+    Basic(String, String),
```
mattsse, line 102: can we add constructors for both variants so that a user knows that `Basic` is name, password?

mattsse (changes requested):
thanks for this,
some suggestions and few nits

ethers-providers/src/transports/common.rs
```diff
@@ -103,6 +103,16 @@ pub enum Authorization {
     Bearer(String),
 }
 
+impl Authorization {
+    pub fn basic(username: impl ToString, password: impl ToString) -> Self {
```
mattsse, line 107: Use `impl Into<String>` because calling `String::to_strig` effectively clones, whereas `into` does not
```suggestion
    pub fn basic(username: impl Into<String>, password: impl Into<String>) -> Self {
```

ethers-providers/src/transports/common.rs
```diff
@@ -103,6 +103,16 @@ pub enum Authorization {
...
+        Self::Basic(username.to_string(), password.to_string())
+    }
+
+    pub fn bearer(token: impl ToString) -> Self {
```
mattsse, line 111:
```suggestion
    pub fn bearer(token: impl Into<String>) -> Self {
```

mattsse (changes requested):

th4s:
Thanks for your feedback!

ethers-providers/src/transports/ws.rs
th4s: Is this point still valid? If yes, could you please be a little bit more specific?

ethers-providers/src/transports/common.rs
```diff
@@ -94,6 +94,33 @@ impl ResponseData<serde_json::Value> {
...
+    pub fn bearer(token: impl Into<String>) -> Self {
+        Self::Bearer(token.into())
+    }
+
+    pub(crate) fn into_auth_string(self) -> String {
+        match self {
+            Authorization::Basic(auth_secret) => format!("Basic {}", auth_secret),
+            Authorization::Bearer(token) => format!("Bearer {}", token),
+        }
+    }
```
mattsse, line 121: smol nit
this can be moved to a `impl Display for Authorization`

mattsse (approved):
last nit but lgtm!

ethers-providers/src/transports/common.rs
th4s: I didn't do this on purpose, since I thought this is more of a transformation than just a way of displaying it. But I haven't a strong opinion on that. Happy to change it.

th4s:
Thanks a lot!

mattsse (approved):

gakonst (approved):
very nic

PR #831, opened 2022-01-26
feat(chain): add BSC and BSC testnet

optionseskimo:
## Motivation
As noted in gakonst/foundry#579, BSC mainnet and testnet are missing from the list of networks that do not support EIP-1559 yet. Thus, sending transactions to those networks using [`cast`](https://github.com/gakonst/foundry/tree/master/cast) fails because currently [`Eip1559TransactionRequest` is used to create and submit transactions.](https://github.com/gakonst/foundry/blob/9de25d1928da362cd3b185ebaccc0e3d00aee2c8/cast/src/lib.rs#L234)

## Solution

Add BinanceSmartChain and BinanceSmartChainTestnet to the `Chain` enum in ethers-core types.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Unreleased
 
+- Add BSC mainnet and testnet to the list of known chains
+  [831](https://github.com/gakonst/ethers-rs/pull/831)
 - Returns error on invalid type conversion instead of panicking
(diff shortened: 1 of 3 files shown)


mattsse (approved):
thanks!
lgtm

PR #833, opened 2022-01-26
write `CompactContractBytecode` instead of `CompactContract`

brockelmore:
## Motivation
The existing `CompactContract` doesn't have enough information for our needs in foundry. So we write out `CompactContractBytecode` instead which is a slightly stripped down version of `ContractBytecode`. The reason to strip it down is because we lose a decent amount of performance if we write all the other things in `ContractBytecode`. This is a compromise between speed and the needed information.

## Solution

Write `CompactContractBytecode` instead of `CompactContract`

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts.rs
+++ b/ethers-solc/src/artifacts.rs
@@ -814,6 +814,21 @@ impl ContractBytecode {
             deployed_bytecode: self.deployed_bytecode.unwrap(),
         }
     }
+
+    /// Looks for all link references in deployment and runtime bytecodes
+    pub fn all_link_references(&self) -> BTreeMap<String, BTreeMap<String, Vec<Offsets>>> {
+        let mut links = BTreeMap::new();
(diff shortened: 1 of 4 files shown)


mattsse (commented):
necessary workaround until we find a more flexible solution.

still fails to compile due to changed types.

the `Contracts..` representation is a bit of a mess...

brockelmore:
> still fails to compile due to changed types.

wdym? compiles fine for me?

mattsse:
looks like a doc-test https://github.com/gakonst/ethers-rs/runs/4952579664?check_suite_focus=true#step:9:539

brockelmore:
oh ye the *test* fails, but it compiles okay

gakonst (approved):
LGTM - we should do a pass at the whole crate to see if we got duplicate functionality / if we can clean things up

PR #835, opened 2022-01-27
feat(types/chain): is_legacy helper

tarrencev:
moving this into ethers-rs from foundry so it can be more easily use elsewhere


gakonst:
Idea LGTM some syntax error remains

gakonst:
Needs fmt!

tarrencev:
@gakonst i think we're good now? fmt seems to be failing on other code. do we want to include that fix in here?

PR #838, opened 2022-01-28
chore: add info messages to solc install/compile

odyslam:
Add informational messages to solc installation and compilation. 

The compilation of a project could give the appearance of a hanging program because of downloading and installing solc. Informational messages allow the user to know what's happening and why it might take more than expected.

### DIFF
--- a/ethers-solc/src/compile.rs
+++ b/ethers-solc/src/compile.rs
@@ -192,7 +192,7 @@ pub struct Solc {
 impl Default for Solc {
     fn default() -> Self {
         if let Ok(solc) = std::env::var("SOLC_PATH") {
-            return Solc::new(solc);
+            return Solc::new(solc)
         }
         #[cfg(not(target_arch = "wasm32"))]
         {
@@ -200,7 +200,7 @@ impl Default for Solc {
                 .and_then(|vers| Solc::find_svm_installed_version(&vers.to_string()).ok())
                 .flatten()
             {
-                return solc;
+                return solc
             }
         }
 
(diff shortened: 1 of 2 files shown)


ethers-solc/src/compile.rs
```diff
@@ -493,7 +497,7 @@ impl Solc {
             .spawn()
             .map_err(|err| SolcError::io(err, &self.solc))?;
         let stdin = child.stdin.take().unwrap();
-
+        println!("project compilation completed");
```
gakonst, line 500:
```suggestion
```

gakonst (approved):
LGTM for the download step, no need to add extra info elsewhere imo, as one may want to have more control over that message

mattsse (changes requested):
I'm leaning against this for the simple reason that this is a foundry use case and ethers-solc is a standalone crate that ideally should not print anything unconditionally.

A better solution imo would be to have something like

```rust
pub trait SolcProgressMonitor {
   fn on_install(version: &Version) {}
   fn on_installed(version: &Version) {}
   fn on_compile(input: &CompilerInput) {}
}
```

we could then do something similar to how logging works in general
basically using a global `SolcProgressMonitor` that calls the callbacks.
This global instance (static mut) can then be instantiated with a `StdOutProgressMinitor` implementation for example.

this way the user can control, if and how things are logged.

odyslam:
@mattsse -- The reason I made the contributions here is that all these steps are exposed via a single function call in the forge build, as far as i understood it. 

I agree with your approach.  Let me riff on that as it is on the edge of my rust understanding.

gakonst:
@mattsse agree on the hook for logging to stdout, good idea.

PR #839, opened 2022-01-29
core: fix overflow and panic in priority fee estimation

juniorbeef:
The U256 priority fees were being coerced into u32, which was not big
enough for actual values. The overflow was happening but not checked.
This led to a possible divide-by-zero panic in this code as well.

This change does the math as I256 -- overkill, but it works.

## Motivation

This is a simple bug fix for the issue described above, which could lead in some 
cases to somewhat incorrect gas estimation and the occasional panic.

## Solution

The fix just uses I256 math instead.

## PR Checklist

- [ x] Added Tests
- [ x] Added Documentation
- [ x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Unreleased
 
+- Fix overflow and possible divide-by-zero in `estimate_priority_fee`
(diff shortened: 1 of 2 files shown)


gakonst (approved):
Good catch

PR #844, opened 2022-01-29
[WIP] Forge middleware

wolflo:
## Motivation

**Goal:** Write rust tests for smart contracts that leverage Foundry's performance, debugging, and usability. Ideally, minimal effort is required to switch your rust tests between using an rpc endpoint and using a Foundry `evm_adapter`.

See https://github.com/gakonst/foundry/issues/22.

## Solution

An ethers middleware that is instantiated with an implementer of Foundry's [`evm_adapters::Evm`](https://github.com/gakonst/foundry/blob/0773fb41582d7ad02269902b0097e5db58198831/evm-adapters/src/lib.rs#L57) trait. Instead of sending transactions over rpc, this middleware interfaces directly with the evm instance via `Evm::call_raw()`.

## Challenges

It's not clear that this is really the best way to do things. Currently, this middleware needs to make every effort to avoid delegating anything to an inner middleware, because there is a good chance that inner middleware simply does not exist. It also will need to re-implement many of the methods that the `Provider` struct currently implements, because every sub-call needs to go through this middleware, and `Provider` is currently enshrined as the innermost middleware. 

The sputnik [`Backend`](https://github.com/rust-blockchain/evm/blob/78c49debfc65ddf3c0635edd02045a937dcc15d5/src/backend/mod.rs#L52-L86) trait gives a good idea of what kind of data we have available from an `evm_adapter` instance. In short, it doesn't look like we can get anything block or node related, including most historical data.

There are situations where a meaningful implementation of an inner middleware might exist, though. For instance, if you instantiate this middleware with an `evm_adapter` using forked state from mainnet, you could also provide an inner rpc client that can query historical data from a mainnet rpc endpoint. Or, there may ultimately be an rpc endpoint talking to the same `evm_adapter` instance: https://github.com/gakonst/foundry/pull/174.

## Alternative Solutions

1. Have the middleware live at the level of a `JsonRpcClient` instead, intercepting requests to rpc methods and dispatching to either the evm instance or to another optional inner provider. This seems like a more natural place for this to live, and it would avoid needing to reimplement many of the methods already defined on `Provider`. But, it would require deserializing requests that were just serialized in the provider, then doing the same round trip with the response from the evm instance. This poses both performance and correctness challenges.

2. Split the work between a `Middleware` and a `JsonRpcClient` impl, allowing the `JsonRpcClient` impl to catch requests from the `Provider` middleware impl and pass them back up to the Forge middleware. The same concerns above apply for this approach in addition to the potential to end up stuck in a  `middleware -> provider -> client -> middleware -> ...` loop. But, this avoids needing to redefine all of the Provider methods on our middleware, and it would allow us to use the inner provider more effectively.

3. Put the middleware right where `Provider` lives now. Currently, the `Provider<P: JsonRpcClient>` struct is enshrined as the "base" of the middleware. If you were able to replace the Provider with any "innerware", that would be a logical place for the Forge middleware to live.

## TLDR
The interesting pieces here are probably the implementation of the `Middleware` trait's `call()` and `send_transaction()` methods, and the associated `apply_tx()` method.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,6 +35,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.4.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "739f4a8db6605981345c5654f3a85b056ce52f37a39d34da03f25bf2151ea16e"
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


wolflo:
After some discussion, this would be cleaner as a separate repo. Moving the implementation to https://github.com/wolflo/forge-rs.

I will update this PR to a feature-gated re-export, but any input on the approach is appreciated, either here or in the original issue: https://github.com/gakonst/foundry/issues/22.

wolflo:
I'm going to close this for now. Please feel free to discuss / critique on the original foundry issue (https://github.com/gakonst/foundry/issues/22) or open an issue on the current repo (https://github.com/wolflo/forge-middleware).

PR #847, opened 2022-01-31
feat/ethers-solc: write outputs (asm, ir...) to file, if selected

joshieDo:
## Motivation
Ref PR https://github.com/gakonst/foundry/pull/644
Ref Issue https://github.com/gakonst/foundry/issues/611

## Solution
They should be written to file, only if selected on settings. But it still doesn't feel good to be here... Suggestions?
 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts.rs
+++ b/ethers-solc/src/artifacts.rs
@@ -752,6 +752,7 @@ impl<'a> fmt::Display for OutputDiagnostics<'a> {
 
 /// Represents a compiled solidity contract
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
 pub struct Contract {
     /// The Ethereum Contract Metadata.
     /// See https://docs.soliditylang.org/en/develop/metadata.html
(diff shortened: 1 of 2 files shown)


mattsse (changes requested):
this is a bit unsound because we have a generic `Artifact` output type which could be in conflict with the additional files.

one thing we'd need to decide is whether we want to bundle this in a single artifact file or separate files like it's done here.

imo it is cleaner to have them separate, but this way they'll always be written if the value is present which might not be expected.

Not quite sure, what would be the best way of doing this.

we could solve this by
* adding an additional config type that specifies what should be written in addition to the core artifact. Basically turning the `extrac_output` into a separate type and add this to the `Project's` settings?

joshieDo:
> one thing we'd need to decide is whether we want to bundle this in a single artifact file or separate files like it's done here.

I think it has to be a mix. Certain artifacts , `ir`, `evm.assembly` or even `abi` are basically used as standalone files. It doesn't make sense imo to put them all in one json file.

> imo it is cleaner to have them separate, but this way they'll always be written if the value is present which might not be expected.
correct me if i'm wrong, but they'll only be present if added to the compiler settings no?

mattsse:
> > one thing we'd need to decide is whether we want to bundle this in a single artifact file or separate files like it's done here.
> 
> I think it has to be a mix. Certain artifacts , `ir`, `evm.assembly` or even `abi` are basically used as standalone files. It doesn't make sense imo to put them all in one json file.
> 
> > imo it is cleaner to have them separate, but this way they'll always be written if the value is present which might not be expected.

Agree

> 
> correct me if i'm wrong, but they'll only be present if added to the compiler settings no?

makes sense, +1 on merging this, there's no harm in emitting those files anyway.

gakonst:
Merging, and we can re-consider as we iterate on this idea.

PR #848, opened 2022-01-31
add eth_syncing RPC

rakita:
Add `eth_syncing` endpoint: https://eth.wiki/json-rpc/API#eth_syncing

Tested it with infure and return `false` successfully, don't have any syncing node to check return when it is syncing.

### DIFF
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -27,7 +27,7 @@ pub use pubsub::{PubsubClient, SubscriptionStream};
 use async_trait::async_trait;
 use auto_impl::auto_impl;
 use ethers_core::types::transaction::{eip2718::TypedTransaction, eip2930::AccessListWithGasUsed};
-use serde::{de::DeserializeOwned, Serialize};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::{error::Error, fmt::Debug, future::Future, pin::Pin};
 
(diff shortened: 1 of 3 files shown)


ethers-providers/src/lib.rs
```diff
@@ -79,6 +79,14 @@ where
     }
 }
 
+/// Structure used in eth_syncing RPC
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(untagged)]
+pub enum SyncingStatus {
+    IsFalse(bool),
```
gakonst, line 86: could you add a note on how this works? not clear to me what `IsFalse(false)` vs `IsFalse(true)` means?

gakonst (approved):

ethers-providers/src/lib.rs
rakita: I added comments, but `IsFalse(true)` is not possible to happen by eth spec. wdyt about adding an additional struct, first for deserialization and second for returning to library call?

gakonst: oh as in the commit you pushed does not work? yeah I'm OK with adding an intermediate non-public datatype that helps with the desert and then having the "simple" enum.

rakita: Yea i made mistake with that commit, now it should work as expected. Can you check it again if it is okay, I embedded intermediate type inside function.

gakonst (approved):
Gg

Issue #850, opened 2022-01-31
New ganache output causes hang in `Ganache::new().spawn()`


wolflo:
**Version**
Current master branch (commit `5fefb2b`).

**Description**

`ganache-cli` version `7.0.1` changed what is printed to stdout slightly, and ethers currently relies on this output to determine whether `Ganache::new().spawn()` command has succeeded. This causes `Ganache::new()::spawn()` to hang.

Specifically, ethers looks for "Listening on..." and v7.0.1 prints "RPC Listening on...".

Note that `v7.0.1` can only be installed with `npm install -g ganache` (not `ganache-cli`).

### `ganache-cli` output, `v6.12.2`
```
Ganache CLI v6.12.2 (ganache-core: 2.13.2)

Available Accounts
==================
(0) 0x476167068F9B5A2664e0371a787E1ed4b5D92F15 (100 ETH)
(1) 0x02E71018660C92C916BddF15e114764E10Cef7b9 (100 ETH)
(2) 0x501e5084E929B34cEB88d3416ec4Fa91362C05D2 (100 ETH)
(3) 0x90F45668A993BA74aA7084b569857fC6B40014f9 (100 ETH)
(4) 0x058115A3A4E05Ee39e7d8c4B111E61BE479b3A74 (100 ETH)
(5) 0x57719368fFeda041e7b136754cf6461851E4ac0A (100 ETH)
(6) 0x51AFc78108099A271A836DFdFDd290496F8D3912 (100 ETH)
(7) 0x5DBb4AEC31016c015DD2aF269aE688e2B33B6290 (100 ETH)
(8) 0xA61EbB8fF0F2db4B6A3d224d8FC51D575B3A9860 (100 ETH)
(9) 0xe4786b0C8328D6B6194AD8f05FC6ae70A58Be1BB (100 ETH)

Private Keys
==================
(0) 0x284d673f0824fc62fc0519afd21578702099edc1a0c24d28ba10ad8f7029193c
(1) 0x05c0bd5cc51545c72f9b64b9ca6a13a8b4fed15e17dc56c6f84432a0136cf590
(2) 0x2960390cf06e17d525ee0b50f40bba80a503e55c216866e984d8aab076cd2a0d
(3) 0xc796caa1dcd7b80e9bfd49311c7551e3f486f1c9f239c5d13cca1e9d82f5a6e3
(4) 0x56884f6d71f9c669aa733012c6db031b9e25b16f3a55d6f18cf1743d5661cf1c
[...]

PR #851, opened 2022-01-31
fix(core): adjust Ganache for new cli output

wolflo:
## Motivation

Addresses #850. In general, there is probably a less fragile way to launch ganache and decipher the output. Open to suggestions on a possible change to the approach here.

## PR Checklist

- [ ] Added Tests - X - Tough to test in CI, given its reliance on locally installed `ganache-cli` version.
- [ ] Added Documentation - X
- [ ] Updated the changelog - X


mattsse (approved):
lgtm

gakonst (approved):

PR #852, opened 2022-02-01
refactor: MultiAbigen rework

prestwich:
Draft status until I update docs and changelog

## Motivation

- allow distinguishing between binding generation and serialization/write failures by introducing an explicit `build` step
- allow lib generation
- allow building bindings in memory to prevent unnecessary disk usage on `ensure_consistent` 

## Solution

- refactor of `MultiAbigen`.
- `MultiAbigen` is now a thin vec wrapper that must be explicitly built into a `MultiBindings` (rather than implicit while writing)
- `MultiBindings` holds a map of `contract name -> ContractBinding`
- `MultiBindings` generates file contents in memory, and makes minimal disk writes
- `MultiBindings` no longer writes a tmp dir when running its checks
- `Multibindings::ensure_consistent_crate` and `::ensure_consistent_module` are now available outside of `#[cfg(test)]` as they no longer use a tmp dir

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/lib.rs
+++ b/ethers-contract/ethers-contract-abigen/src/lib.rs
@@ -219,7 +219,7 @@ impl ContractBindings {
         self.tokens
     }
 
-    /// Generate the defaul module name (snake case of the contract name)
+    /// Generate the default module name (snake case of the contract name)
     pub fn module_name(&self) -> String {
         self.name.to_snake_case()
     }
--- a/ethers-contract/ethers-contract-abigen/src/lib.rs
+++ b/ethers-contract/ethers-contract-abigen/src/lib.rs
@@ -84,6 +84,19 @@ impl Abigen {
         })
     }
 
+    /// Attemtps to load a new builder from an ABI JSON file at the specific
+    /// path.
+    pub fn from_file(path: impl AsRef<Path>) -> Result<Self> {
+        let name = path
+            .as_ref()
+            .file_stem()
+            .ok_or_else(|| anyhow::format_err!("Missing file stem in path"))?
+            .to_str()
+            .ok_or_else(|| anyhow::format_err!("Unable to convert file stem to string"))?;
+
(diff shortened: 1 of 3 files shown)


ethers-contract/ethers-contract-abigen/src/multi.rs
```diff
@@ -0,0 +1,752 @@
...
+    ///     ├── er20.rs
+    ///     ├── contract1.rs
+    ///     ├── contract2.rs
+    /// ...
+    /// ```
+    ///
+    /// ```no_run
+    /// # use ethers_contract_abigen::MultiAbigen;
+    /// let gen = MultiAbigen::from_json_files("./abi").unwrap();
+    /// gen.write_to_crate("./bindings").unwrap();
```
mattsse, line 333: example doesn't match anymore

mattsse (approved):
lgtm,
think this covers all use cases

prestwich:
Just added some additional convenience:

- `impl FromIterator for MultiAbigen`
- `Abigen::from_file`

ethers-contract/ethers-contract-abigen/src/lib.rs
```diff
@@ -181,289 +202,32 @@ impl ContractBindings {
...
-            let existing_contract_bindings = fs::read_to_string(existing_bindings_file).unwrap();
-
-            let fresh_bindings = fs::read_to_string(fresh_file.path()).unwrap();
-
-            if existing_contract_bindings != fresh_bindings {
-                return false
-            }
-        }
-        true
+    /// Generate the defaul module name (snake case of the contract name)
```
gakonst, line 222:
```suggestion
    /// Generate the default module name (snake case of the contract name)
```

gakonst (approved):

gakonst:
Amazing work

PR #857, opened 2022-02-02
fix(providers): doc test for dev_rpc

wolflo:
## Motivation

The doc test for the `dev_rpc` mod was failing when run directly from `ethers_providers`, because tokio's `rt-multi-thread` feature is not enabled.

## Solution

Specify the `current_thread` runtime flavor instead.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse (approved):
TIL

PR #858, opened 2022-02-02
refactor: replace anyhow with eyre

prestwich:
## Motivation

As foundry is using eyre, we should transition more crates to use it instead of anyhow

## Solution

Move the contract abigen crate to eyre

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/contract.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract.rs
@@ -199,7 +199,7 @@ impl Context {
             };
 
             if method_aliases.insert(signature.clone(), alias).is_some() {
-                return Err(eyre!("duplicate method signature '{}' in method aliases", signature,))
+                eyre::bail!("duplicate method signature '{}' in method aliases", signature)
             }
         }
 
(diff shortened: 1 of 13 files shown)


ethers-contract/ethers-contract-abigen/src/contract.rs
```diff
@@ -199,7 +199,7 @@ impl Context {
             };
 
             if method_aliases.insert(signature.clone(), alias).is_some() {
-                return Err(anyhow!("duplicate method signature '{}' in method aliases", signature,))
+                return Err(eyre!("duplicate method signature '{}' in method aliases", signature,))
```
mattsse, line 202:
```suggestion
              eyre::bail!("duplicate method signature '{}' in method aliases", signature)
```

mattsse (approved):
lgtm!

ethers-contract/ethers-contract-abigen/src/contract.rs
prestwich: fixed several places, thanks

prestwich:
cool, I just reviewed all eyre uses in the lib and made them more idiomatic with `bail!` and `ensure!` whereever possible

gakonst (approved):
gg

PR #863, opened 2022-02-04
Hide Infura API key from docs

MistApproach:
Unless it's intentional API key should be kept private?


mattsse (approved):
thank you!

MistApproach:
You are welcome :)
What's your take on API keys inside tests? Do you want to keep them?

PR #866, opened 2022-02-05
feat(solc): compiler pipeline improvements

mattsse:
## Motivation

improve how versions are autodetected. The old algorithm simply picked the highest compatible version, without considering other source files. So that `A (=0.8.10)` and `B(>0.8.0)` would result in two versions (0.8.10 and 0.8.11). Now we're trying to find a single versions that satisfies all sets, see resolver.rs docs.

limit the `CompilerOutput`'s contracts: solc may include additional contracts in the `CompilerOutput` than were provided in the `CompilerInput` which is not desired in the `ProjectCompiler` since that already provides all contracts, so we simply restrict the `CompilerOutput` to those files provided in the `CompilerInput` `Solc::compile_exact`

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/compile/project.rs
+++ b/ethers-solc/src/compile/project.rs
@@ -409,10 +409,8 @@ fn compile_parallel(
 #[cfg(feature = "project-util")]
 mod tests {
     use super::*;
-    use crate::{
-        artifacts::FileToContractsMap, project_util::TempProject, MinimalCombinedArtifacts,
-    };
-    use semver::Version;
+    use crate::{project_util::TempProject, MinimalCombinedArtifacts};
+
     use std::path::PathBuf;
 
     #[allow(unused)]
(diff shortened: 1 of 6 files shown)


ethers-solc/src/compile/project.rs
```diff
@@ -384,6 +397,8 @@ fn compile_parallel(
             .collect::<Result<Vec<_>>>()
     })?;
 
+    // TODO need to do post filtering as the output can contain more files than provided in the
```
gakonst, line 400: Wdym post filtering?

ethers-solc/src/resolver.rs
```diff
@@ -14,6 +14,26 @@
 //! Finding all dependencies is fairly simple, we're simply doing a DFS, starting the source
 //! contracts
 //!
+//! ## Solc version auto-detection
```
gakonst, line 17: Excellent doc

gakonst (approved):

ethers-solc/src/compile/project.rs
mattsse: oh this should have been removed...no longer relevant

PR #869, opened 2022-02-05
feat(chain): add arbitrum support

0xTomoyo:
## Motivation

Add Arbitrum mainnet and testnet to the list of known chains/block explorers

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog


mattsse (approved):
thanks,

afaik there is no eip1559 support on arbitrum yet, right?
so we should add that them here as well

https://github.com/gakonst/ethers-rs/blob/59d182195cc8c9bf6b23915115ed8b3b6da76e31/ethers-core/src/types/chain.rs#L129

gakonst (approved):
Needs lint plus Matt's change but otherwise lgtm

PR #878, opened 2022-02-07
fix(solc): use nodesiter when printing tree

mattsse:
## Motivation
prevent possible SO when construction import tree

## Solution
use existing `NodesIter` to determine error msg like

```console
   0: Discovered incompatible solidity versions in following
      : src/MyContract.sol (^0.8.0) imports:
          lib/yearn-vaults/contracts/BaseStrategy.sol (>=0.6.0 <0.7.0) 
          lib/yearn-vaults/contracts/BaseStrategy.sol (>=0.6.0 <0.7.0)
``
<!--
Summarize the solution and provide any necessary context needed to understand
the code change.
-->

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### CODE REVIEW

File: ethers-solc/src/resolver.rs [hunk 1]
```diff
@@ -661,6 +654,18 @@ struct SolData {
     version_req: Option<VersionReq>,
 }
 
+impl SolData {
+    fn fmt_version<W: std::fmt::Write>(
```
gakonst, line 658: Any idea why this throws warning?

gakonst (approved):

ethers-solc/src/resolver.rs
mattsse: guess unused in wasm

PR #805, opened 2022-01-18
Implement RLP decoding for transactions

Rjected:
## Motivation

Currently there is no way to decode a transaction from bytes, only encode to bytes.

## Solution

Implemented the [`Decodable`](https://docs.rs/rlp/latest/rlp/trait.Decodable.html) trait for:
 - `TypedTransaction`
   - `TransactionRequest`
   - `Eip2930TransactionRequest`
   - `Eip1559TransactionRequest`
 - `NameOrAddress`
 - `Transaction`

This change uses `rlp_derive` for the `Decodable` implementation on `AccessListItem` and `AccessList`.

This also adjusts each variant of `TypedTransaction` to contain the `chain_id`, and removes `chain_id` it from the `rlp_signed` and `sighash` functions.

Fixes #561 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

PR #879, opened 2022-02-07
fix(core): prevent chain_id from serializing for requests

### LINKED ISSUE #877: unknown field chainId when trying to deploy contract with forge
when deploying a contract to kovan with forge create getting the following error

(code: -32602, message: Invalid params: unknown field chainId, expected one of type, from, to, gasPrice, maxFeePerGas, gas, value, data, nonce, accessList, maxPriorityFeePerGas., data: None)

Version: forge 0.1.0 (8107afe 2022-01-31T10:48:23.593667+00:00)

Rjected:
## Motivation
OpenEthereum seems to not have a `chainId` field in its [`TransactionRequest`](https://github.com/openethereum/openethereum/blob/main/crates/rpc/src/v1/types/transaction_request.rs#L32) type.

## Solution
This PR disables serialization for the `chain_id`, so we don't send it in RPC requests.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

Fixes #877

### DIFF
--- a/ethers-core/src/types/transaction/request.rs
+++ b/ethers-core/src/types/transaction/request.rs
@@ -43,7 +43,7 @@ pub struct TransactionRequest {
 
     /// Chain ID (None for mainnet)
     #[serde(skip_serializing)]
-    #[serde(rename = "chainId")]
+    #[serde(default, rename = "chainId")]
     pub chain_id: Option<U64>,
 
     /////////////////  Celo-specific transaction fields /////////////////
--- a/ethers-core/src/types/transaction/request.rs
+++ b/ethers-core/src/types/transaction/request.rs
@@ -42,7 +42,8 @@ pub struct TransactionRequest {
     pub nonce: Option<U256>,
 
     /// Chain ID (None for mainnet)
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing)]
+    #[serde(rename = "chainId")]
     pub chain_id: Option<U64>,
 
     /////////////////  Celo-specific transaction fields /////////////////
--- a/ethers-core/src/types/transaction/eip1559.rs
+++ b/ethers-core/src/types/transaction/eip1559.rs
@@ -58,7 +58,8 @@ pub struct Eip1559TransactionRequest {
     /// baseFeePerGas + maxPriorityFeePerGas is “refunded” to the user.
     pub max_fee_per_gas: Option<U256>,
 
-    #[serde(rename = "chainId", default, skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing)]
+    #[serde(rename = "chainId")]
     /// Chain ID (None for mainnet)
     pub chain_id: Option<U64>,
 }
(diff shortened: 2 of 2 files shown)


mattsse (commented):
hmm, this is a bit tricky, because I'm not sure what side-effects this could potentially cause, what does? since this is a client issue for openethereum
I believe in geth chainid is omitempty, so this approach seems fine. But not sure if there is some kind of check that chainids match so that tx intended for chain X are rejected on chain Y? so we'd lose that

alternatives:

* should we skip this instead based on the chainid?
* should we rather take the chain id if the client is OE, I believe we already have some calls that check the client first, definitely more effort, not ideal.

Rjected:
Since `chain_id` was [recently](https://github.com/gakonst/ethers-rs/pull/805) introduced to our request types, we shouldn't really be introducing any new issues by not serializing.

I'm not sure that alternative 1 (skip based on the chainid), or any solution that depends on _only_ the chainid would work since OE will not accept any chainId field. [Geth does have omitempty for the chainId](https://github.com/ethereum/go-ethereum/blob/2d20fed893faa894f50af709349b13b6ad9b45db/internal/ethapi/transaction_args.go#L36), but it would be up to us to ensure that the `chain_id` field matches that of the client when running `send_transaction`, `sign_transaction`, etc. This would probably prevent user error.

This would not be necessary for `send_raw_transaction`, since that depends on RLP, which should already include the chainid. [Geth does check the chainId](https://github.com/ethereum/go-ethereum/blob/2d20fed893faa894f50af709349b13b6ad9b45db/core/types/transaction_signing.go#L272), so if we were to do chainId checks there it would be in case other clients don't do these checks.

ethers-core/src/types/transaction/request.rs
```diff
@@ -42,7 +42,8 @@ pub struct TransactionRequest {
     pub nonce: Option<U256>,
 
     /// Chain ID (None for mainnet)
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing)]
+    #[serde(rename = "chainId")]
```
mattsse, line 46:
```suggestion
    #[serde(default, rename = "chainId")]
```

mattsse (changes requested):
smol nit

ethers-core/src/types/transaction/request.rs
Rjected: added to both request and 1559 request

gakonst (approved):

tarrencev:
@Rjected I'm still running into this issue after updating foundry to use this PR

```
(code: -32602, message: Invalid params: unknown field `chainId`, expected one of `type`, `from`, `to`, `gasPrice`, `maxFeePerGas`, `gas`, `value`, `data`, `nonce`, `accessList`, `maxPriorityFeePerGas`., data: None)
```

It looks like it is still serialized into the txn `\"chainId\":\"0x2a\"`

tarrencev:
Ah nevermind. I sanity checked forge paths and it was using an old path under `.cargo` rather than the one in `.foundry`

PR #882, opened 2022-02-08
Artifact ids

lattejed:
@mattsse this is related to https://github.com/gakonst/foundry/pull/662

This just adds in an `ArtifactId` and has `into_artifacts` return it.

I think this warrants ongoing discussion, but I've continued that in the PR linked above.

(There are some unrelated warnings and errors here that I've left alone)

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifact_output.rs
+++ b/ethers-solc/src/artifact_output.rs
@@ -16,6 +16,7 @@ use std::{
 };
 
 /// Represents unique artifact metadata for identifying artifacts on output
+#[derive(Debug, Clone)]
 pub struct ArtifactId {
     /// `artifact` cache path
     pub path: PathBuf,
@@ -24,6 +25,17 @@ pub struct ArtifactId {
     pub source: PathBuf,
     /// `solc` version that produced this artifact
     pub version: Version,
+}
+
+impl ArtifactId {
+    /// Returns a <filename>:<name> slug that identifies an artifact
+    pub fn slug(&self) -> String {
+        format!(
+            "{}.json:{}",
+            self.path.file_stem().unwrap().to_string_lossy(),
+            self.name
+        )
+    }
 }
 
 /// Represents an artifact file representing a [`crate::Contract`]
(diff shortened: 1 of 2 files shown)


ethers-solc/src/artifact_output.rs
```diff
@@ -15,6 +15,36 @@ use std::{
     path::{Path, PathBuf},
 };
 
+/// Represents unique artifact metadata for identifying artifacts on output
+#[derive(Debug, Clone)]
```
mattsse, line 19: Also needs PartialEq Eq Hash so we can use them as keys

mattsse (changes requested):
this is great, 

before we can merge this:
* needs a companion on foundry as this breaks, so we can merge both at once
* some derives

lattejed:
Thanks @mattsse, good idea to make them key-able.

Please look at the discussion [here](https://github.com/gakonst/foundry/pull/662#issuecomment-1032332855) it's related to this. Trying to clarify a couple of things.

lattejed:
@mattsse I'm adding a multi-version compile test to this for adding versioned linking. I've found an issue in `Graph::resolve_multiple_versions` where needed versions are getting pruned incorrectly, which results in a failed compile. Do you happen to know what that might be?

lattejed:
Worth pointing out that I'm not sure this is a regression since the same branch works ok with foundry locally. Possibly a config issue? I'm pretty sure I've dealt with those already tho.

mattsse:
Good catch, you probably found an existing issue.
with this repro I can debug this.

But I think we can go ahead with it anyway? so let's ignore the `can_get_versioned_linkrefs` until I found the issue.

what's missing to complete https://github.com/gakonst/foundry/pull/662 ?

gakonst (approved):
lgtm. issue seems unrelated, i ignored the failing test, fixed fmt'ing and fixed the remaining failing tests. let's get this over the line so we merge the foundry PR, and we can follow up with
1. Making Project accept filters
2. Fix the linking stuff

PR #883, opened 2022-02-08
feat(solc): add Reporter type

mattsse:
## Motivation
Ref https://github.com/gakonst/foundry/issues/628
Povide a way to emit custom output at certain points in the compiler pipeline

## Solution
use a global `Reporter` that gets invoked in the code, the `Reporter` can the invoked exactly once `ethers_solc::report::init(BasicStdoutReporter::default());`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/compile/project.rs
+++ b/ethers-solc/src/compile/project.rs
@@ -80,6 +80,7 @@ use crate::{
     cache::ArtifactsCache,
     error::Result,
     output::AggregatedCompilerOutput,
+    report,
     resolver::GraphEdges,
     ArtifactOutput, CompilerInput, Graph, Project, ProjectCompileOutput, ProjectPathsConfig, Solc,
     Sources,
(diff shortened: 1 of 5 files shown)


ethers-solc/src/compile/project.rs
```diff
@@ -343,7 +344,10 @@ fn compile_sequential(
             input.sources.len(),
             input.sources.keys()
         );
-        let output = solc.compile(&input)?;
+
+        report::solc_spawn(&solc, &version, &input);
+        let output = solc.compile_exact(&input)?;
```
gakonst, line 349: why compile -> compile_exact?

gakonst (approved):
lgtm - we need to use `init(Report::new(BasicStdoutReporter::default()))` on foundry right?

ethers-solc/src/compile/project.rs
mattsse: drive by fix to make it the same as in the parallel version, since the input already includes the sources and all its imports we want to limit the output to exactly those.

mattsse:
> lgtm - we need to use `init(Report::new(BasicStdoutReporter::default()))` on foundry right?

exactly, if we need more verbosity we can simply implement another reporter

mattsse:
will fix all cargo warnings in a follow up, there are a bunch of those

PR #884, opened 2022-02-08
feat(providers): Log methods for PendingTransaction

wolflo:
## Motivation

I frequently want to print the transaction hash of a pending transaction before it is mined. Currently, this is pretty verbose, because you need a let binding for the transaction:
```rust
let tx = token.transfer(recipient, amt);
let pending = tx.send().await?;
println!("Tx hash: {:?}", *pending);
let mined = pending.await?;
```

## Solution

Add two convenience methods on the `PendingTransaction<_>` type, `log()` and `log_msg()`.
```rust
let mined = token.transfer(recipient, amt).send().await?.log_msg("Tx hash").await?;
```

Another approach that isn't just for printing might be to add a method to `PendingTransaction<_>` that returns `(TxHash, Self)`, which would turn the above into:
```rust
let (hash, fut) = token.transfer(recipient, amt).send().await?.hash();
println!("Tx hash: {:?}", hash);
let mined = fut.await?;
```

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-providers/src/pending_transaction.rs
+++ b/ethers-providers/src/pending_transaction.rs
@@ -25,6 +25,32 @@ use wasm_timer::Delay;
 /// once the transaction has enough `confirmations`. The default number of confirmations
 /// is 1, but may be adjusted with the `confirmations` method. If the transaction does not
 /// have enough confirmations or is not mined, the future will stay in the pending state.
+///```
+/// use ethers_providers::{Provider, Http, Middleware};
+/// use ethers_core::types::TransactionRequest;
+/// use ethers_core::utils::Ganache;
+/// use std::convert::TryFrom;
+///
+/// # #[tokio::main(flavor = "current_thread")]
+/// # async fn main() -> Result<(), Box<dyn std::error::Error>> {
+/// # let ganache = Ganache::new().spawn();
+/// # let client = Provider::<Http>::try_from(ganache.endpoint()).unwrap();
+/// # let accounts = client.get_accounts().await?;
+/// # let from = accounts[0];
+/// # let to = accounts[1];
+/// # let balance_before = client.get_balance(to, None).await?;
+/// let tx = TransactionRequest::new().to(to).value(1000).from(from);
+/// let receipt = client
+///     .send_transaction(tx, None)
+///     .await?                           // PendingTransaction<_>
+///     .log_msg("Pending transfer hash") // print pending tx hash with message
+///     .await?;                          // Result<Option<TransactionReceipt>, _>
+/// # let _ = receipt;
+/// # let balance_after = client.get_balance(to, None).await?;
+/// # assert_eq!(balance_after, balance_before + 1000);
+/// # Ok(())
+/// # }
+///```
 #[pin_project]
 pub struct PendingTransaction<'a, P> {
     tx_hash: TxHash,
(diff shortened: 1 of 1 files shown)


mattsse:
I can see how this is useful,

alternatively we could make this akin to the `std::iter::inscpect` function:

```rust

pub fn inspect<F>(self, f: F) -> Self where F: FnMut(Self) {
   f(&self);
   self
}
```

so you can simply to 
```
let mined = token.transfer(recipient, amt).send().await?.inspect(|tx| println!(".... {}", *tx)).await?;

```

wolflo:
I like that too. I probably like the current version the least, because it's obviously use-case specific. Though maybe we want to keep `log()` and/or `log_msg()` in conjunction with a more useful version just for the succinctness.

ethers-providers/src/pending_transaction.rs
```diff
@@ -78,6 +104,22 @@ impl<'a, P: JsonRpcClient> PendingTransaction<'a, P> {
...
 }
 
+impl<'a, P> PendingTransaction<'a, P> {
+    pub fn inspect<F>(self, mut f: F) -> Self
+    where
+        F: FnMut(&Self),
+    {
+        f(&self);
+        self
+    }
```
gakonst, line 114:
```suggestion
    }
    
```

ethers-providers/src/pending_transaction.rs
```diff
@@ -78,6 +104,22 @@ impl<'a, P: JsonRpcClient> PendingTransaction<'a, P> {
...
+    pub fn log_msg<S: std::fmt::Display>(self, msg: S) -> Self {
+        self.inspect(|s| println!("{}: {:?}", msg, *s))
+    }
```
gakonst, line 117:
```suggestion
    }
    
```

ethers-providers/src/pending_transaction.rs
```diff
@@ -78,6 +104,22 @@ impl<'a, P: JsonRpcClient> PendingTransaction<'a, P> {
     }
 }
 
+impl<'a, P> PendingTransaction<'a, P> {
+    pub fn inspect<F>(self, mut f: F) -> Self
```
gakonst, line 108:
```suggestion
    /// Allows inspecting the content of a pending transaction in a builder-like way to avoid
    /// more verbose calls, e.g.: 
    /// `let mined = token.transfer(recipient, amt).send().await?.inspect(|tx| println!(".{}", *tx)).await?;`
    pub fn inspect<F>(self, mut f: F) -> Self
```

ethers-providers/src/pending_transaction.rs
```diff
@@ -78,6 +104,22 @@ impl<'a, P: JsonRpcClient> PendingTransaction<'a, P> {
...
+    where
+        F: FnMut(&Self),
+    {
+        f(&self);
+        self
+    }
+    pub fn log_msg<S: std::fmt::Display>(self, msg: S) -> Self {
+        self.inspect(|s| println!("{}: {:?}", msg, *s))
+    }
+    pub fn log(self) -> Self {
```
gakonst, line 118:
```suggestion
    /// Logs the pending transaction's hash
    pub fn log(self) -> Self {
```

ethers-providers/src/pending_transaction.rs
```diff
@@ -78,6 +104,22 @@ impl<'a, P: JsonRpcClient> PendingTransaction<'a, P> {
...
 
+impl<'a, P> PendingTransaction<'a, P> {
+    pub fn inspect<F>(self, mut f: F) -> Self
+    where
+        F: FnMut(&Self),
+    {
+        f(&self);
+        self
+    }
+    pub fn log_msg<S: std::fmt::Display>(self, msg: S) -> Self {
```
gakonst, line 115:
```suggestion
    /// Logs the pending transaction hash along with a custom message before it.
    pub fn log_msg<S: std::fmt::Display>(self, msg: S) -> Self {
```

gakonst (approved):

PR #889, opened 2022-02-09
Add ENS avatar and TXT records resolution

sbihel:
This PR adds support for resolution of text records and also adds a helper to resolve avatars and validate them (as well as provide a HTTP link to make it easier for the user). This is mostly a translation of `ethers.js` implementation.

I am not terribly happy with the current interface and place in the code, so I'm hoping you have an opinion on the matter.

## Motivation

We simply need to resolve records and specifically avatars of ENS domains.

## Solution

TXT records can be resolved using `resolve_field` (I have encountered both the terms "record" and "field", which isn't great), and the `avatar` record can also be resolved using `resolve_avatar` to automatically find the target image and validate ownership in the case of NFTs.

## PR Checklist

- [x] Added Tests
  * Only for avatars, and only using live domains, which will inevitably be a problem.
- [x] Added Documentation
- [x] Updated the changelog

## Questions
1. Do you have an idea on how to split up the ERC-721/1155 code nicely?
2. Would it be preferable to use an `Option` in the return type of `resolve_avatar`?
3. Is there a configuration system in place right now to be able to modify the IPFS HTTP gateway?


ethers-providers/src/provider.rs
```diff
@@ -786,6 +786,151 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+    /// if not configured)
+    ///
+    /// # Panics
+    ///
+    /// If the bytes returned from the ENS registrar/resolver cannot be interpreted as
+    /// a string. This should theoretically never happen.
+    async fn resolve_avatar(&self, ens_name: &str) -> Result<Url, ProviderError> {
+        let field = self.resolve_field(ens_name, "avatar").await?;
+        let url = Url::from_str(&field).map_err(|e| ProviderError::CustomError(e.to_string()))?;
+        let owner = self.resolve_name(ens_name).await?;
```
mattsse, line 799: looks like we can join these? `try_join!`?

ethers-providers/src/provider.rs
```diff
@@ -786,6 +786,151 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+        match url.scheme() {
+            "https" | "data" => Ok(url),
+            "ipfs" => erc::http_link_ipfs(url).map_err(ProviderError::CustomError),
+            "eip155" => {
```
mattsse, line 803: I'd like to make this a standalone function

ethers-providers/src/provider.rs
```diff
@@ -786,6 +786,151 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+                let split: Vec<&str> = url.path().trim_start_matches("1/").split(':').collect();
+                let (inner_scheme, inner_path) = if split.len() == 2 {
+                    (split[0], split[1])
```
mattsse, line 806: this could be `str::split_once(':').ok_or_else(..)`, right?

ethers-providers/src/provider.rs
```diff
@@ -786,6 +786,151 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+                } else {
+                    return Err(ProviderError::CustomError("Unsupported ERC link".to_string()));
+                };
+
+                let token_split: Vec<&str> = inner_path.split('/').collect();
```
mattsse, line 811: also `split_once`?

ethers-providers/src/provider.rs
```diff
@@ -786,6 +786,151 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+                };
+                let selector: Selector = match inner_scheme {
+                    "erc721" => {
+                        let tx = TransactionRequest {
+                            data: Some([&erc::ERC721_SELECTOR[..], &token_id].concat().into()),
+                            to: Some(NameOrAddress::Address(contract_addr)),
+                            ..Default::default()
+                        };
+                        let data = self.call(&tx.into(), None).await?;
+                        if decode_bytes::<Address>(ParamType::Address, data) != owner {
```
mattsse, line 845: I think `Address::decode(data)` should also work here? using the `Abidecode` trait

ethers-providers/src/provider.rs
```diff
@@ -786,6 +786,151 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+                        &metadata_url.path().replace("%7Bid%7D", &hex::encode(&token_id)),
+                    );
+                }
+                if metadata_url.scheme() == "ipfs" {
+                    metadata_url =
+                        erc::http_link_ipfs(metadata_url).map_err(ProviderError::CustomError)?;
+                }
+                let metadata: erc::Metadata = reqwest::get(metadata_url)
+                    .await
+                    .map_err(|e| ProviderError::CustomError(e.to_string()))?
```
mattsse, line 897: is this necessary?
does ProviderError not have a variant that represents the reqwest Error type? if not probably worth adding so we can use `?` instead

ethers-providers/src/lib.rs
```diff
@@ -263,6 +266,14 @@ pub trait Middleware: Sync + Send + Debug {
         self.inner().lookup_address(address).await.map_err(FromErr::from)
     }
 
+    async fn resolve_avatar(&self, ens_name: &str) -> Result<Url, Self::Error> {
+        self.inner().resolve_avatar(ens_name).await.map_err(FromErr::from)
+    }
+
+    async fn resolve_field(&self, ens_name: &str, field: &str) -> Result<String, Self::Error> {
```
mattsse, line 273: please document these trait functions, or link to the matching functions, that are document.

ethers-providers/src/provider.rs
```diff
@@ -786,6 +786,151 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
         self.query_resolver(ParamType::String, &ens_name, ens::NAME_SELECTOR).await
     }
 
+    /// Returns the avatar HTTP link of the avatar that the `ens_name` resolves to (or None
+    /// if not configured)
+    ///
+    /// # Panics
+    ///
```
mattsse, line 793: can we add an example here, basically copy one test case from the test

ethers-providers/src/provider.rs
```diff
@@ -786,6 +786,151 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+                }
+            }
+            _ => Err(ProviderError::CustomError("Unsupported scheme".to_string())),
+        }
+    }
+
+    /// Fetch a field for the `ens_name` (no None if not configured).
+    ///
+    /// # Panics
+    ///
```
mattsse, line 919: would benefit from an example

ethers-providers/src/provider.rs
```diff
@@ -786,6 +786,151 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+                            to: Some(NameOrAddress::Address(contract_addr)),
+                            ..Default::default()
+                        };
+                        let data = self.call(&tx.into(), None).await?;
+                        if decode_bytes::<Address>(ParamType::Address, data) != owner {
+                            return Err(ProviderError::CustomError("Incorrect owner.".to_string()));
+                        }
+                        [0xc8, 0x7b, 0x56, 0xdd]
+                    }
+                    "erc1155" => {
```
mattsse, line 850: if possible, we should make them 2 separate functions

mattsse (changes requested):
> Do you have an idea on how to split up the ERC-721/1155 code nicely?

few nits and suggestions, can improve once addressed.

> Would it be preferable to use an Option in the return type of resolve_avatar?

how would this manifest itself in the response? statuscode? failed decoded?

> Is there a configuration system in place right now to be able to modify the IPFS HTTP gateway?

think this is fixed to ipfs.io?

ethers-providers/src/provider.rs
sbihel: Added in cd056af

ethers-providers/src/provider.rs
sbihel: Improved in 78d6118

ethers-providers/src/provider.rs
sbihel: I can't find that method, I just copied how it was done in `query_resolver`

ethers-providers/src/provider.rs
sbihel: `split_once` won't fail if there are more of the split character, I feel like it's better to fail early? But because of the checks/deserialisation that happen right after, we wouldn't make invalid queries. So I'm on the fence

ethers-providers/src/provider.rs
sbihel: 307bdcc adds a `ERCToken` type and another method to resolve a token. Not sure about how future this is though :thinking:

sbihel:
> > Would it be preferable to use an Option in the return type of resolve_avatar?
> 
> how would this manifest itself in the response? statuscode? failed decoded?

I was thinking of scenarios like: unsupported scheme, or invalid ownership. They're not necessarily errors -- but I don't think it matters too much considering I'm mostly using the opaque `CustomError`.

> > Is there a configuration system in place right now to be able to modify the IPFS HTTP gateway?
> 
> think this is fixed to ipfs.io?

Anyone can deploy a gateway, another big one is Cloudflare's. So someone could want to use a different one (e.g. private one, or if ipfs.io is down). But I really don't think it's that important for now.

sbihel:
Oh and also, I'm curious why `cargo fmt` uses nightly?

mattsse:
> Oh and also, I'm curious why `cargo fmt` uses nightly?

we use `cargo +nightly fmt` because that includes more options

ethers-providers/src/provider.rs
```diff
@@ -786,6 +789,145 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
         self.query_resolver(ParamType::String, &ens_name, ens::NAME_SELECTOR).await
     }
 
+    /// Returns the avatar HTTP link of the avatar that the `ens_name` resolves to (or None
+    /// if not configured)
+    ///
+    /// # Example
+    /// ```
```
mattsse, line 796: let's do no_run here
```suggestion
    /// ```no_run
```

ethers-providers/src/provider.rs
```diff
@@ -786,6 +789,145 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+                }
+            }
+            _ => Err(ProviderError::CustomError("Unsupported scheme".to_string())),
+        }
+    }
+
+    /// Returns the URL (not necesserily HTTP) of the image behind a token.
+    ///
+    /// # Example
+    /// ```
```
mattsse, line 875:
```suggestion
    /// ```no_run
```

mattsse (approved):
lgtm!
thanks!
needs nightly fmt

gakonst (approved):
nice work!

gakonst:
Tests seem to be failing here, any clue why? Maybe the IPFS path changed?

```
  left: `Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("creature.mypinata.cloud")), port: None, path: "/ipfs/QmVHmNwVpiSebryKDZX7B3MXEUZVSACgTpDMPiD6j6Nbcj/9018.jpg", query: None, fragment: None }`,
 right: `Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("creature.mypinata.cloud")), port: None, path: "/ipfs/QmNwj3aUzXfG4twV3no7hJRYxLLAWNPk6RrfQaqJ6nVJFa/9018.jpg", query: None, fragment: None }`', ethers-providers/src/provider.rs:1520:9
````

sbihel:
Indeed, hopefully the new one won't change for a while

Issue #900, opened 2022-02-11
Increase Multicall Call Limit to Past 16 


3lpsy:
**Is your feature request related to a problem? Please describe.**
Currently, the Multicall interface can only contain up to 16 calls at a time as shown [here](https://github.com/gakonst/ethers-rs/blob/master/ethers-contract/src/multicall/mod.rs#L196). Part of this reason is the fact that the call will need to detokenize the responses into a tuple of a known size.

**Describe the solution you'd like**
From my understanding, I would like to be able to make a multi call that 1) contains more than 16 calls, and 2) allows the multi call to contain an unknown number of calls (computed at runtime) which returns an uknown number of responses. I would be interested in working on this; however, I'm not sure I possess enough familiarity with the library to accomplish it on my own. 

**Describe alternatives you've considered**
One potential solution, from my understanding, is to not attempt to detokenize the response. This could push the responsibility from recovering the responses to the user. 

**Additional context**
From my understanding, accomplishing this would require replacing / providing alternatives for the use of both Detokenize and Token. I've included some of the relevant code below. 

```
pub fn add_call<D: Detokenize>(&mut self, call: ContractCall<M, D>) -> &mut Self {
  assert!(self.calls.len() < 16, "Cannot support more than {} calls", 16);
  ///....
}

pub async fn call<D: Detokenize>(&self) -> Result<D, ContractError<M>> {
        let (_block_number, return_data) = contract_call.call().await?;
        let tokens = self.calls.iter().zip(&return_data).map(|(call, bytes)| {
                let mut tokens: Vec<Token> = call.function.decode_output(bytes.as_ref())?;
                Ok(match tokens.len() {
                    0 => Token::Tuple(vec![]),
                    1 => tokens.remove(0),
                    _ => Token::Tuple(tokens),
                })
            }).collect::<Result<Vec<Token>, ContractError<M>>>()?;
        let tokens = vec![Token::Tuple(tokens)];
        let data = D::from_tokens(tokens)?;
        Ok(data)
}
```

gakonst:
I think the "right" way to do this would be:

```rust
pub async fn call_raw(&self) -> Result<Vec<Token>, ContractError<M>> {
        let (_block_number, return_data) = contract_call.call().await?;
        let tokens = self.calls.iter().zip(&return_data).map(|(call, bytes)| {
                let mut tokens: Vec<Token> = call.function.decode_output(bytes.as_ref())?;
                Ok(match tokens.len() {
[...]

PR #915, opened 2022-02-15
Expose New `call_raw` API that permits MultiCalls without Detokenization

3lpsy:
## Motivation

Multicall's `add_call` method is capped at 16 calls due to detokenization (tuple) contraints. Such a cap may force users to make more calls than necessary. 

References:  #900 

## Solution

Expose a new `call_raw` method on the MultiCall interface that skips detokenization. The `call` method can then wrap `call_raw` to avoid code duplication and add in the detokenization. 

## Additional Notes
- While current tests should ensure no regressions for `call`, only a simple test was added for `call_raw`. This may need to be improved.
- The 16 call limit panic is moved from `add_call` to `call`. As it's a panic, it is not likely to cause issues in current deployments; however, it may be worth pointing out in Changelog in case anyone relied upon it.
- The test cases for `multicall_aggregate` did not appear to be compatible with ganache ^7.0.0 as they assumed default accounts were seeded with 100 ether, not 1000 ether. Changes were made to `spawn` to accommodate this
- Documentation was added / modified for `call_raw` and `call`; however, formatting still needs to be checked

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-contract/src/multicall/mod.rs
+++ b/ethers-contract/src/multicall/mod.rs
@@ -187,14 +187,7 @@ impl<M: Middleware> Multicall<M> {
     }
 
     /// Appends a `call` to the list of calls for the Multicall instance
-    ///
-    /// # Panics
-    ///
-    /// If more than the maximum number of supported calls are added. The maximum
-    /// limits is constrained due to tokenization/detokenization support for tuples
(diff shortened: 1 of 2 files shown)


roynalnaruto (approved):
Looks good to me!

gakonst (approved):
nice work!

PR #917, opened 2022-02-16
Remove `limit` on `Filter`

scrogson:
## Motivation

While building a CLI to decode logs I reached for the `limit` method on `Filter` in order to implement some sort of pagination. To my surprise, it does not seem to have any effect. Combing through the documentation for Alchemy and Infura, the entries for `eth_getLogs` do not appear to include any `limit` option.

After pinging @gakonst in telegram, he suggested I make a PR to remove this as it doesn't appear to be part of the JSON-RPC specification.

## Solution

Removed the `limit` option from `Filter`.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


gakonst:
good catch

PR #934, opened 2022-02-19
wip/experimental - add vyper support

joshieDo:
I initially tried to make something of a compiler trait, but with how much `solc` is spread, I kinda gave up and tried to keep the changes as compartmentalized as possible.

What do you think? Is it something that might be of interest pursuing? What would be the direction that you'd suggest?

Depends on: https://github.com/vyperlang/vyper/issues/2674
``
I tried out building `year/yearn-vaults` with foundry and this branch

`brownie compile  56.32s user 0.77s system 97% cpu 58.326 total`
`target/debug/forge build --evm-version berlin  51.66s user 0.35s system 96% cpu 53.705 total`

`forge run` also works fine with this branch.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -1872,7 +1872,11 @@ pub struct StorageType {
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq, Hash)]
 #[serde(rename_all = "camelCase")]
 pub struct Error {
-    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        deserialize_with = "serde_helpers::default_on_error"
+    )]
     pub source_location: Option<SourceLocation>,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub secondary_source_locations: Vec<SecondarySourceLocation>,
--- a/ethers-solc/src/compile/project.rs
+++ b/ethers-solc/src/compile/project.rs
@@ -76,18 +76,19 @@
 
 use crate::{
     artifact_output::Artifacts,
-    artifacts::{Settings, VersionedSources},
+    artifacts::{Settings, Source, VersionedSources},
     cache::ArtifactsCache,
     error::Result,
     output::AggregatedCompilerOutput,
     report,
     resolver::GraphEdges,
+    vyper::Vyper,
     ArtifactOutput, CompilerInput, Graph, Project, ProjectCompileOutput, ProjectPathsConfig, Solc,
     Sources,
 };
 use rayon::prelude::*;
 
-use std::collections::btree_map::BTreeMap;
+use std::{collections::btree_map::BTreeMap, path::PathBuf};
 
 #[derive(Debug)]
 pub struct ProjectCompiler<'a, T: ArtifactOutput> {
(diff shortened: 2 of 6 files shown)


ethers-solc/src/artifacts/mod.rs
```diff
@@ -1872,7 +1872,11 @@ pub struct StorageType {
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq, Hash)]
 #[serde(rename_all = "camelCase")]
 pub struct Error {
-    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        deserialize_with = "serde_helpers::default_on_error"
+    )]
```
mattsse, line 1879: does this now always deserialize into `Some(SourceLocation::default())`?
what's the motivation behind this?

ethers-solc/src/compile/project.rs
```diff
@@ -324,37 +325,69 @@ fn compile_sequential(
...
+                .normalize_evm_version(&version)
+                .with_remappings(paths.remappings.clone());
+
+            input.language = "Vyper".to_string();
 
-        report::solc_spawn(&solc, &version, &input);
-        let output = solc.compile_exact(&input)?;
-        report::solc_success(&solc, &version, &output);
-        tracing::trace!("compiled input, output has error: {}", output.has_error());
+            let output = vyper.compile_exact(&input)?;
```
mattsse, line 358: this assumes that there are solidity _and_ vyper files mixed in the same project, is that common?
I'd rather have separate modes depending on the project so _either_ solidity _or_ vyper?

mattsse (changes requested):
Thanks for looking into this!

what are the differences between `solc` and `vyper` executable args?

It seems like vyper has a limited subset of the solc `CompilerOutput` `Input`?

https://vyper.readthedocs.io/en/stable/compiling-a-contract.html#compiler-input-and-output-json-description

so perhaps it'd be justified to add proper vyper bindings `VyperCompilerInput/Output` as well?

I think we can use this to kick off the `ethers-compile` abstraction, and put everything compiler (executable) related behind a trait.
which I guess includes
* input
* output
* CLI wrapper (like `Solc`/`Vyper`)

Ideally we can reuse everything related to caching and the whole `ProjectCompiler` pipeline in general for every implementation of this trait.

So I'm thinking, on its own this PR is great initial progress to get vyper supported but should perhaps be it's own `ethers-vyper` crate.

eventually, we'll have

```
└── ethers-compile
    ├── ethers-solc
    └── ethers-vyper
    └── ethers-cairo
    └── ethers-fe even
```

So the next step would be to come up with a `trait Compiler` abstraction that we can plug into the `ethers-compile` `Project` and `ProjectCompiler`

joshieDo:
opened https://github.com/gakonst/ethers-rs/pull/936 
sorry for the mess. feel free to close this one

ethers-solc/src/compile/project.rs
joshieDo: https://github.com/yearn/yearn-vaults

has both for example, which is the one i've been using to try it out

ethers-solc/src/artifacts/mod.rs
joshieDo: `SourceLocation` format is different for vyper compiler output, and was giving out an error.

ethers-solc/src/compile/vyper.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    pub fn compile_as<T: Serialize, D: DeserializeOwned>(&self, input: &T) -> Result<D> {
+        let output = self.compile_output(input)?;
+        Ok(serde_json::from_slice(&output)?)
+    }
+
+    pub fn compile_output<T: Serialize>(&self, input: &T) -> Result<Vec<u8>> {
+        let mut cmd = Command::new(&self.vyper);
+
+        // Filter out solc arguments
```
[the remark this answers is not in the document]
joshieDo: exactly

ethers-solc/src/compile/vyper.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn default() -> Self {
+        if let Ok(vyper) = std::env::var("VYPER_PATH") {
+            return Vyper::new(vyper)
+        }
+
+        Vyper::new(VYPER)
+    }
+}
+
+impl Vyper {
```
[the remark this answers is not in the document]
joshieDo: Most likely. I copied everything at first since I didn't know how much it would be needed

gakonst:
Closing and we can move over the discussion to 936!

PR #935, opened 2022-02-19
feat(etherscan): add caching

gakonst:
As title, will be useful for avoidign rate limiting when doing mainnet forking in foundry and fetching ABIs/sources for tracing

### DIFF
--- a/ethers-etherscan/src/contract.rs
+++ b/ethers-etherscan/src/contract.rs
@@ -234,6 +234,14 @@ impl Client {
     /// # }
     /// ```
     pub async fn contract_abi(&self, address: Address) -> Result<Abi> {
+        // apply caching
+        if let Some(ref cache) = self.cache {
+            let path = cache.join("abis").join(format!("{:?}.json", address));
+            let reader = std::io::BufReader::new(std::fs::File::create(path)?);
+            let abi = serde_json::from_reader(reader)?;
+            return Ok(abi)
+        }
+
         let query = self.create_query("contract", "getabi", HashMap::from([("address", address)]));
         let resp: Response<String> = self.get_json(&query).await?;
         Ok(serde_json::from_str(&resp.result)?)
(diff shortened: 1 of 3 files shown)


ethers-etherscan/src/contract.rs
```diff
@@ -234,6 +234,14 @@ impl Client {
     /// # }
     /// ```
     pub async fn contract_abi(&self, address: Address) -> Result<Abi> {
+        // apply caching
+        if let Some(ref cache) = self.cache {
+            let path = cache.join("abis").join(format!("{:?}.json", address));
+            let reader = std::io::BufReader::new(std::fs::File::create(path)?);
+            let abi = serde_json::from_reader(reader)?;
+            return Ok(abi)
```
mattsse, line 242: should we instead continue with the request if this results in an error, like file does not exist?

ethers-etherscan/src/lib.rs
```diff
@@ -26,9 +26,21 @@ pub struct Client {
...
 
 impl Client {
+    pub fn new_cached(
+        chain: Chain,
+        api_key: impl Into<String>,
+        cache: Option<PathBuf>,
+    ) -> Result<Self> {
+        let mut this = Self::new(chain, api_key)?;
+        this.cache = cache;
+        Ok(this)
```
mattsse, line 41: alternatively call `Self::new_cached(chain, api_key, None)` from `Self::new`

ethers-etherscan/src/contract.rs
```diff
@@ -254,6 +262,14 @@ impl Client {
     /// # }
     /// ```
     pub async fn contract_source_code(&self, address: Address) -> Result<ContractMetadata> {
+        // apply caching
+        if let Some(ref cache) = self.cache {
+            let path = cache.join("sources").join(format!("{:?}.json", address));
+            let reader = std::io::BufReader::new(std::fs::File::create(path)?);
+            let src = serde_json::from_reader(reader)?;
+            return Ok(src)
```
mattsse, line 270: see comment above.

gakonst:
FYI @onbjerg this had seemed to work but I never tested it. May want to take it over so we don't hit rate limits when tracing

gakonst:
Ack

PR #936, opened 2022-02-20
wip ethers-compile 

joshieDo:
A more structured approach at abstracting away the compiler than https://github.com/gakonst/ethers-rs/pull/934

* Added ethers-compile, and kept ethers-solc

ethers-compile
* `CompilerTrait`
* `GenericCompiler` (with a `CompilerKindEnum` inside), `Vyper` and `Solc` all `impl CompilerTrait`
* Uses a `GenericCompiler`  for as long as possible, and then instantiates either `vyper` or `solc` at `compile_sequential` or `compile_parallel` with a `into_kind()` method
* `Report` moved from `solc`
* Changed the structure so the compilers are under `src/compile/compilers`
* `CompilerSources` has a method called `triage` that loops through all sources and separates `*.vy` from `*.solc` and creates the actual compiler object that it needs. (probably this could be done on `Graph` but i didn't go much deep there...)

Working with https://github.com/joshiedo/foundry/tree/feat/ethers-compile

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


ethers-compile/src/compile/compilers/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+    fn arg(&mut self, arg: String);
+    fn args(&mut self, args: Vec<String>);
+    fn get_args(&self) -> Vec<String>;
+
+    fn version(&self) -> Version;
+    fn language(&self) -> String;
+
+    fn compile_exact(&self, input: &CompilerInput) -> Result<CompilerOutput>;
+    fn compile(&self, input: &CompilerInput) -> Result<CompilerOutput>;
+}
```
mattsse, line 28: this is a reasonable first draft.

I don't think we'd need the `arg` related functions.

coming from this comment https://github.com/gakonst/ethers-rs/pull/934#pullrequestreview-888078740 this trait really should do compile only and we want to that be generic over input and output.

so we need something like 

```rust
pub trait Compiler {
  type Input;
  type Output;

  fn compile(&self, input: &Self::Input) -> Result<Self::Output>;
}
```

mattsse (changes requested):
This is great progress.
before we can move this forwards I'd suggest to make `ethers-compile` a clean slate crate. right now it's basically a duplicate of `ethers-solc`?

let's start in ethers-compile with the compile trait and then gradually move more core logic from solc -> compile.

ethers-compile/src/compile/compilers/mod.rs
gakonst: Agree! @joshieDo one way to approach this would be the following:
1. Create the trait mentioned above
2. Impl the trait for ethers-solc
3. Make Foundry's compilation pipeline use the trait
4. Implement the trait for ethers-vyper (or whatever)
5. Create a multi-language implementor of the trait (e.g. instantiated with Ethers-solc, Ethers-vyper, Ethers-fe) which separates sol/vy/fe files, sends them to each job and creates a unified artifact output

charles-cooper:
@abigger87 @mattsse @joshieDo is https://github.com/vyperlang/vyper/issues/2674 still a blocker?

joshieDo:
@charles-cooper no it's not. @abigger87 is preparing the groundwork, and then I can pick this one up again and just plug vyper in.

mattsse:
awesome! lmk if you need some pointers/ have some questions

gakonst:
Gonna close this as unplanned for now, and consider this for a larger refactor if we figure out a nice abstraction for sourcemaps, asts etc.

mattsse:
yep, would need to revisit for fuelvm integration anyways

PR #939, opened 2022-02-21
feat(etherscan): account endpoints

tomtau:
## Motivation

https://github.com/gakonst/ethers-rs/issues/485

Add etherscan.io account endpoints

## Solution

Added them.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-etherscan/src/account.rs
+++ b/ethers-etherscan/src/account.rs
@@ -323,8 +323,8 @@ impl Client {
     /// # async fn main() {
     ///     let client = Client::new(Chain::Mainnet, "API_KEY").unwrap();
     ///     let balance = client
-    ///         .get_ether_balance_single("0x58eB28A67731c570Ef827C365c89B5751F9E6b0a", None)
-    ///         .await.unwrap();
+    ///         .get_ether_balance_single(&"0x58eB28A67731c570Ef827C365c89B5751F9E6b0a".parse().unwrap(),
+    ///         None).await.unwrap();
     /// # }
     /// ```
     pub async fn get_ether_balance_single(
@@ -358,8 +358,8 @@ impl Client {
     /// # async fn main() {
     ///     let client = Client::new(Chain::Mainnet, "API_KEY").unwrap();
     ///     let balances = client
-    ///         .get_ether_balance_multi(&vec!["0x58eB28A67731c570Ef827C365c89B5751F9E6b0a"], None)
-    ///         .await.unwrap();
+    ///         .get_ether_balance_multi(&vec![&"0x58eB28A67731c570Ef827C365c89B5751F9E6b0a".parse().unwrap()],
+    ///         None).await.unwrap();
     /// # }
     /// ```
     pub async fn get_ether_balance_multi(
@@ -393,8 +393,8 @@ impl Client {
     /// # async fn main() {
     ///     let client = Client::new(Chain::Mainnet, "API_KEY").unwrap();
     ///     let txs = client
-    ///         .get_normal_transactions("0x58eB28A67731c570Ef827C365c89B5751F9E6b0a", None)
-    ///         .await.unwrap();
+    ///         .get_normal_transactions(&"0x58eB28A67731c570Ef827C365c89B5751F9E6b0a".parse().unwrap(),
+    ///         None).await.unwrap();
     /// # }
     /// ```
     pub async fn get_normal_transactions(
--- a/ethers-etherscan/src/account.rs
+++ b/ethers-etherscan/src/account.rs
@@ -73,6 +73,28 @@ pub enum GenesisOption<T> {
     Some(T),
 }
 
+impl<T> From<GenesisOption<T>> for Option<T> {
+    fn from(value: GenesisOption<T>) -> Self {
+        match value {
+            GenesisOption::Some(value) => Some(value),
+            _ => None,
+        }
+    }
+}
+
+impl<T> GenesisOption<T> {
+    pub fn is_genesis(&self) -> bool {
+        matches!(self, GenesisOption::Genesis)
+    }
+
+    pub fn value(&self) -> Option<&T> {
+        match self {
+            GenesisOption::Some(value) => Some(value),
+            _ => None,
+        }
+    }
+}
+
 /// The raw response from the transaction list API endpoint
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
@@ -108,18 +130,21 @@ pub struct NormalTransaction {
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct InternalTransaction {
-    pub block_number: String,
+    pub block_number: BlockNumber,
     pub time_stamp: String,
-    pub hash: String,
-    pub from: String,
-    pub to: String,
-    pub value: String,
-    pub contract_address: String,
-    pub input: String,
+    pub hash: H256,
+    pub from: Address,
+    #[serde(with = "jsonstring")]
+    pub to: GenesisOption<Address>,
+    pub value: U256,
+    #[serde(with = "jsonstring")]
+    pub contract_address: GenesisOption<Address>,
+    #[serde(with = "jsonstring")]
+    pub input: GenesisOption<Bytes>,
     #[serde(rename = "type")]
     pub result_type: String,
-    pub gas: String,
-    pub gas_used: String,
+    pub gas: U256,
+    pub gas_used: U256,
     pub trace_id: String,
     pub is_error: String,
     pub err_code: String,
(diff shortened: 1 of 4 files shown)


tomtau:
FYI I checked forks and noticed @1wilkens had a WIP version  https://github.com/1wilkens/ethers-rs/commit/f2eb53732fca388b0af21a756828e8efab8106d0 -- feel free to review if this PR works for you!

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,532 @@
...
+            let txs = client
+                .get_erc721_token_transfer_events(
+                    TokenQueryOption::ByAddressAndContract(
+                        "0x6975be450864c02b4613023c2152ee0743572325",
+                        "0x06012c8cf97bead5deae237070f9587f8e7a266d",
+                    ),
+                    None,
+                )
+                .await;
+            dbg!(&txs);
```
mattsse, line 512:
```suggestion
```

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,532 @@
...
+    #[tokio::test]
+    #[serial]
+    async fn get_mined_blocks_success() {
+        run_at_least_duration(Duration::from_millis(250), async {
+            let client = Client::new_from_env(Chain::Mainnet).unwrap();
+
+            let blocks = client
+                .get_mined_blocks("0x9dd134d14d1e65f84b706d6f205cd5b1cd03a46b", None, None)
+                .await;
+            dbg!(&blocks);
```
mattsse, line 527:
```suggestion
```

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,532 @@
...
+    #[tokio::test]
+    #[serial]
+    async fn get_ether_balance_multi_success() {
+        run_at_least_duration(Duration::from_millis(250), async {
+            let client = Client::new_from_env(Chain::Mainnet).unwrap();
+
+            let balances = client
+                .get_ether_balance_multi(&vec!["0x58eB28A67731c570Ef827C365c89B5751F9E6b0a"], None)
+                .await;
+            dbg!(&balances);
```
mattsse, line 418:
```suggestion
```

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,532 @@
...
+    #[tokio::test]
+    #[serial]
+    async fn get_ether_balance_single_success() {
+        run_at_least_duration(Duration::from_millis(250), async {
+            let client = Client::new_from_env(Chain::Mainnet).unwrap();
+
+            let balance = client
+                .get_ether_balance_single("0x58eB28A67731c570Ef827C365c89B5751F9E6b0a", None)
+                .await;
+            dbg!(&balance);
```
mattsse, line 403:
```suggestion
```

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,532 @@
...
+
+impl ToString for BlockType {
+    fn to_string(&self) -> String {
+        match self {
+            BlockType::CanonicalBlocks => "blocks",
+            BlockType::Uncles => "uncles",
+        }
+        .to_string()
+    }
+}
```
mattsse, line 242: impl `fmt::Display` instead which provides `to_string` automatically

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,532 @@
...
+
+/// Options for querying internal transactions
+pub enum InternalTxQueryOption<A: AsRef<str>> {
+    ByAddress(A),
+    ByTransactionHash(A),
+    ByBlockRange,
+}
+
+/// Options for querying ERC20 or ERC721 token transfers
+pub enum TokenQueryOption<A: AsRef<str>> {
```
mattsse, line 201: I think you can omit the `: AsRef<str>` bound here, since the type is restricted in the API calls as well

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,532 @@
...
+                params
+            }
+            TokenQueryOption::ByContract(contract) => {
+                params.insert("contractaddress", contract.as_ref().to_string());
+                params
+            }
+            TokenQueryOption::ByAddressAndContract(address, contract) => {
+                params.insert("address", address.as_ref().to_string());
+                params.insert("contractaddress", contract.as_ref().to_string());
+                params
```
mattsse, line 222: looks like we're always allocating via `as_ref().to_string()`

we could alternatively make `A: Into<Cow>` so that we don't need to alloc a new string if an address is given as &str

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,532 @@
...
+        run_at_least_duration(Duration::from_millis(250), async {
+            let client = Client::new_from_env(Chain::Mainnet).unwrap();
+
+            let txs = client
+                .get_internal_transactions(
+                    InternalTxQueryOption::ByAddress("0x2c1ba59d6f58433fb1eaee7d20b26ed83bda51a3"),
+                    None,
+                )
+                .await;
+            dbg!(&txs);
```
mattsse, line 453:
```suggestion
```

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,532 @@
...
+
+            let txs = client
+                .get_internal_transactions(
+                    InternalTxQueryOption::ByTransactionHash(
+                        "0x40eb908387324f2b575b4879cd9d7188f69c8fc9d87c901b9e2daaea4b442170",
+                    ),
+                    None,
+                )
+                .await;
+            dbg!(&txs);
```
mattsse, line 473:
```suggestion
```

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,532 @@
...
+        run_at_least_duration(Duration::from_millis(250), async {
+            let client = Client::new_from_env(Chain::Mainnet).unwrap();
+
+            let txs = client
+                .get_erc20_token_transfer_events(
+                    TokenQueryOption::ByAddress("0x4e83362442b8d1bec281594cea3050c8eb01311c"),
+                    None,
+                )
+                .await;
+            dbg!(&txs);
```
mattsse, line 491:
```suggestion
```

mattsse (changes requested):
Awesome, thanks for this.

A few nits and suggestions.

I'm fine with the `AsRef<str>` approach. 

It'd be nice if you could copy some of the tests as examples to the API functions' docs

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,621 @@
...
+
+/// The raw response from the transaction list API endpoint
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct NormalTransaction {
+    pub is_error: String,
+    pub block_number: String,
+    pub time_stamp: String,
+    pub hash: String,
+    pub nonce: String,
```
gakonst, line 26: e.g. this should be U256?

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,621 @@
...
+    pub block_hash: String,
+    pub transaction_index: String,
+    pub from: String,
+    pub to: String,
```
gakonst, line 30: and these `Address`?

gakonst (changes requested):
nice work! just small nit around type defs

tomtau:
@gakonst I tried to put more restricted types on the response fields, but I'm not sure whether it'll be worth it: https://github.com/gakonst/ethers-rs/pull/939/commits/e14d0545deab0fda7b9d200540d02a6e96811f26

There are a few issues at this moment:
- I don't know if Etherscan API response schema is documented anywhere, this page only shows a few examples: https://docs.etherscan.io/api-endpoints/accounts so I'm not sure whether there are more edge cases it may fail to parse etc.
- I haven't checked in detail whether every field is interpreted correctly -- for example, whether the nonce value is a plain decimal string or a hexadecimal string (and I think U256's default deserialization assumes the latter?)
- I'm not sure how common the edge case values are -- my sense is that they are not (at least for the `GENESIS` ones), so it may be unnecessary for many end use cases to worry about them.

Maybe a middle ground could be to keep the returned response types flexible as before (maybe they could be renamed as e.g. `RawNormalTransaction` to make it clear they are raw responses that weren't fully checked) + add more restricted cleaned up domain types with a `TryFrom` conversion from the raw types.
What do you think?

mattsse (commented):
Awesome!
can we also use Address, U256 etc in the new InternalTransaction and Erc20TokenTransferEven types?

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,692 @@
...
+    ///
+    /// # #[tokio::main]
+    /// # async fn main() {
+    ///     let client = Client::new(Chain::Mainnet, "API_KEY").unwrap();
+    ///     let txs = client
+    ///         .get_normal_transactions(&"0x58eB28A67731c570Ef827C365c89B5751F9E6b0a".parse().unwrap(),
+    ///         None).await.unwrap();
+    /// # }
+    /// ```
+    pub async fn get_normal_transactions(
```
mattsse, line 400:
```suggestion
    pub async fn get_transactions(
```

ethers-etherscan/src/account.rs
```diff
@@ -0,0 +1,692 @@
...
+    }
+}
+
+/// Possible values for some field responses
+#[derive(Debug)]
+pub enum GenesisOption<T> {
+    None,
+    Genesis,
+    Some(T),
+}
```
mattsse, line 74: we should add helper functions here to
`is_genesis`
`value(&self) -> Option<&T>`

mattsse:
> * I don't know if Etherscan API response schema is documented anywhere, this page only shows a few examples: https://docs.etherscan.io/api-endpoints/accounts so I'm not sure whether there are more edge cases it may fail to parse etc.
> * I haven't checked in detail whether every field is interpreted correctly -- for example, whether the nonce value is a plain decimal string or a hexadecimal string (and I think U256's default deserialization assumes the latter?)

that'll be discovered eventually so we can fix once that happens if there's a serde issue

> * I'm not sure how common the edge case values are -- my sense is that they are not (at least for the `GENESIS` ones), so it may be unnecessary for many end use cases to worry about them.

not familiar with the GENESIS values, but that sounds reasonable

tomtau:
> can we also use Address, U256 etc in the new InternalTransaction and Erc20TokenTransferEvent types?

There are now updated to be like that if that's the preferred way (as opposed to keeping them flexible and having conversions to domain types).

mattsse (approved):
lgtm

gakonst (approved):
Sick - thank you for bearing with our nits sir.

1wilkens:
> FYI I checked forks and noticed @1wilkens had a WIP version [1wilkens@f2eb537](https://github.com/1wilkens/ethers-rs/commit/f2eb53732fca388b0af21a756828e8efab8106d0) -- feel free to review if this PR works for you!

Sorry didn't even see your mention. Awesome that this is now merged, my branch was indeed early WIP so thanks for finishing this :)

PR #943, opened 2022-02-21
feat(solc): compile project components in parallel

mattsse:
## Motivation
WIP Compile components of a project in parallel.

Ref https://github.com/gakonst/foundry/issues/769 https://github.com/gakonst/foundry/issues/166

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst:
Closing this as unplanned. Turns out that we cannot really compile A.sol and B.sol in parallel and then link A.sol's bytecode in the places B.sol needs it, as it's always inlined. Open minded to approaches here that may work, as it'd speed things up significantly. Honestly this should be a part of `solc`.

PR #946, opened 2022-02-22
fix(solc): flatten import loc

rkrasiuk:
## Motivation

The import location detection on solang failure change introduced in [this commit](https://github.com/gakonst/ethers-rs/commit/331caf9418f38646f331a0ee22af4f5725aec77e) erroneously identified the import location within the file contents. 
i.e. 
`import "./Contract.sol";` would capture only `./Contract.sol` instead of the full import statement together with `import` keyword and trailing semicolon.

## Solution

Revert the change & add the test case that forces solang failure and ensures the correct flattening of the target file. 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


ethers-solc/tests/project.rs
```diff
@@ -426,12 +429,31 @@ fn can_flatten_file_with_duplicates() {
...
+
+    let target = root.join("contracts/Contract.sol");
+
+    let result = project.flatten(&target);
+    assert!(result.is_ok());
+
+    let result = result.unwrap();
+    assert!(!result.contains("import"));
+    assert_eq!(result.matches("contract Contract {").count(), 1);
+    assert_eq!(result.matches("library Lib {").count(), 1);
```
mattsse, line 454: can we compare against the whole string instead, maybe even for the other test cases?

mattsse (changes requested):

ethers-solc/tests/project.rs
rkrasiuk: yeah, we can do that

mattsse (approved):
lgtm

gakonst (approved):

PR #949, opened 2022-02-22
feat: cached provider PoC

gakonst:
# Problem

We want to avoid repeating requests which we expect to have the same response. This is useful when used together with Forge's mainnet forking tests.

# Solution

Add a `Cache` which stores all request/responses pairs in-memory, which then proceeds to store them in disk when it gets Dropped

### DIFF
--- a/ethers-providers/src/cache.rs
+++ b/ethers-providers/src/cache.rs
@@ -1,0 +1,98 @@
+use crate::ProviderError;
+use dashmap::DashMap;
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::{
+    fs::File,
+    io::{BufReader, BufWriter},
+    path::PathBuf,
+};
+
+#[derive(Clone, Debug, Default)]
+/// Simple in-memory K-V cache using concurrent dashmap which flushes
+/// its state to disk on `Drop`.
+pub struct Cache {
+    path: PathBuf,
+    // serialized request / response pair
+    requests: DashMap<String, String>,
+}
+
+// Helper type for (de)serialization
+#[derive(Serialize, Deserialize)]
+struct CachedRequest<'a, T> {
+    method: &'a str,
+    params: T,
+}
+
+impl Cache {
+    /// Instantiates a new cache at a file path.
+    pub fn new(path: PathBuf) -> Result<Self, ProviderError> {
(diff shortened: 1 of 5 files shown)


ethers-providers/src/cache.rs
```diff
@@ -0,0 +1,98 @@
...
+        let addr = Address::random();
+
+        assert!(provider.cache().unwrap().requests.is_empty());
+
+        mock.push(U256::from(100u64)).unwrap();
+        let res = provider.get_balance(addr, None).await.unwrap();
+        assert_eq!(res, 100.into());
+
+        assert!(!provider.cache().unwrap().requests.is_empty());
+        dbg!(&provider.cache());
```
mattsse, line 96:
```suggestion
```

ethers-providers/src/cache.rs
```diff
@@ -0,0 +1,98 @@
...
+        let key = serde_json::to_string(&CachedRequest { method, params })?;
+        let value = serde_json::to_string(&response)?;
+        self.requests.insert(key, value);
+        Ok(())
+    }
+}
+
+impl Drop for Cache {
+    fn drop(&mut self) {
+        let file = match File::options().write(true).read(true).create(true).open(&self.path) {
```
mattsse, line 61: this should be `File::create`? which

> Opens a file in write-only mode. reate a file if it does not exist, and will truncate it if it does.

ethers-providers/src/cache.rs
```diff
@@ -0,0 +1,98 @@
...
+    path::PathBuf,
+};
+
+#[derive(Clone, Debug, Default)]
+/// Simple in-memory K-V cache using concurrent dashmap which flushes
+/// its state to disk on `Drop`.
+pub struct Cache {
+    path: PathBuf,
+    // serialized request / response pair
+    requests: DashMap<String, String>,
```
mattsse, line 16: should this be <String, serde_json::Value>?

mattsse (changes requested):

gakonst:
@brockelmore curious what you think on this approach?

tynes:
Depending on how forge implements its caching strategy, it might be useful to have a way to clear the cache

gakonst:
Closing - will make it part of the Forking provider on Foundry

PR #963, opened 2022-02-24
handle case where the data from the resolver is empty

joshieDo:
## Motivation

Sometimes the resolver returns an empty Bytes.

In those cases we shouldn't go forward, otherwise `decode_bytes` will panic.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse (approved):
nice catch

gakonst (approved):

PR #968, opened 2022-02-26
feat(ethers-providers): expose pending tx_hash with a getter

meetmangukiya:
## Motivation

`ContractCall::send` returns `PendingTransaction`s, there is no way to access the tx hash since its a private field.

## Solution

Adding a getter so the user can get a contract call's tx hash.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
LGTM, FYI you can also get it by dereferencing the pending TX!

meetmangukiya:
Interesting, didn't notice the `DeRef` impl

PR #975, opened 2022-02-28
chore: remove sha256 proxy

### LINKED ISSUE #966: Suggestion: use PrehashSignature::Digest and remove Sha256Proxy
I think it is possible to remove the `Sha256Proxy` code in the signers crate because `recoverable::Signature` implements the `PrehashedSignature` trait which sets the `Digest` to `Keccak256`:

https://docs.rs/k256/latest/k256/ecdsa/recoverable/struct.Signature.html#impl-PrehashSignature

You would then not need to call `tx.sighash()` just `tx.rlp()` instead and let `k256` perform the hashing by calling `sign()` instead of `sign_digest()` then the `Sha256Proxy` code could be removed.

Of course there may be some subtle issue that prevents this I am not aware of 🙏

gakonst:
Fixes #966 by adding raw methods for calculating EIP712 and Eth Prefixed messages, then replaces the `sign_digest` calls with `sign`. Sha256 proxy is removed.

# TODO
Interestingly enough, it seems like this doesn't maintain the same transaction hash as before? See 
```

---- signer::tests::signs_tx stdout ----
thread 'signer::tests::signs_tx' panicked at 'assertion failed: `(left == right)`
  left: `[31, 128, 50, 153, 192, 99, 92, 234, 175, 189, 88, 103, 243, 102, 235, 197, 167, 200, 184, 252, 203, 6, 238, 72, 208, 230, 173, 163, 191, 161, 12, 22]`,
 right: `[222, 141, 185, 36, 136, 91, 8, 3, 210, 237, 195, 53, 247, 69, 178, 184, 117, 12, 136, 72, 116, 73, 5, 104, 76, 32, 185, 135, 68, 58, 149, 147]`', ethers-middleware/src/signer.rs:306:9
```

@tmpfs do you have any time to investigate this perhaps?


tmpfs:
@gakonst,

I had a quick look at this and when I parse the expected bytes for the transaction I get this:

```javascript
{
  nonce: 0,
  gasPrice: BigNumber { _hex: '0x04e3b29200', _isBigNumber: true },
  gasLimit: BigNumber { _hex: '0x1e8480', _isBigNumber: true },
  to: '0xF0109fC8DF283027b6285cc889F5aA624EaC1F55',
  value: BigNumber { _hex: '0x3b9aca00', _isBigNumber: true },
  data: '0x',
  chainId: 1,
  v: 37,
  r: '0xc9cf86333bcb065d140032ecaab5d9281bde80f21b9687b3e94161de42d51895',
  s: '0x727a108a0b8d101465414033c3f705a9c7b826e596766046ee1183dbc8aeaa68',
  from: '0x2c7536E3605D9C16a7a3D7b1898e529396a65c23',
  hash: '0xde8db924885b0803d2edc335f745b2b8750c8848744905684c20b987443a9593',
  type: null
}
```

When I parsed the bytes generated after your change I get this:

```javascript
{
  nonce: 0,
  gasPrice: BigNumber { _hex: '0x04e3b29200', _isBigNumber: true },
  gasLimit: BigNumber { _hex: '0x1e8480', _isBigNumber: true },
  to: '0xF0109fC8DF283027b6285cc889F5aA624EaC1F55',
  value: BigNumber { _hex: '0x3b9aca00', _isBigNumber: true },
  data: '0x',
  chainId: 1,
  v: 38,
  r: '0xac4282cde52f2a2518ccbaf7023be76223d8b15c68705ed7474487261082627b',
  s: '0x5770a8cc425296aef2987af5319edc0f3f9de092bfd51d794e1019359c0fd547',
  from: '0x2c7536E3605D9C16a7a3D7b1898e529396a65c23',
  hash: '0x1f803299c0635ceaafbd5867f366ebc5a7c8b8fccb06ee48d0e6ada3bfa10c16',
  type: null
}
```

So it looks to me like the recovery id has changed (which makes sense); given that the recovery id can change each time we sign a message using hard-coded values seems brittle to me. I think it would be more robust to decode the signed RLP bytes and assert on those values.

FYI, before I suggested this change I did remove `Sha256Proxy` on some derived code I am working on and tested it against a `ganache` network so I am reasonably confident it does work 🙏

gakonst:
Yeah I think this is not possible to do actually, the reason why we had done it is because we want to use SHA256 for calculating the ECDSA nonce in RFC6979. By removing the Sha256 proxy, we use Keccak256 for it, resulting in a different nonce -> different signature -> different hash.

tmpfs:
Thanks for clarifying and apologies for my lack of understanding!

PR #993, opened 2022-03-07
Yul Rust Bindings

ControlCplusControlV:
This is a draft so a lot is still left undone, and I only have a single test to show for it. Got some help from 0xKitsune on this one though.

It's not too many changes, just a lot of redundant functions. I considered making language a forced parameter of every compiler object but the downstream horror and added arg seemed like it would be too much. 

Still need linting but wanted your overall thoughts @gakonst now that new bindings are in a working state


gakonst:
@ControlCplusControlV have you looked at https://github.com/gakonst/ethers-rs/pull/994? It seems like the same is achieved with a simpler path, so maybe we should consolidate the work?

gakonst:
Closing as we consolidate on 994!

PR #994, opened 2022-03-07
feat(solc): yul compilation

alexeuler:
## Motivation

Currently Foundry and ethers-rs is not capable of compiling Yul sources (https://github.com/gakonst/foundry/issues/759)

## Solution

Add yul compilation


ethers-solc/src/artifacts/mod.rs
```diff
@@ -52,13 +52,37 @@ pub struct CompilerInput {
...
     /// Reads all contracts found under the path
-    pub fn new(path: impl AsRef<Path>) -> Result<Self, SolcIoError> {
+    pub fn new(path: impl AsRef<Path>) -> Result<Vec<Self>, SolcIoError> {
         Source::read_all_from(path.as_ref()).map(Self::with_sources)
     }
 
     /// Creates a new Compiler input with default settings and the given sources
-    pub fn with_sources(sources: Sources) -> Self {
-        Self { language: "Solidity".to_string(), sources, settings: Default::default() }
+    pub fn with_sources(sources: Sources) -> Vec<Self> {
```
mattsse, line 60: Since this returns at most `vec![sol, yul]`
can we add something like this perhaps

```rust
enum ProjectCompilerInput {
  Solidity(CompilerInput),
  Yul(..),
  Mixed(..,..)
}
```
?

mattsse (changes requested):
Thanks!
this looks good!
only nit is the `Vec<CompilerInput>` return type

ethers-solc/src/artifacts/mod.rs
alexeuler: I think the `ProjectCompilerInput` abstraction is correct right now - i.e. it's the unit that get's fed into solc compiler. The `with_sources` method though doesn't work because if sources have different types that means that you'll have 2 compiler inputs the needs to be fed into solc sequentially. So to me the signature looks good. Plus the vec is extensible for other sources (if any :))

To be 100% clean we could update the `new` method so it's just a plain constructor taking language as an input. For that method it's really uncommon to return a `Vec`. I just left it as it is for now because it could be used in many places.

Wdyt?

mattsse: makes sense, however working with a `Vec<CompilerInput>` can be a bit annoying.
Should be fine for now, but I think we can make this more ergonomic with `ProjectCompilerInput` which requires some refactoring, which we should leave for a follow up.

gakonst: I am OK with this as-is, seems good enough for now!

ControlCplusControlV:
> Yeah I was definitely thinking about adding abi for yul. My thinking was to load file with the same name as abi. E.g. if you have contracts/tokens/ERC20.yul then abi would be loaded from contracts/tokens/ERC20.abi.json. But that'd a good next step. 
> In any case I'm ok to merge this one or yours or some combined version. The key is to have a mix of Solidity / Yul sources in one project which will be kinda cool since hardhat and other guys don't have it :)

I can move over some of my changes here (mostly with the test) as suggestions and you can move them in as needed. 

The mix is definitely really cool as only Truffle can do it. What would you think of an approach where we have a way to cast a Solidity interface over the Yul file (so yul_file_name.abi.sol) and compile the interface, and put the ABI on top of the Yul artifact, then those who want to skip it with call() have the option to, while others can use it like nothing is different

ControlCplusControlV:
@gakonst yeah will make that PR later today as I am going to add in that change to handle Yul ABI alongside with it.

ControlCplusControlV:
My PR is up to @alexeuler's branch, still needs some work with one last type resolution issue which I can hopefully get resolved tomorrow as ABI should be fully functional from there!

ControlCplusControlV:
So type issues is fixed, but need help with one last thing. In this part of my fork of [artifact_output mod.rs](https://github.com/ControlCplusControlV/ethers-rs-1/blob/feature/yul_compilation/ethers-solc/src/artifact_output/mod.rs#L580) I have both the abi I need to replace and the artifact of which I need to insert the ABI into. But it doesn't appear there is a clean way to do that as Artifact won't let me modify the ABI directly and if I convert it into a CompactContract I am running into issues converting it back, so was curious as to the best way to do this or any other approaches. Will continue testing things but wanted to ask here in case there is something I am missing which could speed things up

mattsse:
I see, what modifications are necessary?
it seems like the proper way is to extend the `Artifact` trait some functions specifically for yul?

ControlCplusControlV:
The ABI needs to be accessible so that it can be replaced for a given `Artifact` or `ArtifactFile`. I will try extending artifact although there is a lot of unwrapping needed to get to the ABI and with all the types that implement Artifact I was hoping there would be another way which requires less refactoring

ControlCplusControlV:
@mattsse is there a specific way you envision extending `Artifact` ? As I need to replace the ABI but I haven't found a good way to access the ABI directly, as all the functions involve converting to a CompactContract to fetch values like ABI or Bytecode, but as there is no way to convert it back I haven't been able to get that approach to work. Is there a better way to directly access that value?

mattsse:
> @mattsse is there a specific way you envision extending `Artifact` ? As I need to replace the ABI but I haven't found a good way to access the ABI directly, as all the functions involve converting to a CompactContract to fetch values like ABI or Bytecode, but as there is no way to convert it back I haven't been able to get that approach to work. Is there a better way to directly access that value?

you mean the `ethers::abi::Abi` type? that type is baked into everything. 
What kind of modification/conversion do need need?

I don't have the full picture yet, and would need some more context on what needs to be done here https://github.com/ControlCplusControlV/ethers-rs-1/blob/feature/yul_compilation/ethers-solc/src/artifact_output/mod.rs#L580

ControlCplusControlV:
@mattsse I was actually able to resolve it by avoiding type conversions and just saving the Contract Object for each Yul contract, so that I could then just modify and replace it later. Just have one bug where it is compiling correctly but it seems the caching is overwriting the file contents instead of a cache, if you want to help I have pushed all my code to my fork and the test is fairly reliable producing it (It overwrites SimpleStore.yul with the cache of the artifact)

ControlCplusControlV:
@gakonst (and @alexeuler as it) could you review my changes for ABI to start moving it upstream? PR is [here](https://github.com/alexeuler/ethers-rs/pull/1) and my fork is [here](https://github.com/ControlCplusControlV/ethers-rs-1). So sorry for how long this took, did realize how bad my rust type skills are (although they've improved much more now!). I've added tests but as I was iterating so much code style is a bit sloppy so if there are any code style choices I can gladly implement them. 

It is a bit polluted with commits so squashing and merging will be much cleaner.

gakonst:
My candid recommendation would be that we merge this as-is, and @ControlCplusControlV can post a separate PR on top of master, given that it's not a blocker for this PR

PR #995, opened 2022-03-07
fix(solc): flatten nodes only once

mattsse:
## Motivation
https://github.com/gakonst/foundry/issues/811

## Solution
Moved the insert above the recursive function, which should guarantee that we include a node only once.

Which would prevent `A import B imports A` which would start recursive `flatten_node` at `B` but `A` is not yet included in the imports set so it will be included again.

CC @rkrasiuk 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
good fix

PR #996, opened 2022-03-08
Enforce commutativity of ENS reverse resolution

sbihel:
As specified in https://docs.ens.domains/dapp-developer-guide/resolving-names#reverse-resolution

## PR Checklist

- [ ] Added Tests
  * I'm not sure where to find a negative test.
- [x] Added Documentation
- [x] Updated the changelog


mattsse (approved):
lgtm

gakonst (approved):
I Guess this adds an extra RPC Call, but not much we can do about it. Massive footgun otherwise. LGTM

PR #1005, opened 2022-03-10
Fix websocket connection request

th4s:
In tungstenite 0.17 a change was introduced considering the logic of
`IntoClientRequest`. [Look here](https://github.com/snapview/tungstenite-rs/issues/270)

When using this trait on a `HttpRequest` it just
uses the request without any changes to initiate the websocket
connection. This makes it necessary for the consumer of the api to supply the
correct http headers for initiating a websocket connection when using
`IntoClientRequest` on `HttpRequest`.

This commit adapts the logic in `connect_with_auth` to this new
behavior.


mattsse (approved):
awesome!

gakonst (approved):

PR #1007, opened 2022-03-10
feat(ethers-compile): ethers-compile crate to refactor compilation

abigger87:
## Overview

This PR picks up on #936 .

## Motivation

Currently, compiler support is unstructured.

## Solution

Introduces a `Compiler` trait among other abstractions to create a canonical interface for implementing compiler support.

## PR Checklist

- [x] :gear: `Compiler` Trait
- [ ] :ship: Implement the `Compiler` trait for `ethers-solc`
- [ ] :test_tube: Testing

### DIFF
--- a/ethers-compile/src/ctrait.rs
+++ b/ethers-compile/src/ctrait.rs
@@ -1,0 +1,37 @@
+/// A Compiler-specific Result Type
+pub type Result<T, E> = std::result::Result<T, E>;
+
+/// # Compiler
+///
+/// A Generalized Compiler Trait
+/// Must be implemented to support specific compilation.
+///
+/// ### Example Implementation
+///
+/// ```no_run
+/// use ethers_compile::{Compiler};
+///
+/// pub struct Sanskrit {};
+///
+/// pub enum SEroor {
+///   IO,
+///   Compile,
+///   Unknown
+/// };
+///
+/// impl Compiler<u64, u64, SEroor> for Sanskrit {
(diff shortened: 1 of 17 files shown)


ethers-compile/src/ctrait.rs
```diff
@@ -0,0 +1,37 @@
...
+///
+/// let scompiler = Sanskrit {};
+/// let result = scompiler.compile(&input).unwrap();
+/// ```
+pub trait Compiler {
+  type Input;
+  type Output;
+  type Error;
+
+  fn compile(&self, input: &Self::Input) -> Result<Self::Output, Self::Error>;
```
mattsse, line 36: great, this is minimal interface we need

mattsse (changes requested):
so far so good.
now that we have a clean `ethers-compile` crate we can gradually move parts of ethers-solc into ethers-compile.

Ideally a `Project` will accept an instance `Compiler` and then start the pipeline with it.

So what we need as next step is to untangle the whole `ProjectCompiler` pipeline from `Solc` and make it `Compiler` agnostic instead.

there are some problems as all steps are currently tied to `Solc`, for example the dependency `Graph` only knows about solidity, which means resolving sources is also `Compiler` agnostic.

We want to reuse as much as possible from `self.preprocess()?.compile()?.write_artifacts()?.write_cache()`
in `ProjectCompiler::compile()`

I'd probably start with a `SolcCompiler` implementation then abstract everything and move it into the `Compiler` trait.

For example, resolving the `Graph` is language-specific and therefor needs to be part of the trait, etc...

PR #1012, opened 2022-03-12
fix: Export `ethers_providers::IpcError` and `ethers_providers::QuorumError`

oblique:
## Motivation

I was writing some wrappers and I noticed that `ethers_providers::IpcError` and `ethers_providers::QuorumError` were not exported.

## Solution

Export them in `pub use`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,13 @@
 
 ## ethers-providers
 
+### Unreleased
+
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

Issue #1014, opened 2022-03-13
Remove lifetime from `Event`


oblique:
**Is your feature request related to a problem? Please describe.**

I'm writing some high-level abstraction that combine multiple contracts in a single struct. In this abstraction I want to use `Event` and store it in an owned struct. However `Event` needs a lifetime because it borrows `self.client` from `Contract`.

**Describe the solution you'd like**

`self.client` of `Contract` is an Arc, so instead of borrowing we can clone it. This will remove lifetime from `Event` struct.

If this is acceptable from you, I can open a PR.

mattsse:
that sounds reasonable, please go for it!

gakonst:
SGTM! Doubt that we're at the point where the Arc vs Lifetime performance is the bottleneck, should always be the RPC so is fine

PR #1015, opened 2022-03-13
feat: Make `Event` an owned struct by storing `Arc<M>` instead of `&M`

oblique:
## Motivation

Motivation was described in #1014.

## Solution

`Event` now stores `Arc<M>` instead of `&M`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-contract/src/event.rs
+++ b/ethers-contract/src/event.rs
@@ -7,7 +7,7 @@ use ethers_core::{
     types::{BlockNumber, Filter, Log, ValueOrArray, H256},
 };
 use ethers_providers::{FilterWatcher, Middleware, PubsubClient, SubscriptionStream};
-use std::{borrow::Cow, marker::PhantomData};
+use std::{borrow::Cow, marker::PhantomData, sync::Arc};
 
 /// A trait for implementing event bindings
 pub trait EthEvent: Detokenize + Send + Sync {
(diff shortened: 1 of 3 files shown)


ethers-contract/src/event.rs
```diff
@@ -111,13 +111,13 @@ impl<M, D: EthLogDecode> Event<'_, M, D> {
...
 
-impl<'a, M, D> Event<'a, M, D>
+impl<M, D> Event<M, D>
 where
     M: Middleware,
     D: EthLogDecode,
 {
     /// Returns a stream for the event
-    pub async fn stream(
+    pub async fn stream<'a>(
```
mattsse, line 120: perhaps we can do the same for `SubscriptionStream`?

mattsse (approved):
lgtm

ethers-contract/src/event.rs
oblique: This is something I wanted too, but the changes were too many. I will give another try.

oblique: Ok so, to make `SubscriptionStream` an owned struct, the `Middleware` need some radical changes.

This is because `Middleware::Inner` (or `Middleware::Provider`) must be an `Arc`, otherwise we can not construct `SubscriptionStream` by calling `subscribe` method.

oblique: I prefer not to do it based on my comment above.

oblique:
After trying to remove lifetimes from `EventStream`/`FilterWatcher`/`SubscriptionStream` and failed, I started thinking that this PR will not give any benefits to the API. So you can close it if you want, unless if you have an idea to propose on refactoring `Middleware` trait.

In my use case I will take a different approach: I will have a `task::spawn` and a channel to get the results, so an owned struct will not be needed.

gakonst:
We had some discussion on a refactor of the middleware trait to make it Object Safe here: https://github.com/prestwich/ethers-middleware-experimental, but we never touched the lifetimes. 

What do you have in mind that we could/should change in the API of Middleware?

oblique:
I don't really have any clear thought, I was also experimenting. But if I end up to something I will open a discussion.

gakonst:
sgtm!

PR #1016, opened 2022-03-13
feat(provider): add RwClient

mattsse:
## Motivation
Add a Provider that uses a different client for sending tx.

## Solution
Wrap two clients but only the `write` client for ["eth_sendTransaction", "eth_sendRawTransaction"]

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


gakonst:
What would be an example use case of such a client? One would be Flashbots RPC + <Your Node>, anything else in mind?

mattsse:
> What would be an example use case of such a client? One would be Flashbots RPC + , anything else in mind?

yes, this is what I had in mind, not really elaborated, but maybe useful as a starting point

meetmangukiya:
Hi @mattsse, can we get this merged?

gakonst:
@meetmangukiya i'm supportive FWIW, do you also mind sharing your use case?

meetmangukiya:
My use case is also Flashbots RPC

mattsse:
rebased

gakonst (approved):

PR #1029, opened 2022-03-14
feat(solc): optimize output selection in cache mode

### LINKED ISSUE #1009: Optimize cache enabled recompile
with caching enabled only dirty files are provided to solc. in #998 this was changed so that the CompilerInput also included all the file's imports, so that solc does not need to resolve all required files itself. However this results in solc compiling all those clean files with the shared output settings, which is unnecessary as for clean files we don't require bytecode etc.

Take #998 approach and modify the settings dynamically, to tell solc not to include bytecode etc in the output for clean files.

mattsse:
## Motivation
Closes #1009 

## Solution
only select output for files that are actually dirty and select nothing for other files that are pulled in by dirty files.

Some preliminary testing on https://github.com/maple-labs/loan/releases/tag/v3.0.0-beta.0

cached and modified MapleLoan.t.sol:

with this PR:

```console
❯ ../../../foundry/target/release/forge build                                                                                                         ! tags/v3.0.0-beta.0 
[⠢] Compiling...
[⠑] Compiling 30 files with 0.8.7
Compiler run successful (with warnings)
../../../foundry/target/release/forge build  6.64s user 0.32s system 99% cpu 6.990 total
```
note the 30 files are 1 dirty + 29 imported. There's currently no easy way to fix this as the total number of source files in the CompilerInput is reported

current master

```console
❯ forge build                                                                                                                                              ! tags/v3.0.0-beta.0 
[⠢] Compiling...
[⠒] Compiling 1 files with 0.8.7
Compiler run successful (with warnings)
forge build  10.13s user 0.40s system 96% cpu 10.924 total

```

## Todo
- [x] additional tests
- [x] more caching docs

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


gakonst (commented):
so far so good - excited about this, will be even better on larger repos

ethers-solc/src/artifacts/output_selection.rs
```diff
@@ -1,7 +1,143 @@
...
+                ("".to_string(), vec!["*".to_string()]),
+            ]),
+        )])
+        .into()
+    }
+
+    /// Default output selection for compiler output:
+    ///
+    /// `{ "*": { "*": [ "*" ], "": [
+    /// "abi","evm.bytecode","evm.deployedBytecode","evm.methodIdentifiers"] } }`
```
gakonst, line 93: nit formatting here prob pretty print better

gakonst (approved):

PR #1034, opened 2022-03-15
refactor(solc): bump svm-rs and use returned install path

mattsse:
## Motivation
smv-rs now returns the path where a solc was installed.

## Solution
Use the path instead of the additional `find_installed_version` 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/resolver/mod.rs
+++ b/ethers-solc/src/resolver/mod.rs
@@ -627,23 +627,22 @@ impl VersionedSources {
 
         let mut sources_by_version = std::collections::BTreeMap::new();
         for (version, sources) in self.inner {
-            let mut solc = None;
-            if !version.is_installed() {
+            let solc = if !version.is_installed() {
                 if self.offline {
                     return Err(SolcError::msg(format!(
                         "missing solc \"{}\" installation in offline mode",
                         version
                     )))
                 } else {
-                    solc = Some(Solc::blocking_install(version.as_ref())?);
+                    // install missing solc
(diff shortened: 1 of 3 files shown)


ethers-solc/src/resolver/mod.rs
```diff
@@ -627,18 +627,22 @@ impl VersionedSources {
...
+            let mut solc = None;
             if !version.is_installed() {
                 if self.offline {
                     return Err(SolcError::msg(format!(
                         "missing solc \"{}\" installation in offline mode",
                         version
                     )))
                 } else {
-                    Solc::blocking_install(version.as_ref())?;
+                    solc = Some(Solc::blocking_install(version.as_ref())?);
```
roynalnaruto, line 638: Tiny point:

Why do we use an `Option<T>` here?

We could even do:
```suggestion
                    solc = Solc::blocking_install(version.as_ref());
```

and then:
```rust
let solc = solc.unwrap_or_else(|| {
    Solc::find_svm_installed_version(version.to_string())?.ok_or_else(|| {
        SolcError::msg(format!("solc \"{}\" should have been installed", version))
    })
})
```

roynalnaruto (commented):
Looks good! Just a tiny point

ethers-solc/src/resolver/mod.rs
mattsse: hmm this way we need another else branch, which is fine I guess, will add

gakonst (approved):

PR #1039, opened 2022-03-16
feat(solc): support customized output selection pruning

mattsse:
## Motivation
After #1029 we can extend this output pruning behavior and tell solc to only include output for certain files (test files)

## Solution
* introduce `FileFilter` trait that is a simple predicate. `TestFileFilter` is such a `FileFilter` that matches `.t.sol` files.
* add `Project::compile_sparse` to only compile those files that match a given predicate and prune all other files, this is intended to be used in `forge test`. not sure about the naming though.

By default, we prune all output from sources that are not _dirty_. the `FileFilter` will extend this to _dirty_ + `FileFilter::is_match`

May need a follow up for ergonomics for interacting with `forge`.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/lib.rs
+++ b/ethers-solc/src/lib.rs
@@ -26,8 +26,10 @@ pub mod remappings;
 use crate::artifacts::Source;
 
 pub mod error;
+mod filter;
 pub mod report;
 pub mod utils;
+pub use filter::{FileFilter, TestFileFilter};
 
 use crate::{
     artifacts::{Contract, Sources},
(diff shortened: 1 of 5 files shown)


ethers-solc/src/lib.rs
```diff
@@ -287,6 +289,48 @@ impl<T: ArtifactOutput> Project<T> {
...
+    /// # fn demo(project: Project) {
+    /// let project = Project::builder().build().unwrap();
+    /// let output = project
+    ///     .compile_sparse(
+    ///         |path: &Path| path.ends_with("Greeter.sol")
+    ///     ).unwrap();
+    /// # }
+    /// ```
+    #[cfg(all(feature = "svm", feature = "async"))]
+    pub fn compile_sparse<F: FileFilter + 'static>(
```
gakonst, line 323: Instead of `compile_sparse`, should we add a filter in the Project struct / in the builder? and then `compile()` would check if there's a filter and override the sources with sparse output?

mattsse: this would work, but has the disadvantage that the filter is then tied to the project so you'd need to modify the project if you need another filter, on recompile with a different filter for example.
I think this also won't add convenience to foundry because what we do is `config.project()`, if we add a filter to the project itself it will look like `config.project().with_filter(..).compile()` instead of `config.project().compile_sparse(..)`

gakonst (approved):
lgtm

PR #1040, opened 2022-03-16
fix(examples): adjust breaking changes and detect failures in ci

### LINKED ISSUE #1036: examples running forever
For some reason, Examples keep running without ending. here's an example workflow https://github.com/gakonst/ethers-rs/runs/5553524129?check_suite_focus=true. this is not reproduced locally. the command is basically "run all bins under the examples/ dir" https://github.com/gakonst/ethers-rs/blob/master/scripts/examples.sh. https://github.com/gakonst/ethers-rs/blob/master/.github/workflows/ci.yml#L178

mattsse:
## Motivation
Close #1036

## Solution
<s>apparently the step hang due to missing permissions `chmod +x ./scripts/examples` solved this</s>

Also added a separate build examples step to catch breaking changes, couple of our examples were not working

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -214,10 +214,17 @@ jobs:
       - uses: Swatinem/rust-cache@v1
         with:
           cache-on-failure: true
+      - name: Build all examples
+        run: |
+            export PATH=$HOME/bin:$PATH
+            for file in ./examples/*.rs; do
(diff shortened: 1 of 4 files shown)


gakonst (approved):

mattsse:
weird it seems to hang again,
I definitely got that working in a previous run...

PR #1041, opened 2022-03-16
fix(solc): use lowercase when comparing paths

mattsse:
## Motivation
https://github.com/odyslam/foundry-upgrades/tree/feat/core highlighted an issue in which errors went undetected, turns out solc not necessarily emits the exact file name, e.g. `src/utils/upgradeProxy.sol` is emitted as `src/utils/UpgradeProxy.sol` so we ended up swallowing some errors.

needs a followup with some tests.

## Solution
use lowercase

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -741,12 +741,16 @@ impl CompilerOutput {
     where
         I: IntoIterator<Item = &'a str>,
     {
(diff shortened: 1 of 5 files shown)


gakonst (approved):

mattsse:
test failures unrelated, version error will be addressed #1045

gakonst (approved):

PR #1047, opened 2022-03-17
feat(ethers-core/block): add gas_target and next_block_base_fee fns a…

meetmangukiya:
…ccording to EIP-1559

## Motivation

Target gas usage and next block's base fee can be useful helper methods.

## Solution

Implemented the same referring the EIP.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/types/block.rs
+++ b/ethers-core/src/types/block.rs
@@ -1,5 +1,7 @@
 // Taken from https://github.com/tomusdrw/rust-web3/blob/master/src/types/block.rs
 use crate::types::{Address, Bloom, Bytes, H256, U256, U64};
+#[cfg(not(feature = "celo"))]
+use core::cmp::Ordering;
 use serde::{ser::SerializeStruct, Deserialize, Deserializer, Serialize, Serializer};
 use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


ethers-core/src/types/block.rs
```diff
@@ -89,6 +89,49 @@ pub struct Block<TX> {
...
+
+    /// The next block's base fee, it is a function of parent block's base fee and gas usage.
+    /// Reference: https://eips.ethereum.org/EIPS/eip-1559
+    pub fn next_block_base_fee(&self) -> Option<U256> {
+        let target_usage = self.gas_target();
+        if let None = self.base_fee_per_gas {
+            return None
+        }
+
+        let base_fee_per_gas = self.base_fee_per_gas.unwrap();
```
mattsse, line 111: use `?` operator
```suggestion
        let base_fee_per_gas = self.base_fee_per_gas?;
```

ethers-core/src/types/block.rs
```diff
@@ -89,6 +89,49 @@ pub struct Block<TX> {
...
+
+        if self.gas_used > target_usage {
```
mattsse, line 113: this if elsif else can optionally be rewritten with a `match self.gas_used.cmp(&target_usage)`

mattsse (changes requested):
smol nit

ethers-core/src/types/block.rs
meetmangukiya: Is it supposed to be like:
```rs
match self.gas_used.cmp(&target_usage) {
    Ordering::Less => {...},
    Ordering::Greater => {...},
    Ordering::Equal => {...},
}
```

mattsse: yes!
usually, clippy complains if that's an if elseif else

gakonst (approved):

PR #1053, opened 2022-03-17
chore: decrease tx count in nonce manager test

mattsse:
## Motivation
decrease number of tx to send in nonce manager test, 10 -> 5 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
good call, should make the flakiness better and drain the rinkeby wallet slower

PR #1054, opened 2022-03-17
Create decode_tx_input.rs

elyase:
## Motivation
The question comes up often in the Telegram chat

## Solution
Provided an example

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/examples/decode_tx_input.rs
+++ b/examples/decode_tx_input.rs
@@ -1,0 +1,28 @@
+use ethers::abi::AbiDecode;
+use ethers::prelude::*;
+use eyre::Result;
+use hex;
+
+// Abigen creates a SwapExactTokensForTokensCall struct that can be used to decode
+// the call data for the swapExactTokensForTokens function in the IUniswapV2Router02 contract
+abigen!(
+    IUniswapV2Router02,
+    r#"[
+        swapExactTokensForTokens(uint256 amountIn, uint256 amountOutMin, address[] calldata path, address to, uint256 deadline)
+    ]"#,
+);
+fn main() -> Result<()> {
+    println!("Decoding https://etherscan.io/tx/0xd1b449d8b1552156957309bffb988924569de34fbf21b51e7af31070cc80fe9a");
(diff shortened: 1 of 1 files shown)


examples/decode_tx_input.rs
```diff
@@ -0,0 +1,28 @@
...
+    IUniswapV2Router02,
+    r#"[
+        swapExactTokensForTokens(uint256 amountIn, uint256 amountOutMin, address[] calldata path, address to, uint256 deadline)
+    ]"#,
+);
+fn main() -> Result<()> {
+    println!("Decoding https://etherscan.io/tx/0xd1b449d8b1552156957309bffb988924569de34fbf21b51e7af31070cc80fe9a");
+    let tx_input = "0x38ed173900000000000000000000000000000000000000000001a717cc0a3e4f84c00000000000000000000000000000000000000000000000000000000000000283568400000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000201f129111c60401630932d9f9811bd5b5fff34e000000000000000000000000000000000000000000000000000000006227723d000000000000000000000000000000000000000000000000000000000000000200000000000000000000000095ad61b0a150d79219dcf64e1e6cc01f0b64c4ce000000000000000000000000dac17f958d2ee523a2206206994597c13d831ec7";
+    let calldata = tx_input.strip_prefix("0x").unwrap_or(tx_input);
+    let calldata = Bytes::from(hex::decode(calldata).unwrap());
```
mattsse, line 18: I believe you can simply do
```suggestion
    let calldata: Bytes = input.parse().unwrap();
```
here

examples/decode_tx_input.rs
```diff
@@ -0,0 +1,28 @@
...
+    let decoded = SwapExactTokensForTokensCall::decode(&calldata)?;
```
mattsse, line 19: this makes me think wether we should add a `decode_hex` function?

cc @gakonst

mattsse (changes requested):
mind trying my suggestion, pretty sure this works

examples/decode_tx_input.rs
gakonst: Yeah `decode_hex(data: &str)` which does the conversion internally, is blanked impl'ed on the trait and forwards to `decode` seems right

examples/decode_tx_input.rs
elyase: nice! added the suggestion.

PR #1056, opened 2022-03-17
fix(core): more sophisticated crate detection

### LINKED ISSUE #1055: abigen mistakenly uses ethers_core instead of ethers
If you have ethers-core as a dependency and ethers w abigen feature as a dev-dependency, abigen thinks it should use the `ethers_x` packages instead of `ethers::`

```
error[E0433]: failed to resolve: use of undeclared crate or module `ethers_contract`
  --> tests/solidity.rs:74:1
   |
74 | abigen!(Groth16Verifier, "./tests/verifier_abi.json");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `ethers_contract`
```

mattsse:
## Motivation
Closes #1055 

## Solution
Improve crate detection, and make `ethers::` the fallback if `ethers-core`, `ethers-contract`, `ethers-providers` are not present

This also includes some checks in order to make it work in RA, because somehow RA ends up expanding abigen! while being in the `ethers-middleware` crate's directory...

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/macros/ethers_crate.rs
+++ b/ethers-core/src/macros/ethers_crate.rs
@@ -75,12 +75,6 @@ pub fn determine_ethers_crates() -> (&'static str, &'static str, &'static str) {
                     return sub_crates
                 }
 
-                // Note(mattsse): somehow rustanalyzer can end up expanding `abigen!` from inside the `ethers-middleware` crate: https://github.com/rust-analyzer/rust-analyzer/issues/11518
-                // so we need to bypass detection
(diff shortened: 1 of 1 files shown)


mattsse:
the RA fix breaks ethers-middleware I'm afraid...
needs some additional work

gakonst (approved):
LGTM - RA such a pain at this point to integrate with this :(

PR #1057, opened 2022-03-17
feat(solc): more artifact trait functions

mattsse:
## Motivation
Add additional `Artifacts` trait functions to access inner data

## Solution
use a `Cow` based type to allow borrowing, this will get rid of some (expensive) clones in forge

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst:
think there's some trait bounds not being met

gakonst (approved):
LGTM - just flagging again that we should prob consolidate these types, e.g. maybe now that we have the `Cow` type the `Ref` type doesn't make sense anymore? Do the `Compact` vs non-Compact types always make sense? etc

PR #1058, opened 2022-03-17
Examples/watch_block: show how to retrieve block metadata

nikkolasg:
## Motivation

I was searching how to get the timestamp and block number from this example and it turns out I need to make another call, so I'm showing this in the example so ppl can find it more easily.


mattsse (approved):
lgtm

gakonst (approved):

PR #1062, opened 2022-03-19
fix(core/TypedTransaction): eip1559 gas price should be max_fee_per_gas

meetmangukiya:
Gas price is `max_fee_per_gas`, not `max_fee_per_gas + max_priority_fee_per_gas`.


gakonst (approved):

gakonst:
Good catch. Thx!

PR #1065, opened 2022-03-19
chore(etherscan): Return error for unverified contracts

jubos:
Given contract source code is not useful in the case of unverified contracts
have the contract_source_code method return an error in the case if the item.abi
has the "Contract source code not verified" string.

Uncomment unit test ignore and run test:
`cargo test --lib ethers-etherscan -- get_error`


gakonst (approved):

gakonst:
Interesting - unit tests seem to hang?

jubos:
Looks like an unrelated test?

I set the new one to ignore since it hits etherscan API.

gakonst:
Yep -nice work, thank you again!

PR #1067, opened 2022-03-20
fix(core/Transaction): remove debug print

### LINKED ISSUE #1066: Extraneous debug print when RLP decoding legacy transactions
**Version**
2d75f9f1

**Platform**
All

**Description**
When RLP decoding legacy transactions `are we a list <true|false>` is printed to the console.

```rust
ethers::utils::rlp::decode::<Transaction>(some_rlp_legacy_tx);
// "are we a list" printed to console
```

jquesnelle:
Fixes #1066


gakonst (approved):
Oops good catch

PR #1075, opened 2022-03-23
Tx optional from

meetmangukiya:
## Motivation

Txs may not always have the `from` field set, make it optional and add method for being able to recover the same from signature and set it on the tx.

## Solution

Allow txs that do not have a `from` field by defaulting to `Address::zero()` instead of making it optional to not break compat. Add method to recover the sender from signature and also set the same on the tx.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-core/src/types/transaction/response.rs
+++ b/ethers-core/src/types/transaction/response.rs
@@ -1,5 +1,7 @@
 //! Transaction types
-use super::{decode_signature, eip2930::AccessList, normalize_v, rlp_opt};
+use super::{
+    decode_signature, eip2718::TypedTransaction, eip2930::AccessList, normalize_v, rlp_opt,
+};
 use crate::{
     types::{Address, Bloom, Bytes, Log, Signature, SignatureError, H256, U256, U64},
     utils::keccak256,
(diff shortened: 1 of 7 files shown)


ethers-core/src/types/transaction/response.rs
```diff
@@ -311,6 +312,11 @@ impl Transaction {
         *offset += 1;
         Ok(())
     }
+
+    pub fn recover_from(&self) -> Result<Address, SignatureError> {
+        let signature = Signature { r: self.r, s: self.s, v: self.v.as_u64() };
+        signature.recover(self.hash)
+    }
```
meetmangukiya, line 319: Not sure what's wrong here, recovers incorrectly

ethers-core/src/types/transaction/response.rs
```diff
@@ -636,4 +642,46 @@ mod tests {
...
+            v: 1.into(),
+            r: U256::from_str_radix(
+                "12010114865104992543118914714169554862963471200433926679648874237672573604889",
+                10,
+            )
+            .unwrap(),
+            s: U256::from_str_radix(
+                "22830728216401371437656932733690354795366167672037272747970692473382669718804",
+                10,
+            )
```
meetmangukiya, line 675: verified these are correct against https://goerli.etherscan.io/tx/0x5e2fc091e15119c97722e9b63d5d32b043d077d834f377b91f80d32872c78109

ethers-core/src/types/transaction/response.rs
meetmangukiya: Hash should be without the signature, thanks Dan!

gakonst (approved):

PR #1077, opened 2022-03-24
chore(source_tree): add .sol extension when missing

jubos:
Just ensure that any paths with no extension default to a .sol

Tests: cargo test --lib ethers-etherscan -- source_tree


mattsse (approved):
lgtm

PR #1080, opened 2022-03-27
feat: ability to get artifacts + sources

onbjerg:
## Motivation

We switched out `known_contracts` in Foundry to use `ArtifactId` as a key instead of just the contract name to resolve name collisions. In `forge run` we need both a `BTreeMap<ArtifactId, Artifact>` collection *and* the sources of the project (for source mapping).

## Solution

Add `ProjectCompileOutput::into_artifacts_with_sources`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse (approved):
lgtm

gakonst (approved):

PR #1081, opened 2022-03-27
feat(solc): include source file ast in artifact

### LINKED ISSUE #1028: Include source file AST when configured in ArtifactOutput
https://github.com/gakonst/foundry/issues/705#issuecomment-1066910698 

> The `ast` can only be set for the whole solc CompilerInput, it's enabled by default but we currently do not move the `ast`, which is stored separately in the CompilerOutput to the `Artifact`.
will add this shortly.

_Originally posted by @mattsse in https://github.com/gakonst/foundry/issues/705#issuecomment-1066910698_

mattsse:
## Motivation
Closes #1028

Adds the AST of the source file to the configurable artifact.

this is a bit tricky because ast is a per-file value and inserting this on a per contract/artifact basis is not really ideal.
So currently the `ast` field of an artifact will include the ast for the whole file.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifact_output/configurable.rs
+++ b/ethers-solc/src/artifact_output/configurable.rs
@@ -47,6 +47,8 @@ pub struct ConfigurableContractArtifact {
     pub ir_optimized: Option<String>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub ewasm: Option<Ewasm>,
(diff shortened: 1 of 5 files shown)


gakonst (approved):
Let's get this over the line and expose it On foundry so we can get the ball rolling for the slither folks

PR #1093, opened 2022-03-29
feat(core): add dev as a chain

Rjected:
## Motivation

Rather than refer to 1337 when using the default Ganache chain id for a provider, transaction, wallet, or anything else, it would be nice to use `Chain::Dev` instead.

## Solution

Add `Dev` to `Chain`

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


ethers-core/src/types/chain.rs
```diff
@@ -22,6 +22,7 @@ pub enum Chain {
     XDai = 100,
     Polygon = 137,
     Fantom = 250,
+    Ganache = 1337,
```
mattsse, line 25: this is also used by by geth and hardhat,
so I think we should make this `Dev` or something
we can't use Geth,Ganache, Hardhat because we have repr(u64)

mattsse (changes requested):

ethers-core/src/types/chain.rs
Rjected: Makes sense, changed to `Dev`

gakonst (approved):

PR #1095, opened 2022-03-30
Ensure a consistent chain ID between a signer and provider in SignerMiddleware

Rjected:
## Motivation

Frequently users run into errors returned by a `Provider` that are caused by chain ID inconsistencies between the user's `Signer` and `Provider`. For example, if the user tries to sign/send a transaction using a `SignerMiddleware` instantiated with a `Signer` and `Provider` that have different chain IDs, they might get this error from the provider:
```
Error: (code: -32700, message: Invalid signature v value, data: None)
```
Since users often use a `SignerMiddleware` to interact with both a provider and signer, it would be useful to ensure that the `Signer` uses the same chain ID as the `Provider`, to prevent this type of user error. [Other approaches were discussed in a different issue](https://github.com/gakonst/ethers-rs/pull/879#issuecomment-1032994226).

## Solution

Update the `SignerMiddleware` constructor and `with_signer` method to set the passed `Signer`'s chain ID to the ID retrieved from the provider.
This changes the following `SignerMiddleware` methods:

**Old**:
```rust
    pub fn new(inner: M, signer: S) -> Self {
        // constructor...
    }

    pub fn with_signer(&self, signer: S) -> Self {
        // implementation...
    }
```
**New**:
```rust
    pub async fn new(inner: M, signer: S) -> Result<Self, SignerMiddlewareError<M, S>> {
        // constructor...
        // pulls the chain id from the inner middleware
    }

    pub async fn with_signer(&self, signer: S) -> Result<Self, SignerMiddlewareError<M, S>> {
        // implementation...
    }
```
We also have removed the `#[must_use]` attribute from `with_signer`, since [it doesn't do anything for async functions](https://github.com/rust-lang/rust/issues/78149).

Importantly, if a `chain_id` cannot be retrieved from the provider (due to a provider error or some other reason), we bubble up the error.

Previously, the implementation details of the provider were not relevant to the creation of a `SignerMiddleware`.
Now, **the `SignerMiddleware` _must_ be constructed with a provider with an implemented and working `get_chainid` method**, or `new`/`with_signer` will return an error rather than a new `SignerMiddleware`.

Additionally, **this breaks current uses of `SignerMiddleware`**, since the `Result` must be handled, and each method is now `async`, since we're now interacting with a provider.

Tests are added which set a custom chain ID for Ganache and make sure the signer properly pulls the custom chain ID. To set a chain ID on Ganache, it had to be updated. There may be some differences in behavior between the two versions, so we should watch for test failures [after re-enabling the relevant section of CI](https://github.com/gakonst/ethers-rs/blob/3a653472cfb484bab7ae831ad6e3664d24b4e916/.github/workflows/ci.yml#L179).

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-middleware/src/signer.rs
+++ b/ethers-middleware/src/signer.rs
@@ -30,7 +30,7 @@ use thiserror::Error;
 ///     let wallet: LocalWallet = "380eb0f3d505f087e438eca80bc4df9a7faa24f868e69fc0440261a0fc0567dc"
 ///         .parse()?;
 ///
-///     let mut client = SignerMiddleware::new(provider, wallet).await;
+///     let mut client = SignerMiddleware::new(provider, wallet);
 ///
 ///     // You can sign messages with the key
 ///     let signed_msg = client.sign(b"hello".to_vec(), &client.address()).await?;
@@ -112,7 +112,7 @@ where
     ///
     /// [`Middleware`] ethers_providers::Middleware
     /// [`Signer`] ethers_signers::Signer
-    pub async fn new(inner: M, signer: S) -> Self {
+    pub fn new(inner: M, signer: S) -> Self {
         let address = signer.address();
         SignerMiddleware { inner, signer, address }
     }
@@ -182,7 +182,10 @@ where
     ///
     /// [`Middleware`] ethers_providers::Middleware
     /// [`Signer`] ethers_signers::Signer
-    pub async fn new_with_provider_chain(inner: M, signer: S) -> Result<Self, SignerMiddlewareError<M, S>> {
+    pub async fn new_with_provider_chain(
+        inner: M,
+        signer: S,
+    ) -> Result<Self, SignerMiddlewareError<M, S>> {
         let address = signer.address();
         let chain_id =
             inner.get_chainid().await.map_err(|e| SignerMiddlewareError::MiddlewareError(e))?;
@@ -369,7 +372,7 @@ mod tests {
             .parse::<LocalWallet>()
             .unwrap()
             .with_chain_id(chain_id);
-        let client = SignerMiddleware::new(provider, key).await;
+        let client = SignerMiddleware::new(provider, key);
 
         let tx = client.sign_transaction(tx).await.unwrap();
 
@@ -396,7 +399,7 @@ mod tests {
 
         // combine the provider and wallet and test that the chain id is the same for both the
         // signer returned by the middleware and through the middleware itself.
-        let client = SignerMiddleware::new(provider, key).await;
+        let client = SignerMiddleware::new(provider, key);
         let middleware_chainid = client.get_chainid().await.unwrap();
         assert_eq!(chain_id, middleware_chainid);
 
@@ -418,7 +421,7 @@ mod tests {
 
         // combine the provider and wallet and test that the chain id is the same for both the
         // signer returned by the middleware and through the middleware itself.
-        let client = SignerMiddleware::new(provider, key).await;
+        let client = SignerMiddleware::new(provider, key);
         let middleware_chainid = client.get_chainid().await.unwrap();
         assert_eq!(chain_id, middleware_chainid);
 
@@ -440,7 +443,7 @@ mod tests {
             )
             .await
             .unwrap();
-        let client = SignerMiddleware::new(provider, key).await;
+        let client = SignerMiddleware::new(provider, key);
 
         let request = TransactionRequest::new();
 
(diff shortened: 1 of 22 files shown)


Rjected:
Seems like the updated ganache behaves differently and [caused the run to fail](https://github.com/gakonst/ethers-rs/runs/5746590085?check_suite_focus=true), the test probably needs to be updated:
```
failures:

---- eth_tests::deploy_and_call_contract stdout ----
thread 'eth_tests::deploy_and_call_contract' panicked at 'assertion failed: `(left == right)`
  left: `56568`,
 right: `38856`', ethers-contract/tests/contract.rs:72:9


failures:
    eth_tests::deploy_and_call_contract

test result: FAILED. 34 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out; finished in 14.91s

error: test failed, to rerun pass '-p ethers-contract --test contract'
```

ethers-middleware/src/signer.rs
```diff
@@ -107,12 +107,24 @@ where
...
 {
     /// Creates a new client from the provider and signer.
-    pub fn new(inner: M, signer: S) -> Self {
+    /// Sets the address of this middleware to the address of the signer.
+    /// Sets the chain id of the signer to the chain id of the inner [`Middleware`] passed in,
+    /// using the [`Signer`]'s implementation of with_chain_id.
+    ///
+    /// [`Middleware`] ethers_providers::Middleware
+    /// [`Signer`] ethers_signers::Signer
+    pub async fn new(inner: M, signer: S) -> Result<Self, SignerMiddlewareError<M, S>> {
```
mattsse, line 116: this is definitely an improvement but this breaks, and will be redundant if the user manually sets the chain id without requesting it from the inner middleware.

I'd rather like to make this a separate asynchronous function that does this. basically with_signer

ethers-middleware/tests/signer.rs
```diff
@@ -36,7 +36,7 @@ async fn send_eth() {
         .interval(Duration::from_millis(10u64));
     let chain_id = provider.get_chainid().await.unwrap().as_u64();
     let wallet = wallet.with_chain_id(chain_id);
-    let provider = SignerMiddleware::new(provider, wallet);
+    let provider = SignerMiddleware::new(provider, wallet).await.unwrap();
```
mattsse, line 39: here for example, in this scenario, we're now querying the chain id twice

mattsse (commented):
cc @gakonst

gakonst:
I think moving to a new method instead of a breaking change makes sense here. Not sure what the best name for the function is, something that declares that it auto detects chain id ideally

Rjected:
Just switched to a new method `new_with_provider_chain`:
```rust
/// Creates a new client from the provider and signer.
/// Sets the address of this middleware to the address of the signer.
/// Sets the chain id of the signer to the chain id of the inner [`Middleware`] passed in,
/// using the [`Signer`]'s implementation of with_chain_id.
///
/// [`Middleware`] ethers_providers::Middleware
/// [`Signer`] ethers_signers::Signer
pub async fn new_with_provider_chain(
    inner: M,
    signer: S,
) -> Result<Self, SignerMiddlewareError<M, S>> {
    let address = signer.address();
    let chain_id =
        inner.get_chainid().await.map_err(|e| SignerMiddlewareError::MiddlewareError(e))?;
    let signer = signer.with_chain_id(chain_id.as_u64());
    Ok(SignerMiddleware { inner, signer, address })
}
```
Edited the comment on `new` to point to this method & explain what might happen if the chainid is not consistent.
Also needed to [remove an assert in `deploy_and_call_contract`](https://github.com/gakonst/ethers-rs/pull/1095/files#diff-f7ed5ca95548046c8f53e9acec04d1412a8e5db4cece5f3500be57c7a8067359L72) since the gas used & estimation are no longer equal with the updated ganache.

ethers-middleware/src/signer.rs
```diff
@@ -22,44 +22,44 @@ use thiserror::Error;
 /// use std::convert::TryFrom;
 ///
 /// # async fn foo() -> Result<(), Box<dyn std::error::Error>> {
-/// let provider = Provider::<Http>::try_from("http://localhost:8545")
-///     .expect("could not instantiate HTTP Provider");
+///     let provider = Provider::<Http>::try_from("http://localhost:8545")
```
gakonst, line 25: These indents shouldn't be required?

gakonst (changes requested):
Need to remove the formatting change in the doc comments, otherwise lgtm

ethers-middleware/src/signer.rs
Rjected: whoops yeah

Rjected:
Removed indents and rebased

gakonst (approved):
gg thx!

PR #1096, opened 2022-03-30
feat(core): implemented signed transaction RLP decoding

Rjected:
## Motivation

There are often cases where we might have some RLP encoding of a transaction, for example when receiving P2P messages, and want to decode that into a transaction type.

## Solution

This implements RLP decoding for signed transactions, extracting the signature from the decoding as well as building and returning the transaction. This also recovers the sender from the signature so we can populate the `from` field. This can error, so we introduce new error types (`Eip1559RequestError`, `TypedTransactionError`, `RequestError`) which also wrap errors we might encounter when decoding and recovering the sender.

This also introduces a `hash` method on `TypedTransaction` which returns the hash of the signed transaction's RLP encoding.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Pass compilation time as additional argument to `Reporter::on_solc_success` [1098](https://github.com/gakonst/ethers-rs/pull/1098)
 - Fix aws signer bug which maps un-normalized signature to error if no normalization occurs (in `aws::utils::decode_signature`)
+- Implement signed transaction RLP decoding [#1096](https://github.com/gakonst/ethers-rs/pull/1096)
 - `Transaction::from` will default to `Address::zero()`. Add `recover_from` and
   `recover_from_mut` methods for recovering the sender from signature, and also
   setting the same on tx [1075](https://github.com/gakonst/ethers-rs/pull/1075).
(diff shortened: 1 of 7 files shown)


mattsse (commented):
awesome!

before we merge this, mind having a look at this 

https://github.com/gakonst/foundry/blob/71f9e89256815008b5e4e7f8de84513c13110793/node/node-core/src/eth/transaction.rs

that's part of the forge node PR https://github.com/gakonst/foundry/pull/1037

whether we can somehow unify some things

gakonst:
@Rjected have you had the chance to look at Matt's code above? Are there maybe more utils to add?

gakonst:
(bump @rjected)

mattsse:
thanks @Rjected for the write up here https://github.com/gakonst/foundry/pull/1037#issuecomment-1086778075

agree with most of that, current block is that I need to convert from foundry node `TypedTransaction(Request)` to ethers Transaction(Request), in order to reuse ether's signers.

I think for now I'll try to do conversion in both ways and figure out how to unify afterwards, because I believe it could be replaced with ethers' types eventually.

so my recommendation, get this PR and the node PR over the line individually and unify and clean up afterwards.

gakonst (approved):
lgtm if @mattsse thinks this is OK to merge as-is

Rjected:
Cool, just fixed changelog, rebased

gakonst (approved):
based rebase, lgtm thank you for getting this over the line

PR #1104, opened 2022-04-03
feat(solc): add bytecode hash variants

mattsse:
## Motivation
Add concrete BytecodeHash Variants

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -472,6 +472,9 @@ impl From<BytecodeHash> for SettingsMetadata {
     }
 }
 
+/// Determines the hash method for the metadata hash that is appended to the bytecode.
+///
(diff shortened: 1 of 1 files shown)


ethers-solc/src/artifacts/mod.rs
```diff
@@ -457,8 +457,57 @@ pub struct SettingsMetadata {
...
+pub enum BytecodeHash {
+    Ipfs,
+    None,
+    Bzzr1,
+}
+
+impl Default for BytecodeHash {
+    fn default() -> Self {
+        // ipfs general default https://docs.soliditylang.org/en/latest/using-the-compiler.html#compiler-api
+        BytecodeHash::Ipfs
```
gakonst, line 485: Let's make it `None` and document that it's by default None?

mattsse: sg

PR #1108, opened 2022-04-04
feat(etherscan): add caching 

onbjerg:
Supersedes #935

### DIFF
--- a/ethers-etherscan/src/contract.rs
+++ b/ethers-etherscan/src/contract.rs
@@ -293,6 +293,9 @@ impl Client {
 
         let query = self.create_query("contract", "getabi", HashMap::from([("address", address)]));
         let resp: Response<String> = self.get_json(&query).await?;
+        if resp.result.starts_with("Max rate limit reached") {
+            return Err(EtherscanError::RateLimitExceeded)
+        }
         if resp.result.starts_with("Contract source code not verified") {
             return Err(EtherscanError::ContractCodeNotVerified(address))
         }
@@ -330,6 +333,9 @@ impl Client {
         let query =
             self.create_query("contract", "getsourcecode", HashMap::from([("address", address)]));
         let response: Response<Vec<Metadata>> = self.get_json(&query).await?;
+        if response.message == "NOTOK" {
+            return Err(EtherscanError::RateLimitExceeded)
+        }
         if response.result.iter().any(|item| item.abi == "Contract source code not verified") {
             return Err(EtherscanError::ContractCodeNotVerified(address))
         }
--- a/ethers-etherscan/src/contract.rs
+++ b/ethers-etherscan/src/contract.rs
@@ -284,6 +284,14 @@ impl Client {
     /// # }
     /// ```
     pub async fn contract_abi(&self, address: Address) -> Result<Abi> {
+        // apply caching
+        if let Some(ref cache) = self.cache {
+            let path = cache.join("abis").join(format!("{:?}.json", address));
+            let reader = std::io::BufReader::new(std::fs::File::create(path)?);
+            let abi = serde_json::from_reader(reader)?;
+            return Ok(abi)
+        }
+
         let query = self.create_query("contract", "getabi", HashMap::from([("address", address)]));
         let resp: Response<String> = self.get_json(&query).await?;
         if resp.result.starts_with("Contract source code not verified") {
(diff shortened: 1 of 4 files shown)


ethers-etherscan/src/lib.rs
```diff
@@ -59,8 +59,11 @@ impl Cache {
 
     fn set<T: Serialize>(&self, prefix: &str, address: Address, item: T) -> Result<()> {
         let path = self.0.join(prefix).join(format!("{:?}.json", address));
-        let writer = std::io::BufWriter::new(std::fs::File::create(path)?);
-        serde_json::to_writer(writer, &item)?;
+        let mut writer = std::io::BufWriter::new(std::fs::File::create(path)?);
+        serde_json::to_writer(&mut writer, &item)?;
+        // TODO: Trace
+        // TODO: Should we cache if the contract is *not* verified?
```
gakonst, line 65: I think not?

onbjerg:
This should work now

onbjerg:
Actually, I should probably add an actual error for rate limits so we can catch it in Foundry for retries

onbjerg:
This fails when the file doesn't exist, going back in draft

ethers-etherscan/src/lib.rs
gakonst: You could maybe add a TTL? Like only check the cache once a week, and (maybe) add a force flag to bypass the TTL?

gakonst (approved):

gakonst:
* need to change how the responses are parsed
* then add ttl

then should be good to merge

ethers-etherscan/src/contract.rs
```diff
@@ -284,12 +284,28 @@ impl Client {
...
+        // apply caching
+        if let Some(ref cache) = self.cache {
+            if let Ok(Some(abi)) = cache.get_abi(address) {
+                return Ok(abi)
+            }
+        }
+
         let query = self.create_query("contract", "getabi", HashMap::from([("address", address)]));
         let resp: Response<String> = self.get_json(&query).await?;
+        if resp.result.starts_with("Max rate limit reached") {
```
onbjerg, line 296: we have to dupe the check here cus it deserializes `message` as a string.. really annoying but no way around it really

ethers-etherscan/src/lib.rs
```diff
@@ -28,9 +37,91 @@ pub struct Client {
...
+        let mut writer = std::io::BufWriter::new(std::fs::File::create(path)?);
+        serde_json::to_writer(
+            &mut writer,
+            &CacheEnvelope {
+                expiry: SystemTime::now()
+                    .checked_add(self.ttl)
+                    .expect("cache ttl overflowed")
+                    .duration_since(UNIX_EPOCH)
+                    .expect("system time is before unix epoch")
+                    .as_secs(),
```
onbjerg, line 90: Is there a simpler way to do this?

ethers-etherscan/src/lib.rs
```diff
@@ -28,9 +37,91 @@ pub struct Client {
...
+        let _ = writer.flush();
+        Ok(())
+    }
+
+    fn get<T: DeserializeOwned>(&self, prefix: &str, address: Address) -> Result<Option<T>> {
+        let path = self.root.join(prefix).join(format!("{:?}.json", address));
+        let reader = std::io::BufReader::new(std::fs::File::open(path)?);
+        if let Ok(inner) = serde_json::from_reader::<_, CacheEnvelope<T>>(reader) {
+            // If this does not return None then we have passed the expiry
+            if SystemTime::now().checked_sub(Duration::from_secs(inner.expiry)).is_some() {
```
onbjerg, line 103: Same as above

onbjerg:
Oh wait, I re-requested review before I cached the "not verified" objects

ethers-etherscan/src/contract.rs
```diff
@@ -307,13 +331,34 @@ impl Client {
...
+        if let Some(ref cache) = self.cache {
+            // If this is None, then we have a cache miss
+            if let Some(src) = cache.get_source(address) {
+                // If this is None, then the contract is not verified
+                return match src {
+                    Some(src) => Ok(src),
+                    None => Err(EtherscanError::ContractCodeNotVerified(address)),
+                }
+            }
+        }
```
onbjerg, line 344: This is really ugly, but the alternative was to cache `Result<T, EtherscanError>`. Not sure. Any suggestions?

ethers-etherscan/src/lib.rs
```diff
@@ -28,9 +37,92 @@ pub struct Client {
...
+                        .expect("system time is before unix epoch")
+                        .as_secs(),
+                    data: item,
+                },
+            );
+            let _ = writer.flush();
+        }
+    }
+
+    fn get<T: DeserializeOwned>(&self, prefix: &str, address: Address) -> Option<T> {
```
onbjerg, line 99: Switched this to an option since we don't really care about the errors:

- If we can't deserialize it, it's malformed and we need to rewrite it
- If we can't open the file, the directory doesn't exist (we probably care here, should we add tracing?)

ethers-etherscan/src/contract.rs
gakonst: you can prob do something like `if let Some(Some(src)) = cache.get_source(address)` but nbd. feel free to do in follow up

gakonst (approved):

gakonst:
Merging - let's hope it works well :)

PR #1123, opened 2022-04-08
Fix IPC handling of jsonrpc errors

### LINKED ISSUE #1119: json-rpc error responses are not handled correctly
**Version**
ethers-providers v0.6.2

**Platform**
Linux 5.17.1-arch1-1

**Description**
The `ipc` transport does not handle json-rpc errors in a meaningful way.

https://github.com/gakonst/ethers-rs/blob/6e43d46f5753bcdaf2dead2c2d599d13e43338d1/ethers-providers/src/transports/ipc.rs#L272-L288

In line 276, the received response is converted into a `serde_json::Value` regardless of whether it is a `result` or `error` and sent to the oneshot receiver future at the callsite as is. At the callsite, it is tried to be deserialized into the expected type, e.g. a `H256` when attempting to submit a transaction. This fails, if the tx was, e.g., submitted with too low a nonce, because the rpc error (json) object does not deserialize to a `H256` and you get a very unhelpful straight `serde-json` error with zero additional context.

It would be more sensible, to send a `Result<Value, JsonRpcError>` back to the callsite and handle it appropriately.

**Update**: Checked again, `ws` appears to do it right, by calling `into_result`, not sure why but `ipc` uses `into_value`, which returns `Ok` for both response kinds, unless it is not valid json.

BTW: the entire internal handling of jsonrpc requests looks pretty inefficient as it uses `serde_json::Value` everywhere (basically at least one hashmap), where it could easily be replaced with `serde_json::RawValue`, which is just a string. Basically, the current flow is

```
concrete type -> serde_json::Value -> io-writer
io-reader -> serde_json::Value -> concrete type
```
which constructs a ton of intermediate hashmaps for no real purpose

oliver-giersch:
## Motivation

Fixes #1119: JSON-RPC error responses are not propagated to (request) callsites for the IPC transport correctly and end up as unhelpful `serde_json` errors.

Tangentially, the internal handling of all JSON values creates a lot of intermediate `serde_json::Value` instances, which is inefficient, since this creates potentially nested `HashMap`s.

## Solution

Instead of calling `Response::to_value`, which converts both success and error responses into a `serde_json::Value` we call `Response::into_result`, which propagates error responses as `Err` variants.

Internal handling of JSON responses and requests is switched from `serde_json::Value` to `serde_json::value::RawValue`, which does not create any hashmaps.

## Not Solved

A potential issue persists, in that internal IPC server errors (such as an unexpected disconnect of the IPC socket, e.g. a node crash) are **not** propagated to pending subscriptions, for other pending requests, this should be correctly propagated now, but for subscriptions this would require a change to the `PubSubClient` trait definition (i.e., sending `Result` notifications).

### DIFF
--- a/ethers-providers/Cargo.toml
+++ b/ethers-providers/Cargo.toml
@@ -20,7 +20,7 @@ async-trait = { version = "0.1.50", default-features = false }
 hex = { version = "0.4.3", default-features = false, features = ["std"] }
 reqwest = { version = "0.11.10", default-features = false, features = ["json"] }
 serde = { version = "1.0.124", default-features = false, features = ["derive"] }
-serde_json = { version = "1.0.64", default-features = false }
+serde_json = { version = "1.0.64", default-features = false, features = ["raw_value"] }
 thiserror = { version = "1.0.30", default-features = false }
 url = { version = "2.2.2", default-features = false }
 auto_impl = { version = "0.5.0", default-features = false }
--- a/ethers-providers/src/pubsub.rs
+++ b/ethers-providers/src/pubsub.rs
@@ -5,7 +5,7 @@ use ethers_core::types::{TxHash, U256};
 use futures_util::stream::Stream;
 use pin_project::{pin_project, pinned_drop};
 use serde::de::DeserializeOwned;
-use serde_json::Value;
+use serde_json::value::RawValue;
 use std::{
     marker::PhantomData,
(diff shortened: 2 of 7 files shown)


mattsse (approved):
this looks great,
propagating error responses as Err variants is more useable

gakonst (approved):
Nice work!

PR #1124, opened 2022-04-08
fix: fix deploy tx RLP decoding

xJonathanLEI:
The transaction RLP decoding logic assumes `to` is present but it's actually not true for contract deployment transactions. The library fails to decode such transactions as a result.

This PR fixes it.


mattsse (approved):

xJonathanLEI:
Fixed formatting. Didn't realize I should use nightly tool chain for that lol.

gakonst (changes requested):
Good catch - could you refactor the logic to a helper function and call that?

gakonst (approved):
nice! thank you.

PR #1126, opened 2022-04-08
feat(solc): Add standard-json-input

pyk:
## Motivation

This changes will allows `forge verify-contract` to submit contract verification using `standard-input-json` format to etherscan.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-etherscan/src/contract.rs
+++ b/ethers-etherscan/src/contract.rs
@@ -133,7 +133,7 @@ impl AsRef<str> for CodeFormat {
 
 impl Default for CodeFormat {
     fn default() -> Self {
-        CodeFormat::SingleFile
+        CodeFormat::StandardJsonInput
     }
 }
 
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -285,7 +285,7 @@ impl Default for Settings {
         Self {
             stop_after: None,
             optimizer: Default::default(),
-            metadata: Default::default(),
+            metadata: None,
             output_selection: OutputSelection::default_output_selection(),
             evm_version: Some(EvmVersion::default()),
             via_ir: None,
@@ -485,13 +485,6 @@ pub struct SettingsMetadata {
 impl From<BytecodeHash> for SettingsMetadata {
     fn from(hash: BytecodeHash) -> Self {
         Self { use_literal_content: None, bytecode_hash: Some(hash) }
-    }
-}
-
-impl Default for SettingsMetadata {
-    fn default() -> Self {
-        println!("ethers-solc artifact mod settings metadata default");
-        Self { use_literal_content: Some(true), bytecode_hash: Default::default() }
     }
 }
 
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -285,7 +285,7 @@ impl Default for Settings {
         Self {
             stop_after: None,
             optimizer: Default::default(),
-            metadata: None,
+            metadata: Default::default(),
             output_selection: OutputSelection::default_output_selection(),
             evm_version: Some(EvmVersion::default()),
             via_ir: None,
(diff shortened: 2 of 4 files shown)


ethers-solc/src/lib.rs
```diff
@@ -426,6 +426,36 @@ impl<T: ArtifactOutput> Project<T> {
...
+
+    /// Returns standard-json-input to compile the target contract
+    pub fn standard_json_input(&self, target: &Path) -> Result<String> {
+        tracing::trace!("Building standard-json-input");
+        let graph = Graph::resolve(&self.paths)?;
+        let target_index = graph.files().get(target).ok_or_else(|| {
+            SolcError::msg(format!("cannot resolve file at \"{:?}\"", target.display()))
+        })?;
+        let mut sources = Vec::new();
+        let (path, source) = graph.node(*target_index).unpack();
```
mattsse, line 438: should probably work
```suggestion
        let (path, source) = graph.node(*target_index).unpack().clone();
```

ethers-solc/src/artifacts/mod.rs
```diff
@@ -488,6 +488,13 @@ impl From<BytecodeHash> for SettingsMetadata {
     }
 }
 
+impl Default for SettingsMetadata {
+    fn default() -> Self {
+        println!("ethers-solc artifact mod settings metadata default");
```
mattsse, line 493:
```suggestion
```

ethers-solc/src/artifacts/mod.rs
```diff
@@ -285,7 +285,7 @@ impl Default for Settings {
         Self {
             stop_after: None,
             optimizer: Default::default(),
-            metadata: None,
+            metadata: Default::default(),
```
mattsse, line 288: this will deviate from `solc`'s defaults (useLiteralContent)

don't think this matters but interested wdyt?

ethers-etherscan/src/contract.rs
```diff
@@ -133,7 +133,7 @@ impl AsRef<str> for CodeFormat {
 
 impl Default for CodeFormat {
     fn default() -> Self {
-        CodeFormat::SingleFile
+        CodeFormat::StandardJsonInput
```
mattsse, line 136: good call

ethers-solc/src/lib.rs
```diff
@@ -426,6 +426,36 @@ impl<T: ArtifactOutput> Project<T> {
     pub fn flatten(&self, target: &Path) -> Result<String> {
         self.paths.flatten(target)
     }
+
+    /// Returns standard-json-input to compile the target contract
+    pub fn standard_json_input(&self, target: &Path) -> Result<String> {
```
mattsse, line 431: I'd rather return a the `CompilerInput` here instead

ethers-solc/src/lib.rs
```diff
@@ -426,6 +426,36 @@ impl<T: ArtifactOutput> Project<T> {
...
+        let target_index = graph.files().get(target).ok_or_else(|| {
+            SolcError::msg(format!("cannot resolve file at \"{:?}\"", target.display()))
+        })?;
+        let mut sources = Vec::new();
+        let (path, source) = graph.node(*target_index).unpack();
+        sources.push((path.clone(), source.clone()));
+        for index in graph.all_imported_nodes(*target_index) {
+            let (path, source) = graph.node(index).unpack();
+            sources.push((path.clone(), source.clone()));
+        }
```
mattsse, line 443: probably also works
```suggestion
      sources.extend(graph.all_imported_nodes(*target_index).map(|index|graph.node(index).unpack()).cloned())
```

mattsse (changes requested):
thanks for this,
few smol nits and suggestions

ethers-solc/src/lib.rs
mattsse: or even `collect::<Vec<_>>()`

ethers-solc/src/artifacts/mod.rs
pyk: @mattsse I agree, this changes is unnecessary

ethers-solc/src/lib.rs
pyk: @mattsse I make it return string consistent with `flatten`. wdyt?

https://github.com/gakonst/foundry/blob/908c0c8ac47d9ff4ab0494a00fc2896be410e47b/cli/src/cmd/forge/verify.rs#L103-L123

mattsse: still, I think `CompilerInput` would make more sense here and could be useful to for example use it directly with `Solc::compile`

pyk: ok let me update it

pyk: @mattsse updated 👍

gakonst (approved):

PR #1130, opened 2022-04-09
fixes: correct etherscan url address, remove double quotes in solc error

prestwich:
Two simple fixes:

1. Etherscan address generation was using Address display impl, instead of debug impl. This caused incorrect URL generation with address middles snipped e.g. `https://kovan.etherscan.io/address/0xe03d…98db#code`
2. Solc error had escaped double quotes around a string formatted with Debug impl. This caused `""double double quotes""` in that message. 

## Motivation
Fix some bad forge UX :)

## Solution

- Use Debug impls for Address
- make etherscan Tx properly use `H256` instead of string
- make etherscan token properly use `Address` instead of string
- remove the extra quotes

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-etherscan/src/lib.rs
+++ b/ethers-etherscan/src/lib.rs
@@ -14,7 +14,7 @@ use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use errors::EtherscanError;
 use ethers_core::{
     abi::{Abi, Address},
-    types::Chain,
+    types::{Chain, H256},
 };
 
 pub mod account;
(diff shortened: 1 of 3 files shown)


ethers-etherscan/src/lib.rs
```diff
@@ -247,17 +247,17 @@ impl Client {
 
     /// Return the URL for the given address
     pub fn address_url(&self, address: Address) -> String {
-        format!("{}address/{}", self.etherscan_url, address)
+        format!("{}address/{:?}", self.etherscan_url, address)
```
mattsse, line 250: Incredible how often the formatting of the address had to be fixed -.-

mattsse (approved):

prestwich:
okay at this point I don't believe the CI errors are my fault anymore :)

ethers-etherscan/src/lib.rs
prestwich: yeh, I dunno whose idea it was to make `Display` snip the middle 😬

gakonst (approved):

PR #1140, opened 2022-04-13
feat(ethers-core/abi): impl AbiArrayType for u8

meetmangukiya:
## Motivation

Had a compilation error earlier in abigen:
```
1198 |         pub pairs_to_swap: ::std::vec::Vec<(u8, u8)>,
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `AbiArrayType` is not implemented for `u8`
```
So implementing it.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/abi/mod.rs
+++ b/ethers-core/src/abi/mod.rs
@@ -100,12 +100,7 @@ impl<T: AbiArrayType, const N: usize> AbiType for [T; N] {
 
 impl<T: AbiArrayType, const N: usize> AbiArrayType for [T; N] {}
 
-impl<const N: usize> AbiType for [u8; N] {
-    fn param_type() -> ParamType {
-        ParamType::FixedBytes(N)
-    }
-}
-impl<const N: usize> AbiArrayType for [u8; N] {}
+impl AbiArrayType for u8 {}
 
 macro_rules! impl_abi_type {
     ($($name:ty => $var:ident $(($value:expr))? ),*) => {
@@ -123,7 +118,6 @@ macro_rules! impl_abi_type {
 
 impl_abi_type!(
     Bytes => Bytes,
-    Vec<u8> =>  Array(Box::new(ParamType::Uint(8))),
     Address => Address,
     bool => Bool,
     String => String,


meetmangukiya:
I am not entirely sure if this is breaking anything or any compat. I defer to experts.

gakonst (changes requested):
Yeah this is not the right solution here, because now it won't encode fixed length arrays..need to think about this a bit more.

mattsse:
`AbiArrayType ` is such an ugly hack ...

but the right way to fix this is to implement this manually for tuples of u8

impl AbiArrayType for Vec<(u8, u8)>

impl AbiArrayType for Vec<(u8, u8, u8)> 

etc...

meetmangukiya:
@mattsse wouldn't it still fail when there's some tuple like `Vec<(u8, T, ...)>`?

mattsse:
yes this is a limitation and super ugly.

we could generate all those variants, I believe there is already a macro for tuples

meetmangukiya:
Shouldn't
https://github.com/gakonst/ethers-rs/blob/6ff567517e5a6d061cd3802adeec6055c68c1977/ethers-core/src/abi/mod.rs#L159-L162 implement `AbiArrayType` for different variant of tuples and on top of that https://github.com/gakonst/ethers-rs/blob/6ff567517e5a6d061cd3802adeec6055c68c1977/ethers-core/src/abi/mod.rs#L93 implement `AbiArrayType` for `Vec<T>`? 

which should mean we shouldn't need any of these?
```rs
impl AbiArrayType for Vec<(u8, u8)> {}

impl AbiArrayType for Vec<(u8, u8, u8)> {}
```

meetmangukiya:
Pasting complete error for more context:
```
error[E0277]: the trait bound `u8: AbiArrayType` is not satisfied
    --> mean-finance/src/contracts/dcahub_companion.rs:1198:9
     |
1198 |         pub pairs_to_swap: ::std::vec::Vec<(u8, u8)>,
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `AbiArrayType` is not implemented for `u8`
     |
     = help: the following implementations were found:
               <i128 as AbiArrayType>
               <i16 as AbiArrayType>
               <i32 as AbiArrayType>
               <i64 as AbiArrayType>
             and 5 others
     = note: required because of the requirements on the impl of `AbiArrayType` for `(u8, u8)`
     = note: required because of the requirements on the impl of `AbiType` for `Vec<(u8, u8)>`
```

I couldn't find this exact trait bound(`u8: AbiArrayType`) causing this error.

https://github.com/gakonst/ethers-rs/blob/a43a9b8806a13035db5c657406b3542dd0bfc749/ethers-core/src/abi/mod.rs#L88-L92
would require for `T` i.e. `(u8, u8)` to be trait bounded by `AbiArrayType`.

which it should be because of this implementation https://github.com/gakonst/ethers-rs/blob/a43a9b8806a13035db5c657406b3542dd0bfc749/ethers-core/src/abi/mod.rs#L165-L168

now, this implementation in turn bounds the tuple element type `T` to be trait bound by `AbiType`. However, the error seems to be that `T` needs to be bound by `AbiArrayType`. Where is that trait bound coming from?

ethers-core/src/abi/mod.rs
```diff
@@ -100,12 +100,7 @@ impl<T: AbiArrayType, const N: usize> AbiType for [T; N] {
 
 impl<T: AbiArrayType, const N: usize> AbiArrayType for [T; N] {}
 
-impl<const N: usize> AbiType for [u8; N] {
-    fn param_type() -> ParamType {
-        ParamType::FixedBytes(N)
-    }
-}
-impl<const N: usize> AbiArrayType for [u8; N] {}
+impl AbiArrayType for u8 {}
```
mattsse, line 103: this defeats the whole purpose of this trait.

mattsse (changes requested):
this trait is required because of the special `Vec<u8> => Bytes` condition so it's sole purpose is to not be implemented by u8

need some manual impls for u8 tuples unfortunately

ethers-core/src/abi/mod.rs
meetmangukiya: The currently pushed code is certainly wrong. Can you check the comment https://github.com/gakonst/ethers-rs/pull/1140#issuecomment-1098345741 regarding the manual impls for u8 tuples? And for the boader issue and source of error https://github.com/gakonst/ethers-rs/pull/1140#issuecomment-1098363286

mattsse:
> Shouldn't
> 
> https://github.com/gakonst/ethers-rs/blob/6ff567517e5a6d061cd3802adeec6055c68c1977/ethers-core/src/abi/mod.rs#L159-L162
> 
> implement `AbiArrayType` for different variant of tuples and on top of that
> https://github.com/gakonst/ethers-rs/blob/6ff567517e5a6d061cd3802adeec6055c68c1977/ethers-core/src/abi/mod.rs#L93
> 
> implement `AbiArrayType` for `Vec<T>`?
> which should mean we shouldn't need any of these?
> 
> ```rust
> impl AbiArrayType for Vec<(u8, u8)> {}
> 
> impl AbiArrayType for Vec<(u8, u8, u8)> {}
> ```

interesting, need to check this manually what's going on here

mattsse:
@meetmangukiya you're of course right: #1143

can you share the whole struct and its derive macros that's causing the issue, perhaps we're generating wrong trait bounds somewhere

meetmangukiya:
```rs
    #[derive(
        Clone,
        Debug,
        Default,
        Eq,
        PartialEq,
        ethers :: contract :: EthCall,
        ethers :: contract :: EthDisplay,
    )]
    #[ethcall(
        name = "getNextSwapInfo",
        abi = "getNextSwapInfo(address[],(uint8,uint8)[])"
    )]
    pub struct GetNextSwapInfoCall {
        pub tokens: ::std::vec::Vec<ethers::core::types::Address>,
        pub pairs: ::std::vec::Vec<(u8, u8)>,
    }
```

```rs
    #[derive(
        Clone,
        Debug,
        Default,
        Eq,
        PartialEq,
        ethers :: contract :: EthCall,
        ethers :: contract :: EthDisplay,
    )]
    #[ethcall(
        name = "swap",
        abi = "swap(address[],(uint8,uint8)[],address,address,uint256[],bytes)"
    )]
    pub struct SwapCall {
        pub tokens: ::std::vec::Vec<ethers::core::types::Address>,
        pub pairs_to_swap: ::std::vec::Vec<(u8, u8)>,
        pub reward_recipient: ethers::core::types::Address,
        pub callback_handler: ethers::core::types::Address,
        pub borrow: ::std::vec::Vec<ethers::core::types::U256>,
        pub data: ethers::core::types::Bytes,
    }
```

meetmangukiya:
Just rebased my abigen binary over latest ethers-rs and the generated code compiles fine 💀. I apologise for the wasted time @mattsse.

New code looks like
```rs
    #[derive(
        Clone,
        Debug,
        Default,
        Eq,
        PartialEq,
        ethers :: contract :: EthCall,
        ethers :: contract :: EthDisplay,
    )]
    #[ethcall(
        name = "getNextSwapInfo",
        abi = "getNextSwapInfo(address[],(uint8,uint8)[])"
    )]
    pub struct GetNextSwapInfoCall {
        pub tokens: ::std::vec::Vec<ethers::core::types::Address>,
        pub pairs: ::std::vec::Vec<PairIndexes>,
    }
```

mattsse:
great!
just added this to confirm: #1144

PR #1146, opened 2022-04-14
fix(timelag): normalize block if block = None

ltchang2019:
## Motivation

Timelag not lagging behind for normal state queries (calls), only for block indexing.
- Ethers-contract implementation provides [block = None when building a ContractCall](https://github.com/gakonst/ethers-rs/blob/master/ethers-contract/src/contract.rs#L241)
-[ Contract.call() passes tx and block id (set to None) to provider](https://github.com/gakonst/ethers-rs/blob/master/ethers-contract/src/call.rs#L151-L152)
- Provider passes down the stack of inners
- Timelag middleware applies lag by calling normalize_block_number [BUT it is only called if the provided block is Some (call —> normalize_block_id —> normalize_block_number IFF id is Some)](https://github.com/gakonst/ethers-rs/blob/master/ethers-middleware/src/timelag/mod.rs#L60)
- That block field set by the contract was None so the timelag middleware never applies the lag

## Solution

Apply timelag if block = None.

## PR Checklist

- [ ] Added Tests (N/A)
- [ ] Added Documentation (N/A)
- [x] Updated the changelog

### DIFF
--- a/ethers-middleware/src/timelag/mod.rs
+++ b/ethers-middleware/src/timelag/mod.rs
@@ -58,7 +58,7 @@ where
                 Ok(self.normalize_block_number(Some(n)).await?.map(Into::into))
             }
             None => Ok(self.normalize_block_number(None).await?.map(Into::into)),
-            _ => Ok(id)
+            _ => Ok(id),
         }
     }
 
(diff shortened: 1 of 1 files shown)


prestwich (approved):
changes look good to me

gakonst (approved):

mattsse (approved):

PR #1148, opened 2022-04-15
fix: support display for bytes

mattsse:
## Motivation
support the `Bytes` type in `EthDisplay` and format using hex encode

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


onbjerg (approved):
lgtm (fixes the display of bytes in logs in foundry)

PR #1151, opened 2022-04-16
feat(solc): Use IndexedPathBuf for CompilerInput Sources

pyk:
## Motivation

Based on https://github.com/foundry-rs/foundry/issues/1283 it seems that Etherscan is not automatically detect main contract to be displayed as the first source on the list. 

After learning how `hardhat-etherscan` [verification works](https://gist.github.com/pyk/7eff86450abb92f86c3e712f741d1802) turns out that hardhat always place the main contract as the first key of the `sources`.

This behaviour is currently not possible in `ethers-solc` due to CompilerInput sources is using `BTreeMap<PathBuf, Source>`. When compiler input is [serialized as JSON](https://github.com/foundry-rs/foundry/blob/b6451a5c501475892453ee4aadd681de78d7b6ec/cli/src/cmd/forge/verify.rs#L340), it always ordered alphabetically (`PathBuf` default), not by their insertion order.

## Solution

This changes propose `Sources` to be defined as `BTreeMap<IndexedPathBuf, Source>` instead of `BTreeMap<PathBuf, Source>`. `IndexedPathBuf` contains `index` field which we can use to determine how the key is sorted by just simply assign the `index` value.

In the case of verification, the `sources` key will be ordered by source code for main contract first then followed by the imported source codes.

Results:
- [Contract Verified without IndexedPathBuf](https://rinkeby.etherscan.io/address/0xcc544ecdc8b0f00fe8e4e7f796032c3919d3d83f#code)
- [Contract Verified with IndexedPathBuf](https://rinkeby.etherscan.io/address/0x356a7bb15a898e7265f377a71825d21072202123#code)

(`SimpleERC20` is on top of the list and followed by the imports)

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -56,12 +56,10 @@ impl Ord for IndexedPathBuf {
     fn cmp(&self, other: &Self) -> Ordering {
         if self.path == other.path {
             Ordering::Equal // PathBuf deduplication
+        } else if self.index >= other.index {
+            Ordering::Greater
         } else {
-            if self.index >= other.index {
-                Ordering::Greater
-            } else {
-                Ordering::Less
-            }
+            Ordering::Less
         }
     }
 }
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -57,7 +57,11 @@ impl Ord for IndexedPathBuf {
         if self.path == other.path {
             Ordering::Equal // PathBuf deduplication
         } else {
-            self.index.cmp(&other.index)
+            if self.index >= other.index {
+                Ordering::Greater
+            } else {
+                Ordering::Less
+            }
         }
     }
 }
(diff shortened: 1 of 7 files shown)


mattsse (commented):
it seems that this is primarily for verification?

Are there any other advantages, because this touches a lot of internal code and indexedpath is only needed for verification?

I've been thinking a bit about how to implement this, but haven't gotten around to writing it down yet.

What if we make the `standard_json` function return a new type that acts like a `CompilerInput` but seriealises with sorted sources? basically your IndexedPath apporach

the index would probably come from the graph, `imported_nodes` I think?

gakonst:
Yeah I'd say let's try to make it sort only when serializing in a non invasive way if possible!

pyk:
@gakonst ok, noted 🙇

pyk:
replaced by #1169 , thanks @mattsse sorry I haven't finished this yet. i will pickup other issues this weekend 🙇

Issue #1155, opened 2022-04-19
ethers-contract: ContractFactory, set deployer to public so defaults may be overrided.


inconspicuous99:
**Is your feature request related to a problem? Please describe.**
Using abigen!() releases a contract whose deploy method returns ContractDeployer. When deploying ownable contracts you want to set who you're sending it from, especially if your provider has multiple accounts. 

**Describe the solution you'd like**
Set ContractDeployer's deployer field to public.

```rust
pub struct ContractDeployer<M, C> {
    /// the actual deployer, exposed for overriding the defaults
    pub deployer: Deployer<M>, // change this


    /// marker for the `Contract` type to create afterwards
    ///
    /// this type will be used to construct it via `From::from(Contract)`
    _contract: PhantomData<C>,
}
```
```rust
  abigen!(OwnableContract, "$CARGO_MANIFEST_DIR/artifacts/OwnableContract.sol/OwnableContract.json");

  let deployer = OwnableContract::deploy(
    Arc::new(provider.clone()),
    "".to_string(),
  ).expect("Failed to create deployer");
  
  let desired_from_address = provider.getAccounts().await?[0];
  deployer.deployer.tx.from = desired_from_address;
  
  deployer.send().await.expect("failed to send");
```

**Describe alternatives you've considered**
You could create a SingleSigner middleware and deploy it with it, but you could also expose deployer as public and privatize Deployer's private fields (already being done) and modify the public transaction. (and it's already public for situations like these, was just an oversight not to set ContractDeployer's deployer to public).

**Additional context**
![image](https://user-images.githubusercontent.com/83731823/163961845-a4b3fed4-df72-4d1b-b293-97dc07d3b9b7.png)

inconspicuous99:
solved by #1156

PR #1156, opened 2022-04-19
ethers-contract: ContractFactory, set deployer to public so defaults may be overrided. Solves  #1155

inconspicuous99:
solves #1155


gakonst (approved):
good call

PR #1157, opened 2022-04-19
feat: geth abigen to generate go bindings

tynes:
## Motivation

We maintain offchain services in Go and are using forge for contract compilation/testing.
It would be great to be able to generate Go bindings using geth's `abigen` from
foundry's artifacts automatically, similarly to how `forge bind` creates rust bindings.
We currently use a [Makefile](https://github.com/ethereum-optimism/optimistic-specs/blob/23a337b162ed02706aefc54724c48e686b8d09c1/l2os/Makefile#L26) which needs to be updated every time a new binding is needed.

## Solution

This draft PR is a temperature check to see if the approach is at the right abstraction. `foundry`
would be opinionated in generating the file structures themselves such that it should be possible
to just import the Go code

I don't love having two structs `ContractBindings` and `GoContractBindings` which are very similar

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


roynalnaruto (commented):
Minor suggestions from my side :)

roynalnaruto:
@tynes You will need to run:
```
$ cargo +nightly fmt --all
$ cargo +nightly clippy --all
```

tynes:
Closing this for now, will reopen when I have time to fix it up

PR #1165, opened 2022-04-21
feat: function debug data in extra output

onbjerg:
## Motivation

https://github.com/foundry-rs/foundry/issues/1381

## Solution

Add it

Also, it seems like the AST is always present and is not configurable - it's not present in `OutputSelection` and is enabled by default.

`ExtraOutputValues::compact_format` also seems to be unused?


mattsse (approved):
lgtm

gakonst (approved):

PR #1174, opened 2022-04-24
refactors ipc transport internals

oliver-giersch:
ran cargo +nightly fmt

fixes typo

remove some commented out code

## Motivation

The current IPC server code has two issues:

https://github.com/gakonst/ethers-rs/blob/77bd9d49c88ca8bdced37f7554919e3c3d584310/ethers-providers/src/transports/ipc.rs#L163-L186

This is the main IPC server loop (the function is called in a loop). Since the IPC `UnixStream` is split into read & write halves, I assume the intention was to allow reads & writes to happen concurrently, but due to the await in line 169 this is broken: The `await` on the write action (the first arm of the select expression) prevents any reads to be processed while the write is incomplete.

The second issue is in the `handle_request` method itself:

https://github.com/gakonst/ethers-rs/blob/77bd9d49c88ca8bdced37f7554919e3c3d584310/ethers-providers/src/transports/ipc.rs#L195-L199

Here, a call to `tokio::io::AsyncWriteExt::write` is made, which does **not** guarantee all bytes from the given `buf` are actually written (the fix is basically to call `write_all`). 

## Solution

Since the code for the IPC server was fairly convoluted and the error handling in many cases rather questionable (hard errors logged as warnings, normal events logged as errors, etc) I opted to basically rewrite the IPC server part.

Instead of spawning a tokio task, I instead to chose to spawn a regular thread with a single-threaded tokio runtime on it (avoids synchronization overhead, otherwise the hash maps of pending requests and subscriptions would need a mutex) and replaced the `select!` with a simpler `try_join!`, which does allow fully concurrent reads & writes.
I also replaced the manual buffer management with a more straight-forward use of the `bytes` crate (which was already used before, just not using its main advantages).

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1321,6 +1321,7 @@ dependencies = [
  "futures-core",
  "futures-timer",
  "futures-util",
+ "hashers",
  "hex",
  "http",
  "once_cell",
@@ -1635,6 +1636,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "fxhash"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c31b6d751ae2c7f11320402d34e41349dd1016f8d5d45e48c4312bc8625af50c"
+dependencies = [
+ "byteorder",
+]
+
+[[package]]
 name = "generic-array"
 version = "0.12.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


mattsse (approved):
thanks,

looks so much better.
have no way of testing this though, but lgtm

oliver-giersch:
> thanks,
> 
> so looks so much better. have no way of testing this though, but lgtm

FWIW, I had this new version running for a ~week 24/7 (in production, if you will) without any issues

gakonst (approved):
LGTM - thanks for this, IPC has needed more love for a while

PR #1175, opened 2022-04-25
feat: blocknative gas oracle

georgewhewell:
## Motivation
we don't have any 1559 gas oracle

## Solution
adds blocknative gas oracle

notes-
 - this provider requires api key (although it's free) so did not include tests. if we can add one as CI env variable i can update PR with tests
 - the response returns some potentially-useful fields that aren't exposed in the `GasOracle` trait- i left the response structs and `request` method public so users can use these themselves if needed

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -237,6 +237,8 @@
 
 - Ensure a consistent chain ID between a Signer and Provider in SignerMiddleware
   [#1095](https://gakonst/ethers-rs/pull/1095)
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

roynalnaruto (approved):
Looks good to me!

gakonst (approved):

PR #1177, opened 2022-04-25
fix(solc): only write cache file if build was successful

mattsse:
## Motivation
Applies the same checks used to determine whether to write artifacts (successful build, no ephemeral) to the cache file.

Likely related https://github.com/foundry-rs/foundry/issues/1244

## Solution
only write cache file if build was successful

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/project_util/mod.rs
+++ b/ethers-solc/src/project_util/mod.rs
@@ -34,8 +34,10 @@ pub struct TempProject<T: ArtifactOutput = ConfigurableArtifacts> {
 impl<T: ArtifactOutput> TempProject<T> {
     /// Makes sure all resources are created
     pub fn create_new(root: TempDir, inner: Project<T>) -> std::result::Result<Self, SolcIoError> {
-        let project = Self { _root: root, inner };
+        let mut project = Self { _root: root, inner };
         project.paths().create_all()?;
+        // ignore license warnings
+        project.inner.ignored_error_codes.push(1878);
         Ok(project)
     }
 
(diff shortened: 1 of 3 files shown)


ethers-solc/src/project_util/mod.rs
```diff
@@ -34,8 +34,10 @@ pub struct TempProject<T: ArtifactOutput = ConfigurableArtifacts> {
 impl<T: ArtifactOutput> TempProject<T> {
     /// Makes sure all resources are created
     pub fn create_new(root: TempDir, inner: Project<T>) -> std::result::Result<Self, SolcIoError> {
-        let project = Self { _root: root, inner };
+        let mut project = Self { _root: root, inner };
         project.paths().create_all()?;
+        // ignore license warnings
+        project.inner.ignored_error_codes.push(1878);
```
roynalnaruto, line 40: Should we have an `enum` for error codes and then push the appropriate variant here?

roynalnaruto (commented):
Looks good otherwise, minor point.

ethers-solc/src/project_util/mod.rs
mattsse: we have this in foundry config, since this here is only used for testing, think it's fine.
but should think about moving the error code enum to ethers-solc instead

roynalnaruto (approved):

gakonst (approved):

PR #1180, opened 2022-04-26
fix(core): decode `from` field for typed transactions

Rjected:
## Motivation

The `from` field for all non-legacy signed transactions was incorrect after decoding from bytes.

## Solution

 * Properly `sighash` the transactions by using the `TypedTransaction` implementation. Sighashes were improperly calculated due to the lack of a transaction type prefix before hashing.
 * Remove the `Eip1559TransactionRequest` `sighash` method.
 * Add tests for decoding real signed transactions (EIP1559 and EIP2930) and recovering the `from` field.
 * Add test for an EIP2930 transaction with populated access lists.
 * Fixed `chain_id` decoding for EIP2930 transactions.
 * Introduce `Eip2930RequestError` for covering either recovery or rlp decoding errors.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog


roynalnaruto:
@Rjected If I'm not wrong, the `RLP` encoding and `sighash` for typed transactions has already been done here:
1. https://github.com/gakonst/ethers-rs/blob/master/ethers-core/src/types/transaction/eip2718.rs#L257-L274
2. https://github.com/gakonst/ethers-rs/blob/master/ethers-core/src/types/transaction/eip2718.rs#L276-L280

In my opinion, the `sighash` implementation on `Eip1559TransactionRequest` struct [here](https://github.com/gakonst/ethers-rs/blob/master/ethers-core/src/types/transaction/eip1559.rs#L160-L163) should be removed as it's misleading (and of course incorrect). But for all other purposes, one can use the `TypedTransaction` struct and the implementations, and that shouldn't be a problem?

Rjected:
That's a good idea, just pushed a new version that uses the `TypedTransaction` sighash in the decoding, and removed the `TX_TYPE` constants. Also removed the `Eip1559TransactionRequest` sighash method, I prefer the `TypedTransaction` implementation.

gakonst (approved):

Issue #1176, opened 2022-04-25
ethers-solc: compiling USDT on v0.4.18 gives out error


joshieDo:
Trying to build [USDT](https://etherscan.io/address/0xdac17f958d2ee523a2206206994597c13d831ec7#code) on foundry with `solc v0.4.18`  results in the following crash that can be traced over to the json line below. 

the output is a valid one (acquired on the traces), it's just cant be deserialized for some reason

`ethers-solc/src/compile/mod.rs`
```rust
    /// Run `solc --stand-json` and return the `solc`'s output as the given json
    /// output
    pub fn compile_as<T: Serialize, D: DeserializeOwned>(&self, input: &T) -> Result<D> {
        let output = self.compile_output(input)?;
        Ok(serde_json::from_slice(&output)?)
    }
```
error:
```shell
➜  testdir git:(master) ✗ forge build --use 0.4.18
[⠊] Compiling...
[⠆] Compiling 1 files with 0.4.18
Error: 
   0: EOF while parsing a value at line 1 column 0

Location:
   cli/src/compile.rs:99
```

mattsse:
thanks for this
could you try dumping the json in/output via `ETHERS_SOLC_LOG=in=in.json,out=out.json` and perhaps try parsing them separately from a formatted json, that should highlight the serde issue I think

mattsse:
vote close @gakonst

PR #1182, opened 2022-04-26
fix(solc): when compiler-out metadata is empty and there's no `internalType`

joshieDo:
## Motivation
ref #1176 

## Solution
* handle when `metadata` is an empty string
* handle when there is no `internalType`

Also, it was not giving out an error but I specified  the field on `userdoc.methods.method`, since it seems to only have one field `notice`.

[v0.5.8](https://docs.soliditylang.org/en/v0.5.8/natspec-format.html#user-documentation)
[v0.8.13](https://docs.soliditylang.org/en/v0.8.13/natspec-format.html#user-documentation)

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog


mattsse (approved):
thanks!
lgtm

gakonst (approved):

PR #1183, opened 2022-04-26
feat: Relax `Clone` requirements when `Arc<M>` is used

oblique:
`#[derive(Clone)]` was implementing `Clone` only if `M` was `Clone`, however because of `Arc<M>` this can be relaxed.

## Motivation

I was implementing a `Signer` and I was getting a misleading error because it wasn't clonable.
However since the whole middleware was wrapped in an `Arc`, then `Clone` shouldn't be a requirement.

## Solution

When `#[derive(Clone)]` is used on a generic struct, then `Clone` is implemented if generic parameters are also `Clone`.
In the case of a middleware we can avoid this because it is always wrapped in an `Arc`.

A solution is to manually implement `Clone` for structs that take `Middleware` as a generic parameter.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,8 @@
 
 ### Unreleased
 
+- Relax Clone requirements when Arc<Middleware> is used
+  [#1183](https://github.com/gakonst/ethers-rs/pull/1183)
 - Generate a deploy function if bytecode is provided in the abigen! input (json artifact)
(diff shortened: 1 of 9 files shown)


mattsse (approved):
thanks!
faced the selective clone derive myself recently

gakonst (approved):
Ah makes sense, we get the clone from the src, no need for it to be on the generic

PR #1190, opened 2022-04-28
fix(solc): remapping aware libraries

mattsse:
## Motivation
The solc docs say about libraries

```json
"libraries": {
      // The top level key is the the name of the source file where the library is used.
      // If remappings are used, this source file should match the global path
      // after remappings were applied.
      // If this key is an empty string, that refers to a global level.
      "myFile.sol": {
        "MyLib": "0x123123..."
      }
    },
```

currently we're not able to apply remappings, this adds support for this

Ref https://github.com/foundry-rs/foundry/issues/1338

## Solution
* add `Libraries` wrapper type with support for applying remappings
* add more tests

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -425,7 +425,7 @@ impl Libraries {
                 SolcError::msg(format!("failed to parse library name: {}", lib))
             })?;
             let addr = items.next().ok_or_else(|| {
-                SolcError::msg(format!("failed to parse invalid library: {}", lib))
+                SolcError::msg(format!("failed to parse library address: {}", lib))
             })?;
             if items.next().is_some() {
                 return Err(SolcError::msg(format!("failed to parse, too many arguments passed: {}", lib)))
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -16,6 +16,7 @@ use crate::{
 };
 
 use serde::{de::Visitor, Deserialize, Deserializer, Serialize, Serializer};
+use tracing::warn;
 
 pub mod ast;
 pub use ast::*;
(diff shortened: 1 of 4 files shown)


ethers-solc/src/artifacts/mod.rs
```diff
@@ -381,6 +391,98 @@ impl Default for Settings {
...
+    ///     "src/DssSpell.sol:DssExecLib:0xfD88CeE74f7D78697775aBDAE53f9Da1559728E4".to_string(),
+    /// ])
+    /// .unwrap();
+    /// ```
+    pub fn parse(libs: &[String]) -> Result<Self, SolcError> {
+        let mut libraries = BTreeMap::default();
+        for lib in libs {
+            let mut items = lib.split(':');
+            let file = items.next().ok_or_else(|| {
+                SolcError::msg(format!("failed to parse invalid library: {}", lib))
```
gakonst, line 422:
```suggestion
                SolcError::msg(format!("failed to parse path to library file: {}", lib))
```

ethers-solc/src/artifacts/mod.rs
```diff
@@ -381,6 +391,98 @@ impl Default for Settings {
...
+            })?;
+            let lib = items.next().ok_or_else(|| {
+                SolcError::msg(format!("failed to parse invalid library: {}", lib))
```
gakonst, line 425:
```suggestion
                SolcError::msg(format!("failed to parse library name: {}", lib))
```

ethers-solc/src/artifacts/mod.rs
```diff
@@ -381,6 +391,98 @@ impl Default for Settings {
...
+            })?;
+            let addr = items.next().ok_or_else(|| {
+                SolcError::msg(format!("failed to parse invalid library: {}", lib))
```
gakonst, line 428:
```suggestion
                SolcError::msg(format!("failed to parse library address: {}", lib))
```

ethers-solc/src/artifacts/mod.rs
```diff
@@ -381,6 +391,98 @@ impl Default for Settings {
...
+            })?;
+            if items.next().is_some() {
+                return Err(SolcError::msg(format!("failed to parse invalid library: {}", lib)))
```
gakonst, line 431:
```suggestion
                return Err(SolcError::msg(format!("failed to parse, too many arguments passed: {}", lib)))
```

gakonst (approved):
nits - lgtm otherwise

PR #1192, opened 2022-04-28
fix(solc): flatten import aliases

rkrasiuk:
## Motivation

addresses [foundry#1440](https://github.com/foundry-rs/foundry/issues/1440)

## Solution

This PR includes: 
- More robust regex for import statement parsing in case of `solang` failure
- Partial support for import aliases: any explicit import alias will be replaced with its original name 

Known Issues: 
- Since alias replacement rewrites the contents of the file, any previous data unit location might be incorrect. This is a temporary solution and is based on the assumption that any usage of an alias happens after all of the import statements. 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/resolver/parse.rs
+++ b/ethers-solc/src/resolver/parse.rs
@@ -14,7 +14,7 @@ pub struct SolData {
     pub license: Option<SolDataUnit<String>>,
     pub version: Option<SolDataUnit<String>>,
     pub experimental: Option<SolDataUnit<String>>,
-    pub imports: Vec<SolDataUnit<PathBuf>>,
+    pub imports: Vec<SolDataUnit<SolImport>>,
     pub version_req: Option<VersionReq>,
     pub libraries: Vec<SolLibrary>,
     pub contracts: Vec<SolContract>,
@@ -39,7 +39,7 @@ impl SolData {
     pub fn parse(content: &str, file: &Path) -> Self {
         let mut version = None;
         let mut experimental = None;
-        let mut imports = Vec::<SolDataUnit<PathBuf>>::new();
+        let mut imports = Vec::<SolDataUnit<SolImport>>::new();
         let mut libraries = Vec::new();
         let mut contracts = Vec::new();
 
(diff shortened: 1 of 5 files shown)


rkrasiuk:
@mattsse @gakonst this PR is ready for review. updated the description as well ✔️

ethers-solc/src/utils.rs
```diff
@@ -35,6 +40,19 @@ pub static RE_SOL_SDPX_LICENSE_IDENTIFIER: Lazy<Regex> =
 /// A regex used to remove extra lines in flatenned files
 pub static RE_THREE_OR_MORE_NEWLINES: Lazy<Regex> = Lazy::new(|| Regex::new("\n{3,}").unwrap());
 
+/// Create a regex that matches any library or contract name inside a file
+pub fn create_contract_or_lib_name_regex(name: &str) -> Regex {
+    regex::Regex::new(&format!(r#"(?:using\s+(?P<n1>{name})\s+|is\s+(?:\w+\s*,\s*)*(?P<n2>{name})(?:\s*,\s*\w+)*|(?:(?P<ignore>(?:function|error|as)\s+|\n[^\n]*(?:"([^"\n]|\\")*|'([^'\n]|\\')*))|\W+)(?P<n3>{name})(?:\.|\(| ))"#, name = name)).unwrap()
+}
```
mattsse, line 46: let's make this also a Lazy?

ethers-solc/src/resolver/parse.rs
```diff
@@ -164,57 +203,41 @@ impl SolLibrary {
...
+
     /// Returns the underlying data for the unit
     pub fn data(&self) -> &T {
         &self.data
     }
 
     /// Returns the location of the given data unit
-    pub fn loc(&self) -> (usize, usize) {
-        (self.loc.start, self.loc.end)
+    pub fn loc(&self) -> &Range<usize> {
```
mattsse, line 232: should we clone/copy here?

mattsse (approved):
awesome!
tests looking good!

ethers-solc/src/utils.rs
```diff
@@ -41,9 +41,11 @@ pub static RE_SOL_SDPX_LICENSE_IDENTIFIER: Lazy<Regex> =
...
 
 /// Create a regex that matches any library or contract name inside a file
-pub fn create_contract_or_lib_name_regex(name: &str) -> Regex {
-    regex::Regex::new(&format!(r#"(?:using\s+(?P<n1>{name})\s+|is\s+(?:\w+\s*,\s*)*(?P<n2>{name})(?:\s*,\s*\w+)*|(?:(?P<ignore>(?:function|error|as)\s+|\n[^\n]*(?:"([^"\n]|\\")*|'([^'\n]|\\')*))|\W+)(?P<n3>{name})(?:\.|\(| ))"#, name = name)).unwrap()
-}
+pub static CREATE_RE_CONTRACT_OR_LIB_NAME: Lazy<fn(s: &str) -> Regex> = Lazy::new(|| {
+    |name: &str| {
+        regex::Regex::new(&format!(r#"(?:using\s+(?P<n1>{name})\s+|is\s+(?:\w+\s*,\s*)*(?P<n2>{name})(?:\s*,\s*\w+)*|(?:(?P<ignore>(?:function|error|as)\s+|\n[^\n]*(?:"([^"\n]|\\")*|'([^'\n]|\\')*))|\W+)(?P<n3>{name})(?:\.|\(| ))"#, name = name)).unwrap()
+    }
+});
```
mattsse, line 48: oh my bad, didn't realize that this is dynamic and requires `name`, so making this static is kinda pointless, you had it of course right before.

gakonst (approved):

PR #1198, opened 2022-04-30
Use EIP155 for all signers with empty transaction `chain_id`

### LINKED ISSUE #1189: Ledger sign_tx can not work anymore. always show LedgerError(BadRetcode(BadKeyHandle))
**Version**
ethers-signers = { version = "0.6.2", features = ["ledger"]}
ethers-core = "^0.6"

`cargo tree | grep ethers`
├── ethers-core v0.6.3
├── ethers-signers v0.6.2
│   ├── ethers-core v0.6.3 (*)

**Platform**
Darwin molindeMacBook-Pro-2.local 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:01 PDT 2021; root:xnu-8019.41.5~1/RELEASE_ARM64_T6000 arm64

**Description**
Ledger sign_tx can not work anymore. always show LedgerError(BadRetcode(BadKeyHandle))

I tried this code:

```
pub async fn signs_ledger() -> Result<(),()> {
  let ledger = Ledger::new(HDPath::LedgerLive(0), 1).await;
  match ledger {
    Ok(wallet) => {
      println!("Ledger: {:?}", wallet);

      let data = hex::decode("095ea7b30000000000000000000000007a250d5630b4cf539739df2c5dacb4c659f2488dffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff").unwrap();

      let tx_req = TransactionRequest::new()
        .to("0x2ed7afa17473e17ac59908f088b4371d28585476".parse::<Address>().unwrap())
        .gas(1000000)
        .gas_price(400e9 as u64)
        .nonce(5)
        .data(data)
        .value(ethers_core::utils::parse_ether(100).unwrap())
        .into();
      
      println!("tx_req {:?}", tx_req);
      
      let tx = wallet.sign_tx(&tx_req).await;
      match tx {
        Ok(tx_ret) => println!("txRet {:?}", tx_ret),
        Err(e) => println!("Err {:?}", e),
      }
    },
    Err(e) => {
      println!("ERROR {:?}", e.to_string());
    }
  }
  
  Ok(())
}
```

I expected to see this happen: 

```
Ledger: LedgerEthereum { transport: Mutex { is_locked: false, has_waiters: false }, derivation: LedgerLive(0), chain_id: 1, address: 0x0242e3f6ced3817bd4cbc80b66f59bf2970c157c }
tx_req Legacy(TransactionRequest { from: None, to: Some(Address(0x2ed7afa17473e17ac59908f088b4371d28585476)), gas: Some(1000000), gas_price: Some(400000000000), value: Some(100000000000000000000), data: Some(Bytes(b"\t^\xa7\xb3\0\0\0\0\0\0\0\0\0\0\0\0z%\rV0\xb4\xcfS\x979\xdf,]\xac\xb4\xc6Y\xf2H\x8d\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff")), nonce: Some(5) })
Err LedgerError(BadRetcode(BadKeyHandle))
```

Instead, this happened: LedgerError(BadRetcode(BadKeyHandle))

Rjected:
## Motivation

In #805, `chain_id` was removed from the `sighash` method for `TypedTransaction`. This causes issues like #1189, since transactions with an empty `chain_id` would have an incorrect rlp, sighash, and signature. In case the `chain_id` contained in the transaction is `None`, the signer should use its own `chain_id`.

## Solution

 * All signers will now prefer the transaction `chain_id`, using the signer `chain_id` only if the transaction `chain_id` is `None`.
 * Remove `InvalidTransactionError` in the private key wallet - this would need to be re-introduced if we want to enforce that transactions have a `chain_id` consistent with signers.
 * Refactor the AWS signer's `sign_digest_with_eip155` to include a `chain_id` parameter, rather than using the signer's `chain_id` every time. This way, transactions which specify a `chain_id` will have a correct `v`.
 * Refactor the wallet's `sign_hash` method to just sign hashes, moving eip155 `v` normalization to `sign_transaction_sync`.
 * Add a test for signing a transaction with an empty `chain_id` using a local wallet.

I've tested this locally with `chain_id: 1337` and a ledger, against a local wallet with the same private key. While the ledger _does_ receive the correct `chain_id` (I can see it on the screen), the `v` value of the returned signature is incorrect. The `r` and `s` values of the signature are identical to the signature produced from the local wallet.

Test results for different chains:
 * With `chain_id: 1337`, ledger returns `v: 150`. This is not the correct `v` value, for this message it should be `2710`.
 * With `chain_id: 109`, ledger returns `v: 253`. This is correct.
 * With `chain_id: 110`, ledger returns `v: 0`. This is clearly wrong - for this message it should be `256`.
 * With `chain_id: 111`, ledger returns `v: 2`. This is also wrong, but it seems like it's just an overflow, so we might be able to fix this `v` values.

I think this is because we only have [one byte](https://github.com/gakonst/ethers-rs/blob/81e7fea14b4f9bc4c531856f8cb7cd31a5248918/ethers-signers/src/ledger/app.rs#L194) for the resulting `v` value.
Is there any way for ledger to return a `v` that is over one byte in size? If not, we would have to fix the signature `v` value if the `chain_id > 110`.

Fixes #1189 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-signers/src/aws/mod.rs
+++ b/ethers-signers/src/aws/mod.rs
@@ -203,15 +203,19 @@ impl<'a> AwsSigner<'a> {
     }
 
     /// Sign a digest with this signer's key and add the eip155 `v` value
-    /// corresponding to this signer's chain_id
+    /// corresponding to the input chain_id
     #[instrument(err, skip(digest), fields(digest = %hex::encode(&digest)))]
-    async fn sign_digest_with_eip155(&self, digest: H256) -> Result<EthSig, AwsSignerError> {
+    async fn sign_digest_with_eip155(
+        &self,
+        digest: H256,
+        chain_id: u64,
+    ) -> Result<EthSig, AwsSignerError> {
         let sig = self.sign_digest(digest.into()).await?;
 
         let sig = utils::rsig_from_digest_bytes_trial_recovery(&sig, digest.into(), &self.pubkey);
 
         let mut sig = rsig_to_ethsig(&sig);
-        apply_eip155(&mut sig, self.chain_id);
+        apply_eip155(&mut sig, chain_id);
         Ok(sig)
     }
 }
@@ -230,13 +234,17 @@ impl<'a> super::Signer for AwsSigner<'a> {
         trace!("{:?}", message_hash);
         trace!("{:?}", message);
 
-        self.sign_digest_with_eip155(message_hash).await
+        self.sign_digest_with_eip155(message_hash, self.chain_id).await
     }
 
     #[instrument(err)]
     async fn sign_transaction(&self, tx: &TypedTransaction) -> Result<EthSig, Self::Error> {
+        let mut tx_with_chain = tx.clone();
+        let chain_id = tx_with_chain.chain_id().map(|id| id.as_u64()).unwrap_or(self.chain_id);
+        tx_with_chain.set_chain_id(chain_id);
+
         let sighash = tx.sighash();
-        self.sign_digest_with_eip155(sighash).await
+        self.sign_digest_with_eip155(sighash, chain_id).await
     }
 
     async fn sign_typed_data<T: Eip712 + Send + Sync>(
@@ -245,7 +253,7 @@ impl<'a> super::Signer for AwsSigner<'a> {
     ) -> Result<EthSig, Self::Error> {
         let hash = payload.encode_eip712().map_err(|e| Self::Error::Eip712Error(e.to_string()))?;
 
-        let digest = self.sign_digest_with_eip155(hash.into()).await?;
+        let digest = self.sign_digest_with_eip155(hash.into(), self.chain_id).await?;
 
         Ok(digest)
     }
--- a/ethers-signers/src/ledger/mod.rs
+++ b/ethers-signers/src/ledger/mod.rs
@@ -25,6 +25,11 @@ impl Signer for LedgerEthereum {
 
     /// Signs the transaction
     async fn sign_transaction(&self, message: &TypedTransaction) -> Result<Signature, Self::Error> {
+        let mut tx_with_chain = message.clone();
+        if tx_with_chain.chain_id() == None {
+            // in the case we don't have a chain_id, let's use the signer chain id instead
+            tx_with_chain.set_chain_id(self.chain_id);
+        }
         self.sign_tx(message).await
     }
 
--- a/ethers-signers/src/ledger/mod.rs
+++ b/ethers-signers/src/ledger/mod.rs
@@ -30,7 +30,7 @@ impl Signer for LedgerEthereum {
             // in the case we don't have a chain_id, let's use the signer chain id instead
             tx_with_chain.set_chain_id(self.chain_id);
         }
-        self.sign_tx(message).await
+        self.sign_tx(tx_with_chain).await
     }
 
     /// Signs a EIP712 derived struct
--- a/ethers-signers/src/ledger/mod.rs
+++ b/ethers-signers/src/ledger/mod.rs
@@ -26,7 +26,7 @@ impl Signer for LedgerEthereum {
     /// Signs the transaction
     async fn sign_transaction(&self, message: &TypedTransaction) -> Result<Signature, Self::Error> {
         let mut tx_with_chain = message.clone();
-        if tx_with_chain.chain_id() == None {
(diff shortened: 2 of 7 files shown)


mattsse:
> I think this is because we only have [one byte](https://github.com/gakonst/ethers-rs/blob/81e7fea14b4f9bc4c531856f8cb7cd31a5248918/ethers-signers/src/ledger/app.rs#L194) for the resulting v value.

This is correct, the ledger signer sends the  `SIGN_ETH_EIP_712 = 0x0C` command which signs the message following the EIP 712 spec and the v identifier is always 1 byte.

ethers-signers/src/ledger/mod.rs
```diff
@@ -25,6 +25,11 @@ impl Signer for LedgerEthereum {
 
     /// Signs the transaction
     async fn sign_transaction(&self, message: &TypedTransaction) -> Result<Signature, Self::Error> {
+        let mut tx_with_chain = message.clone();
```
mattsse, line 28: is this value even used?
because we're still using `message`: `self.sign_tx(message).await`

ethers-signers/src/trezor/mod.rs
```diff
@@ -25,7 +25,12 @@ impl Signer for TrezorEthereum {
 
     /// Signs the transaction
     async fn sign_transaction(&self, message: &TypedTransaction) -> Result<Signature, Self::Error> {
-        self.sign_tx(message).await
+        let mut tx_with_chain = message.clone();
+        if tx_with_chain.chain_id() == None {
```
mattsse, line 29:
```suggestion
        if tx_with_chain.chain_id().is_none() {
```

Also looks like the clone is only necessary if no chain id is set, so we can move the clone in the if branch instead and execute `sign_tx` in both branches but with different messages.
Same goes for the other calls

mattsse (approved):
thanks so much,
some nits re style :)

ntchjb:
Hi. For the issue on signature's `v` value. I have found the solution used by Ledger's JavaScript library that after got `v` value from Ledger, it is further modified to follow EIP-155 as shown here https://github.com/LedgerHQ/ledgerjs/blob/e78aac4327e78301b82ba58d63a72476ecb842fc/packages/hw-app-eth/src/Eth.ts#L298-L314.

Therefore, I propose a solution with this PR: #1204. Hope this helps 👍 .

ethers-signers/src/aws/mod.rs
```diff
@@ -230,13 +234,17 @@ impl<'a> super::Signer for AwsSigner<'a> {
...
         trace!("{:?}", message);
 
-        self.sign_digest_with_eip155(message_hash).await
+        self.sign_digest_with_eip155(message_hash, self.chain_id).await
     }
 
     #[instrument(err)]
     async fn sign_transaction(&self, tx: &TypedTransaction) -> Result<EthSig, Self::Error> {
+        let mut tx_with_chain = tx.clone();
+        let chain_id = tx_with_chain.chain_id().map(|id| id.as_u64()).unwrap_or(self.chain_id);
```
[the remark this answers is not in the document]
gakonst: unwrap_or seems fine here?

ethers-signers/src/ledger/mod.rs
```diff
@@ -25,6 +25,11 @@ impl Signer for LedgerEthereum {
...
+        if tx_with_chain.chain_id() == None {
+            // in the case we don't have a chain_id, let's use the signer chain id instead
+            tx_with_chain.set_chain_id(self.chain_id);
+        }
         self.sign_tx(message).await
```
gakonst, line 33:
```suggestion
        self.sign_tx(tx_with_chain).await
```

ethers-signers/src/trezor/mod.rs
gakonst: in the interest of fixing this bug asap, and given people don't really sign multiple txs in a row for this to be a perf issue, let's do the clone stuff as a follow up PR if necessary

gakonst (approved):

PR #1201, opened 2022-04-30
feat: remaining AST nodes

onbjerg:
Adds the remaining node types based on https://unpkg.com/solidity-ast@0.4.32/types.d.ts

Currently for coverage in Foundry I just use `other` on `Node` to get extra data, but I wonder what the best way would be to add more node-specific data? E.g. see `FunctionDefinition` where the statements in the function are contained in a `block` key (as opposed to `nodes`)

### DIFF
--- a/ethers-solc/src/artifacts/ast.rs
+++ b/ethers-solc/src/artifacts/ast.rs
@@ -1,7 +1,7 @@
 //! Bindings for solc's `ast` output field
 
 use crate::artifacts::serde_helpers;
-use serde::{Deserialize, Serialize};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::{collections::BTreeMap, fmt, fmt::Write, str::FromStr};
 
 /// Represents the AST field in the solc output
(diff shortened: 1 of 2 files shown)


gakonst:
cc @mattsse @shekhirin

ethers-solc/src/artifacts/ast.rs
```diff
@@ -18,24 +18,47 @@ pub struct Ast {
...
     #[serde(flatten)]
     pub other: BTreeMap<String, serde_json::Value>,
 }
 
-/// Represents the source location of a node : `<start>:<length>:<index>`
+impl Node {
+    /// Deserialize a serialized node attribute.
+    pub fn attribute<D: DeserializeOwned>(&self, key: impl AsRef<str>) -> Option<D> {
+        // TODO: Can we avoid this clone?
+        self.other.get(key.as_ref()).and_then(|v| serde_json::from_value(v.clone()).ok())
```
gakonst, line 57: This can probably improved with Serde's tagged enum by doing `enum NodeType { FunctionCall { foo: u8 } }` or whatever, but nbd now

gakonst (approved):

ethers-solc/src/artifacts/ast.rs
```diff
@@ -18,24 +18,47 @@ pub struct Ast {
...
+    /// Node attributes that were not deserialized.
     #[serde(flatten)]
     pub other: BTreeMap<String, serde_json::Value>,
 }
 
-/// Represents the source location of a node : `<start>:<length>:<index>`
+impl Node {
+    /// Deserialize a serialized node attribute.
+    pub fn attribute<D: DeserializeOwned>(&self, key: impl AsRef<str>) -> Option<D> {
+        // TODO: Can we avoid this clone?
```
mattsse, line 56: can't avoid because from value consumes the `Value`

PR #1203, opened 2022-05-01
Fix bug in provider docs

samlaf:
When running the first provider example, I get this error:
```
Error: ens name not found: 0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359`
```
I managed to fix this by parsing the string to a H160 type.

I'm new to rust so not 100% sure why it's behaving this way. The `get_code` (https://docs.rs/ethers-providers/0.6.2/src/ethers_providers/provider.rs.html#639) function uses
```
let at = match at.into() {
    NameOrAddress::Name(ens_name) => self.resolve_name(&ens_name).await?,
    NameOrAddress::Address(addr) => addr,
};
```
yet when I try doing the same outside of the get_code function:
```
    let address = "0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359".parse::<H160>()?;
    let at = match address.into() {
        NameOrAddress::Name(ens_name) => println!("ens name"),
        NameOrAddress::Address(addr) => println!("addr"),
    };
```
I get an error:
```
the trait bound `ethers_core::types::NameOrAddress: From<ethers::prelude::H160>` is not satisfied
the following implementations were found:
  <ethers_core::types::NameOrAddress as From<&str>>
  <ethers_core::types::NameOrAddress as From<ethers_core::types::H160>>
required because of the requirements on the impl of `Into<ethers_core::types::NameOrAddress>` for `ethers::prelude::H160`rustcE0277
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-providers/README.md
+++ b/ethers-providers/README.md
@@ -21,7 +21,8 @@ let provider = Provider::<Http>::try_from(
 let block = provider.get_block(100u64).await?;
 println!("Got block: {}", serde_json::to_string(&block)?);
 
-let code = provider.get_code("0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359", None).await?;
+let address = "0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359".parse::<H160>()?;
+let code = provider.get_code(address, None).await?;
 println!("Got code: {}", serde_json::to_string(&code)?);
 # Ok(())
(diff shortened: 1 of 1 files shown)


gakonst:
thanks for flagging!

PR #1204, opened 2022-05-01
Fix: Ledger should have correct signature v value for chains with large chain ID

ntchjb:
## Motivation

Since Ledger only return one byte of `signature.v` as stated in #1198, This issue prevent from sending transaction to chain that has large chain ID, such as, Arbitrum One (chain ID: 42161).

## Solution

I have derived a part of code from Ledger's JavaScript library and implement it as Rust. This part of code re-calculate signature's `v` value to fit with target chain.

The reference of Ledger's library can be found here: https://github.com/LedgerHQ/ledgerjs/blob/e78aac4327e78301b82ba58d63a72476ecb842fc/packages/hw-app-eth/src/Eth.ts#L298-L314

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst:
@ntchjb this makes a lot of sense - should we merge this in @Rjected's [PR](https://github.com/gakonst/ethers-rs/pull/1198/files)?

gakonst (approved):

Issue #592, opened 2021-11-18
Object Safe Ethers Middleware


gakonst:
**Is your feature request related to a problem? Please describe.**

The middleware stack right now, while flexible, is not object safe. This means that doing conditional instantiation of middlewares can be hard

**Describe the solution you'd like**

A trait design which is both Object Safe, and is nicely extensible the way it is right now, allowing consumers to only override the methods they specifically want, with everything else being delegated to the `fn inner()`.

**Describe alternatives you've considered**

@prestwich has a very promising candidate design here (big shoutout as this has been discussed for months): https://github.com/prestwich/ethers-middleware-experimental

oliver-giersch:
I have a different and somewhat bold proposal, which is likely going to be controversial as it would break existing APIs quite extensively and definitely deserves further discussion. The gist of it is basically this:

Make the `JsonRpcClient` and `PubSubClient` traits object-safe, deprecate the `Middleware` trait entirely and replace its primary use cases with wrapper structs around `Provider` and extension traits.

## Why?

To me, the design of the `Middleware` trait feels very un-rustlike, like something out of an object-oriented language clumsily adapted for rust. It provides a large subset of the Ethereum JSONRPC API, some additional higher-level abstractions (like `send_escalating`),  all mixed up together in a single trait. For most of these methods there is not really a conceivable use-case for ever overriding them. Some more random pain points, which could probably extended by a lot more I am currently not thinking of:

1. The "base" (innermost) implementation in `Provider` of `sign_transaction` just returns error, to me this indicates that this method does not belong at this level of abstraction.
2. The trait implementation often obscures what is going on (ever Ctrl+clicked on any of these methods to see what's going on? Well, bad luck cause all you get to see is the default impl in the trait forwarding to `inner`)

## Alternative

First, I would redesign the two most low-level transport-related traits as follows:

```rust
// errors are boxed, to keep the return slot small
pub type RequestPayload = Result<Box<RawValue>, Box<TransportErr>>;

pub trait Transport {
    fn request_id(&self) -> u64;
    // without async fn in traits yet
    fn send_request(&self, request: Box<RawValue>) -> Pin<Box<dyn Future<Output = RequestPayload>>>;
}
```
This design trades convenience in the base trait due to generics for full object safety. Note, that convenience can be easily added on top with the help of extension traits and blanket impls, e.g. like so (generics in extension traits are ok, as long as the base trait does not use any):

```rust
pub struct Request<'a, T> {
[...]

PR #1215, opened 2022-05-03
Break transports into a separate crate

prestwich:
Moves RPC transports (and related traits and error) out of `ethers-providers` into a new crate.

## Motivation

This is in service of #592. In order to introduce object-safe traits, we need to completely rewrite the provider stack. However, it is currently very difficult to do that in place. The providers crate is a bit of type spaghetti, with references to transports in many locations throughout the crate.

I started by attempting a refactor in place. However, due to the very-interconnected nature of the `ethers-providers` crate, this got out of hand. Transports rely on `ProviderError` and `Provider` specifically names several specific transports in `impl` blocks.

The goal of this PR is to introduce a simple layer boundary between transports, providers, and middleware. This will make it easier to work on future incremental changes towards object-safety. I think that small incremental changes in service of better layer boundaries are the only way to practically reach object-safe middleware :)

## Solution

- Break `ethers_providers::transports` into its own crate

##  Potential Followups

- Improve `ProviderError` and rename to something like `TransportError`
- differentiate between "transports" like HTTP and "meta-transports" like quorum and RW
- Rename `JsonRpcClient` to `RpcConnection` or `Transport`
- Remove the `Error` associated type from that trait
    - Refactor each transport with new object-safe `Transport` trait 

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Updated the changelog


mattsse (approved):
necessary first step.

agree with all points,

> differentiate between "transports" like HTTP and "meta-transports" like quorum and RW

This might not really be necessary if we use transports like `Arc<dyn Transport>`?

PR #1217, opened 2022-05-04
write smart contract

chappie1998:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/examples/write_smart_contract.rs
+++ b/examples/write_smart_contract.rs
@@ -1,0 +1,50 @@
+use ethers::{core::{abi::Abi,types::{Address, H256}}};
+use ethers::middleware::SignerMiddleware;
+use ethers::contract::Contract;
+use ethers::providers::{Provider, Http};
+use ethers::signers::{LocalWallet, Signer};
+use std::{convert::TryFrom, sync::Arc};
+use eyre::Result;
+
+#[tokio::main]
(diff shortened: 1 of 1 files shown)


chappie1998:
Added an example of writing a smart contract with a wallet's private keys.

examples/write_smart_contract.rs
```diff
@@ -0,0 +1,50 @@
...
+use eyre::Result;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    
+   // this wallet's private key
+   let wallet = "<private key>".parse::<LocalWallet>()?;
+   
+   // (ugly way to write the ABI inline, you can otherwise read it from a file)
+   let abi: Abi = serde_json::from_str(r#"<contract's abi>"#)?;
```
mattsse, line 16: this and the pk will of course fail.

if this "example" is rather a template, then this is fine.

But we should name this example accordingly and move this to a separate function so it's not run by default

mattsse (changes requested):

chappie1998:
Hello,

I wanted to make the write contract a universal file. So any user can put
their wallet's private key, infura API, contract address and contract ABI.
That's why I made the file like this.

What do you suggest? How should I put this file?


Regards
Ankit

PR #1221, opened 2022-05-04
feature: pending transaction retries repeatedly before assuming dropped

prestwich:
Implements this strategy for improving pending tx
https://github.com/foundry-rs/foundry/issues/1362#issuecomment-1106950009

## Motivation

Pending TX often prematurely reports transactions dropped due to mempool or RPC inconsistency across nodes. We'd like to minimize that.

## Solution

- Add a fixed number of retries (default 3) before the pending tx future assumes a tx was dropped.
- Allow the user to specify the number of retries
- if the tx is not found in the mempool, wait the polling interval and retry.
- If the tx cannot be found in the mempool after 3 retries 

is 3 the appropriate number? 3 polling intervals may be quite short

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog


mattsse (approved):
lgtm

gakonst (approved):

PR #1222, opened 2022-05-04
Robust gas oracles

recmo:
## Motivation

Currently most gas oracles construct a default`reqwest::Client`. By default, reqwest does not set a timout, meaning the gas oracles can hang indefinitely if the services is down. (It also prevents configuring proxies, caching and other reqwest options).

There is also no fail over mechanism in case an oracle service is failing.

## Solution

* Always pass explicit `reqwest::Client` to oracles.
* Add a `Median` aggregator oracle.

To discuss:

1. Modifying constructor arguments is a breaking change. I could have added a new constructor. I opted not to, because one oracle already had a `client` argument. At least now it is more consistent.
2. Is it correct to take separate medians for max_fee and priority_fee for EIP1559? This could mix responses from two different oracles.
3. gasnow.org seems shut down. Should we remove it?
4. Etherchain recommends a new gasnow style mempool based API: https://etherchain.org/tools/gasnow should we add it?

To do:

- [ ] Add an example that constructs a gas oracle with all bells and whistles (multiple sources, timeouts, caching, etc).
- [x]  Add a cache layer that stores gas prices for some time.
- [x]  Maybe a failover aggregator instead of median? i.e. use the first source that doesn't fail.
- [ ] A `default_from_chain` function that creates sensible defaults. Maybe a Middleware that recognizes the chain from chainId and constructs this default.
- [x] https://gasstation-mainnet.matic.network/v2
- [x] https://gasstation-mumbai.matic.today/v2
- [ ] https://blockscout.com/xdai/mainnet/api/v1/gas-price-oracle
- [ ] https://community.optimism.io/docs/useful-tools/oracles/
- [ ] Oracle based on `eth_gasPrice`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse (changes requested):
thanks for this, few nits

> Currently most gas oracles construct a defaultreqwest::Client. By default, reqwest does not set a timout, meaning the gas oracles can hang indefinitely if the services is down.

oh interesting, only the blocking version has a default timeout of 30 secs.

> Add a cache layer that stores gas prices for some time.
I guess this can be a simple GasOracly wrapper that records responses

> gasnow.org seems shut down. Should we remove it?
yep, site's gone, so +1 on rm this

recmo:
> * Maybe a failover aggregator instead of median? i.e. use the first source that doesn't fail.

I've added a weighted median instead, which allows you to mimic failover by assigning high weights to prioritized sources. Also allows you to implement more complex strategies.

mattsse (approved):
lgtm

gakonst:
concept ack

gakonst (approved):
Thx @recmo welcome to the contributors list 🤝

Merging in the interest of unblocking you, and we can fix further bugs as they arrive

Issue #1219, opened 2022-05-04
Struct in event gets flattened


nikkolasg:
Hi,

I use a struct in my solidity contract both as argument of a function and as argument in an event.
The function part works fine: ethers creates the equivalent Rust struct and I can give this into the contract method params.
The event part though works by flattening somehow the struct,i.e. ethers produces _another_ struct representing the same data, flattened, and this is the one that is given when retrieving events:

Ideally, both would use the same struct.

You can see a very minimal repro here:
* Here is the simple solidity code: [code](https://github.com/nikkolasg/foundry-event-flat/blob/master/contracts/src/Contract.sol#L4)
* The params for the function works fine here: [code](https://github.com/nikkolasg/foundry-event-flat/blob/master/bindings/src/my_contract.rs#L97)
  * See the struct in Rust
```rust
 pub struct Point {
        pub x: ethers::core::types::U256,
        pub y: ethers::core::types::U256,
    }
```
* The struct for the event  although is different : [code](https://github.com/nikkolasg/foundry-event-flat/blob/master/bindings/src/my_contract.rs#L129)
```rust
    pub struct NewPointFilter {
        pub x: (ethers::core::types::U256, ethers::core::types::U256),
    }
```

Thank you !

mattsse:
Code is correct, but what we#re not doing rn is resolving user-defined types in events.

But I believe we can do that.

PR #1235, opened 2022-05-08
test(abigen): ensure structs in events work

mattsse:
## Motivation
* Sanitiy test to ensure generated code is correct and works for structs in events and functions

ref #1219

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


ethers-contract/tests/abigen.rs
```diff
@@ -566,3 +569,23 @@ fn can_handle_overloaded_events() {
...
+#[tokio::test]
+async fn can_send_struct_param() {
+    abigen!(StructContract, "./tests/solidity-contracts/StructContract.json");
+
+    let server = Ganache::default().spawn();
+    let wallet: LocalWallet = server.keys()[0].clone().into();
+    let provider = Provider::try_from(server.endpoint()).unwrap();
+    let client = Arc::new(SignerMiddleware::new(provider, wallet.with_chain_id(1337u64)));
+
+    let contract = StructContract::deploy(client, ()).unwrap().legacy().send().await.unwrap();
```
gakonst, line 582: Separate point, the deploy function from abigen should not take an argument here?

gakonst (approved):

gakonst:
Any idea why the test is failing?

mattsse:
> Any idea why the test is failing?

ah because of additional celo fields?

disabling the test for celo then

PR #1236, opened 2022-05-08
add support for Emerald (from Oasis)

hiram-abif:
## Motivation

[Emerald](https://docs.oasis.dev/general/developer-resources/emerald-paratime/) is currently not part of the supported chains within `ethers-rs`.

## Solution

Added the Emerald Mainnet and Testnet chains.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


ethers-etherscan/src/lib.rs
```diff
@@ -263,6 +264,13 @@ impl ClientBuilder {
             Chain::Moonriver => {
                 urls("https://api-moonriver.moonscan.io/api", "https://moonriver.moonscan.io")
             }
+            Chain::Emerald => {
+                urls("https://explorer.emerald.oasis.dev/api", "https://explorer.emerald.oasis.dev")
+            }
+            Chain::EmeraldTestnet => urls(
+                "https://testnet.explorer.emerald.oasis.dev/api",
+                "https://testnet.explorer.emerald.oasis.dev",
+            ),
```
mattsse, line 273: is this compatible with etherscan API?

mattsse (changes requested):
chain ids lgtm

but explorer appears to be not related to etherscan, so I think we shouldn't add the endpoints here because etherscan API not supported?

ethers-etherscan/src/lib.rs
hiram-abif: @mattsse This project is using the fork for a small POC: https://gunix.codeberg.page/sycamore-metamask-example/
However that's only getting the balance and nothing else.

The explorer is [blockscout](https://github.com/blockscout/blockscout). According to the docs, it [supports ETH RPC](https://docs.blockscout.com/for-users/api#eth-rpc)

hiram-abif:
> but explorer appears to be not related to etherscan, so I think we shouldn't add the endpoints here because etherscan API not supported?

I double-checked and blockscout is claiming their API is compatible for people transitioning from etherscan. If you have concerns, I can write some rust tests to make sure that is true. They do have [a lot of tests](https://github.com/blockscout/blockscout/blob/master/apps/explorer/test/explorer/etherscan_test.exs), but they are all in Elixir.

0xvv (approved):
lgtm, the explorer appears to be compatible

ethers-etherscan/src/lib.rs
matevz: Yeah, we're using our own fork of the blockscout: https://github.com/oasisprotocol/emerald-blockscout

Let us know, if you require any assistance.

mattsse:
@rkrasiuk I believe we can close this since #1343 added emerald support?

vote close @gakonst

rkrasiuk:
@mattsse that's right

gakonst:
Thx

PR #1240, opened 2022-05-09
Release process and CI

meetmangukiya:


gakonst:
Should we add `cargo release --workspace patch --execute` as a weekly job, with a `workflow_dispatch: true` so that we can also manually trigger it?

onbjerg:
I'd caution against just pushing a patch release automatically; if we forget to manually publish and there are breaking changes, then we'd break people's stuff. Two ways: either we can hard commit to using conventional commits (per their spec) and use one of the many available actions to deduce major/minor/patch, or we do it manually. We could publish pre-releases automatically but I don't think crates.io supports those?

meetmangukiya:
This could potentially include breaking changes yeah. FWIW I think we have made breaking changes before too(would have to confirm) and we can make breaking changes on 0.x until we figure out a stable API we want to support and would've to worry about strict semver after the 1.x release? 

But we can make pre-releases as documented here https://doc.rust-lang.org/cargo/reference/resolver.html#pre-releases `major.minor.patch-alpha.X`. `cargo release --dev` could be used for that.

onbjerg:
Ahh, nice. Then I think we should use pre-releases for the automated workflow to ensure we don't ship breaking changes automatically on patch versions.

gakonst (approved):
LGTM - very excited about this, just configured the secret on the repo, let's see..

PR #1242, opened 2022-05-09
feat(etherscan): lookup solc build metadata

rkrasiuk:
## Motivation

[Foundry Issue #1567](https://github.com/foundry-rs/foundry/issues/1567)

## Solution

Added ability to lookup solc version with build metadata

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-etherscan/src/utils.rs
+++ b/ethers-etherscan/src/utils.rs
@@ -10,6 +10,9 @@ static SOLC_BIN_LIST_URL: &'static str =
 static RE_SOLC_VERSION: Lazy<Regex> =
     Lazy::new(|| Regex::new("soljson-v(?P<version>.*)\\.js").unwrap());
 
+/// Given the compiler version  lookup the build metadata
+/// and return full semver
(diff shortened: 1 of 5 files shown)


ethers-etherscan/src/utils.rs
```diff
@@ -0,0 +1,61 @@
...
+use regex::Regex;
+use semver::Version;
+
+use crate::{EtherscanError, Result};
+
+static SOLC_BIN_LIST_URL: &str =
+    "https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/list.txt";
+
+static RE_SOLC_VERSION: Lazy<Regex> =
+    Lazy::new(|| Regex::new("soljson-v(?P<version>.*)\\.js").unwrap());
```
mattsse, line 11: looks like this can also be achieved via a simple trim_(start|end)_matches, then we don't need once_cell + regex

rkrasiuk: good point 👍

mattsse (approved):
lgtm

PR #1250, opened 2022-05-11
Add Block::time

recmo:
## Motivation

Quality of life addition.

## Solution

It adds `chrono` as a dep which wasn't there before. It's added without default-features so should be pretty lightweight.

Let me know if we want to feature flag the dep though.

Aside: I also really wish we had `TryInto<{u64, i64, u32, usize, …}> for U256`, but this needs to happen in the `ethabi` crate which doesn't see much activity.  Similarly the `u256_from_f64_saturating` should be a member fn. Does it make sense to internalize the `U256` / `H256` implementations?

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/types/block.rs
+++ b/ethers-core/src/types/block.rs
@@ -93,10 +93,14 @@ pub struct Block<TX> {
     pub epoch_snark_data: Option<EpochSnarkData>,
 }
 
+/// Error returned by [`Block::time`].
 #[derive(Clone, Copy, Debug, Error)]
 pub enum TimeError {
+    /// Timestamp is zero.
     #[error("timestamp is zero")]
     TimestampZero,
+
+    /// Timestamp is too large for [`DateTime<Utc>`].
     #[error("timestamp is too large")]
     TimestampOverflow,
 }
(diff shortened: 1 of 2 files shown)


gakonst:
> Aside: I also really wish we had TryInto<{u64, i64, u32, usize, …}> for U256, but this needs to happen in the ethabi crate which doesn't see much activity. Similarly the u256_from_f64_saturating should be a member fn. Does it make sense to internalize the U256 / H256 implementations?

My deep desire is to:
1. Get ride of primitive_types / ethereum_types, and replace them with newer versions using const generics which ideally are also faster
2. Replace rlp with fastrlp
3. Replace ethabi with something like https://github.com/nlordell/solabi-rs/

And we can then solve a lot of issues like:
1. Type conversions
2. Faster ops
3. Better ABI ergonomics
4. ..?

So you can go ahead and try upstreaming any changes you want, but as you said, maintenance is a bit slow on these crates, so might not be worth the time investment.

gakonst (approved):

PR #1251, opened 2022-05-12
More gas oracles

recmo:
## Motivation

This adds the `ProviderOracle` that allows runtime selection between provider gas methods and gas oracles. Such runtime selection is very useful for building stacks that work on multiple chains, such as production and local test.

~~It also adds a `default_for_chain` constructor that looks at the `chain_id` and tries to build a sensible default. This is a bit proof-of-concept. It would be nice to add a builder pattern where you can provide api-keys.~~

Architecturally this all feels a bit off to me. `ProviderOracle` exists basically because `GasOracle` and `Middleware` overlap in functionality. The `Provider` implementation also does quite a bit of magic under the hood to get EIP1559 estimates. IMHO it would be cleaner if this was all done through something like GasOracle's. But simply factoring it out doesn't work because gas prices are required to sign transactions, and at least currently Provider provides a complete implementation. Makes me think the Middleware really should be broken up in separate parts relating to the transaction lifecycle (gas pricing, gas estimation, gas escalation, nonce, signing, broadcasting, monitoring, etc). (Also seems related to #592)

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-middleware/src/gas_oracle/provider_oracle.rs
+++ b/ethers-middleware/src/gas_oracle/provider_oracle.rs
@@ -1,0 +1,40 @@
+use crate::gas_oracle::{GasOracle, GasOracleError};
+use async_trait::async_trait;
+use ethers_core::types::U256;
+use ethers_providers::Middleware;
+use std::fmt::Debug;
+
+/// Gas oracle from a [`Middleware`] implementation such as an
+/// Ethereum RPC provider.
+#[derive(Debug)]
+pub struct ProviderOracle<M: Middleware> {
+    provider: M,
+}
+
+impl<M: Middleware> ProviderOracle<M> {
+    pub fn new(provider: M) -> Self {
+        Self { provider }
+    }
+}
+
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+impl<M: Middleware> GasOracle for ProviderOracle<M>
+where
+    <M as Middleware>::Error: 'static,
+{
+    async fn fetch(&self) -> Result<U256, GasOracleError> {
+        self.provider
+            .get_gas_price()
+            .await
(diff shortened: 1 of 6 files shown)


mattsse:
cc @oliver-giersch

ethers-middleware/src/gas_oracle/provider_oracle.rs
```diff
@@ -0,0 +1,40 @@
...
+{
+    async fn fetch(&self) -> Result<U256, GasOracleError> {
+        self.provider
+            .get_gas_price()
+            .await
+            .map_err(|err| GasOracleError::ProviderError(Box::new(err)))
+    }
+
+    async fn estimate_eip1559_fees(&self) -> Result<(U256, U256), GasOracleError> {
+        // TODO: Allow configuring different estimation functions.
```
mattsse, line 34: perhaps something like a closure that returns a future?

mattsse (approved):
lgtm

thanks for this, definitely something we need to keep in mind for #592

gakonst:
> Makes me think the Middleware really should be broken up in separate parts relating to the transaction lifecycle (gas pricing, gas estimation, gas escalation, nonce, signing, broadcasting, monitoring, etc). (Also seems related to https://github.com/gakonst/ethers-rs/issues/592)

This is not a bad idea, although it does prevent you from join'ing all the futures for gas price + gas estimation + nonce, which is a nice performance boost. Do you wanna drop any thoughts on the architecture in the issue?

gakonst:
(Think this needs some fixes around WASM)

recmo:
I removed the `default_for_provider` function. It was constructing a reqwest client with timeout, which is not available on WASM, causing the build to fail. I split the `LogQueryError` out to it's own PR #1615 .

gakonst (approved):
sweet

PR #1253, opened 2022-05-12
feat(solc): add source map access functions

mattsse:
## Motivation
extend `Artifact` with source map getters

cc @onbjerg 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


onbjerg (approved):
very useful for both the debugger and coverage

PR #1258, opened 2022-05-13
Add basic solc model checker options

leonardoalt:
Foundry's https://github.com/foundry-rs/foundry/pull/1602 depends on this PR.

This PR adds [Solidity's model checker](https://docs.soliditylang.org/en/latest/smtchecker.html#) options to ethers-solc's settings.

## Motivation

Mainly to allow Foundry users to have direct access to the model checker.
The default case of these options do not enable the model checker since it's opt-in, so the PR shouldn't disturb users who are not interested in the model checker.

## Solution

Add data structures that represent a basic subset of the [tool's options](https://docs.soliditylang.org/en/latest/smtchecker.html#smtchecker-options-and-tuning).
The 4 added options are the most basic to enable the tool. Other options such as `divModNoSlacks` and `invariants` can be added later on.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -122,7 +122,11 @@ impl CompilerInput {
                 // introduced in <https://docs.soliditylang.org/en/v0.8.10/using-the-compiler.html#compiler-api>
                 // <https://github.com/ethereum/solidity/releases/tag/v0.8.10>
                 debug.debug_info.clear();
+
             }
+
+            // 0.8.10 is the earliest version that has all model checker options.
+            self.settings.model_checker = None;
         }
 
         self
@@ -844,8 +848,8 @@ pub struct MetadataSource {
 /// Model checker settings for solc
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ModelCheckerSettings {
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub contracts: Option<BTreeMap<String, Vec<String>>>,
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub contracts: BTreeMap<String, Vec<String>>,
     #[serde(
         default,
         with = "serde_helpers::display_from_str_opt",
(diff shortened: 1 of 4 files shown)


ethers-solc/src/artifacts/mod.rs
```diff
@@ -837,6 +841,112 @@ pub struct MetadataSource {
     pub license: Option<String>,
 }
 
+/// Model checker settings for solc
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ModelCheckerSettings {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub contracts: Option<BTreeMap<String, Vec<String>>>,
```
mattsse, line 848:
```suggestion
    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
    pub contracts: BTreeMap<String, Vec<String>>,
```

mattsse (approved):
nice! ty

these are supported by all solc versions, right? otherwise we'd need to update the input based on the targeted version:

https://github.com/gakonst/ethers-rs/blob/44cbbc769a4139c023ca6f5e22b4ea8b6ead23b3/ethers-solc/src/artifacts/mod.rs#L102-L104

leonardoalt:
> these are supported by all solc versions, right? otherwise we'd need to update the input based on the targeted version:

Oh, good point! Definitely not supported by all versions. How about making it available only for compiler versions >=0.8.something? I'll find out the earliest version that has all of these options.

leonardoalt:
- 0.8.4 is the earliest for the options in this PR
- 0.8.7 adds other options that are not here yet
- 0.8.10 same as /\

I'd suggest enabling it only for >=0.8.10.

mattsse:
> * 0.8.4 is the earliest for the options in this PR
> * 0.8.7 adds other options that are not here yet
> * 0.8.10 same as /\
> 
> I'd suggest enabling it only for >=0.8.10.

sounds good! so we can simply reuse the existing `PRE_V0_8_10`

ethers-solc/src/artifacts/mod.rs
leonardoalt: Given this suggestion, would it also make sense to do the same for `    pub targets: Option<Vec<ModelCheckerTarget>>,`?

leonardoalt: Also this forces the test to be changed to
```
    project.project_mut().solc_config.settings.model_checker = Some(ModelCheckerSettings {
        contracts: BTreeMap::new(),
        ...
```
Is there a reason why not using `Option` is preferred?

gakonst (approved):
nice

Issue #1259, opened 2022-05-13
async usage of `svm-rs`


rkrasiuk:
**Is your feature request related to a problem? Please describe.**
usage of `svm-rs` with a `blocking` feature inside a tokio runtime leads to the following error
`Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.`
reqwest::blocking doesn't behave well inside another tokio rt

**Describe the solution you'd like**
use `svm` async threadbound 

cc @mattsse

mattsse:
this is unfortunatelly a limitation of reqwest::blocking

what we could do

1. use svm-async within 

```rust
 runtime::Builder::new_current_thread()
                .enable_io()
                .build().unwrap().block_on(async {svm::install})
```
2. try accessing the Handle or create a new multithreaded rt


actually not sure about the performance benefits of 2 over 1

gakonst:
this is a 2-step process:
1. https://github.com/foundry-rs/foundry/blob/master/utils/src/lib.rs#L27-L54 move this over to ethers-solc (and expose it so we can import it in foundry)
2. Use `RuntimeOrHandle::new().block_on(...)` for the 2 fns where `blocking_install` is called:
    * https://github.com/gakonst/ethers-rs/blob/eb94e53d1f87dc5fe10f25b7268e977cfad24984/ethers-solc/src/resolver/mod.rs#L724
    * https://github.com/gakonst/ethers-rs/blob/eb94e53d1f87dc5fe10f25b7268e977cfad24984/ethers-solc/src/resolver/mod.rs#L742

rkrasiuk:
@gakonst and why don't we `block_on` the actual `svm::blocking_install`?
https://github.com/gakonst/ethers-rs/blob/eb94e53d1f87dc5fe10f25b7268e977cfad24984/ethers-solc/src/compile/mod.rs#L376

gakonst:
I guess that also works :D

mattsse:
vote close @gakonst

rkrasiuk:
second @mattsse. my foundry integration tests seem to be fixed

PR #1260, opened 2022-05-13
fix(solc): add `RuntimeOrHandle` & fix solc blocking installation

rkrasiuk:
## Motivation

#1259 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/Cargo.toml
+++ b/ethers-solc/Cargo.toml
@@ -39,6 +39,7 @@ solang-parser = { default-features = false, version = "=0.1.13" }
 rayon = "1.5.2"
 rand = { version = "0.8.5", optional = true }
 path-slash = "0.1.4"
+cfg-if = "1.0.0"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 home = "0.5.3"
--- a/ethers-solc/Cargo.toml
+++ b/ethers-solc/Cargo.toml
@@ -19,7 +19,7 @@ serde_json = "1.0.68"
 serde = { version = "1.0.130", features = ["derive"] }
 semver = { version = "1.0.9", features = ["serde"] }
 walkdir = "2.3.2"
-tokio = { version = "1.15.0", default-features = false, features = ["process", "io-util", "fs", "time"], optional = true }
+tokio = { version = "1.15.0", default-features = false, features = ["process", "io-util", "fs", "time"] }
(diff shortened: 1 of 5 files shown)


ethers-solc/Cargo.toml
```diff
@@ -19,7 +19,7 @@ serde_json = "1.0.68"
 serde = { version = "1.0.130", features = ["derive"] }
 semver = { version = "1.0.9", features = ["serde"] }
 walkdir = "2.3.2"
-tokio = { version = "1.15.0", default-features = false, features = ["process", "io-util", "fs", "time"], optional = true }
+tokio = { version = "1.15.0", default-features = false, features = ["process", "io-util", "fs", "time"] }
```
mattsse, line 22: I don't think we need all features here just for the executor?

ethers-solc/src/compile/mod.rs
```diff
@@ -371,9 +371,11 @@ impl Solc {
     /// Blocking version of `Self::install`
     #[cfg(all(feature = "svm-solc"))]
     pub fn blocking_install(version: &Version) -> std::result::Result<Self, svm::SolcVmError> {
```
mattsse, line 373: let's add some docs why we're doing this

ethers-solc/src/compile/mod.rs
```diff
@@ -799,6 +801,17 @@ mod tests {
...
     }
 
+    #[test]
+    #[cfg(feature = "svm-solc")]
+    fn can_install_solc_in_tokio_rt() {
+        let ver = "0.8.6";
+        let version = Version::from_str(ver).unwrap();
+        let result = tokio::runtime::Runtime::new()
+            .unwrap()
+            .block_on(async { Solc::blocking_install(&version) });
```
mattsse, line 811: great!

can we also add some integration tests that pin a specific solc version and make the tests `#[tokio::test] async`?

mattsse (changes requested):
great, let's add some additional docs and project tests

mattsse (changes requested):
this currently fails on wasm due to tokio, but `reqwest::blocking` 

lets do some cfgs to use `svm::blocking_install` directly in wasm

mattsse (approved):
lgtm

gakonst (approved):

Issue #954, opened 2022-02-23
Pagination / Streaming for Event Logs


3lpsy:
**Is your feature request related to a problem? Please describe.**
As event logs can yield more results than an RPC will allows (for example, infura failing when the response has over 10k items), looping over a stream of event logs based on a filter or providing a paginated API would be very useful.

**Describe the solution you'd like**

Gnosis has a potential solution [here](https://github.com/gnosis/ethcontract-rs/blob/39d40e06428df7da6d36318054ca672ff4e2d115/ethcontract/src/log.rs#L321) where `past_logs_pages`/`stream` can return a filter based log stream. 

I think something similar would be the easiest implementation. The draft implementation could start by adding a `get_logs_pages` method on the provider that accepts a `Filter` object. However, I am open to feedback / suggestions in terms of how it's designed.

**Additional context**
Understanding the nuances of how eth_getLogs is handled by the RPC provider may be required and may introduce additional complexity. Below is a list of use cases to research to determine how they differ and how to best handle their use cases: 

- eth on infura
- eth on alchemy
- polygon on infura 
- polygon on alchemy
- bsc on infura

3lpsy:
Some additional thoughts and notes:

Most probably knew this but I didn't. `eth_getLogs` doesn't appear to support pagination at the RPC level. There was an [attempt](https://github.com/ethereum/go-ethereum/issues/17487) but it apparenttly didnt' get any traction.

The de facto approach appears to be use the block range as a substitute for a `page_limit`. This is unfortunate, but seems to be the only available solution. 

There are edge cases where a single block may contain >1000 (or an RPCs max responses) events. Infura actually bumped their response limit up to 10k for this reason. I am basing this off of [this thread from 2019](https://community.infura.io/t/getlogs-error-query-returned-more-than-1000-results/358/15). While dynamically resizing the block range down may be a useful feature, I'm not sure if accomadating this edge case where a single block has greater than 1000 responses (or the limit of the RPC) is worth it as it would require splitting topics or addresses. 

So a starting solution could have the user pass in a filter and a `page_range` (though i prefer to call it a `block_range` as it's more explicit what it is). Then the provider's `get_logs_pages` method could create a new filter based off the filter's `FilterBlockOption` with a new `to_block` and then iterate upon it. A nice feature in this case would be catching an `too many responses` error and attempting to downsize the range to handle this. Though then we have to determine whether to keep the downsized range for the remainder of the stream or move it back to the original value.

[...]

PR #1264, opened 2022-05-14
feat(providers): load previous logs before subscribing

### LINKED ISSUE #988: Stream events, including past ones

Currently there are two ways to query events:
- with a filter for a given range:`contract.events().from_block(block_number).query().await`. This allows to query events that have been emitted in the past.
- streaming new events: `contract.events().stream().await`. This will only yield events that are emitted after the stream has been created.

Ideally, there would be a way to get a stream of events from a given point in the past. I.e. a function that combines both above functions.

**Additional context**
See https://github.com/gakonst/ethers-rs/issues/986

meetmangukiya:
Load previous logs and stream it back to the user before establishing
a new stream for streaming future logs.

Closes #988

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-providers/src/provider.rs
+++ b/ethers-providers/src/provider.rs
@@ -28,7 +28,9 @@ use thiserror::Error;
 use url::{ParseError, Url};
 
 use futures_util::{lock::Mutex, try_join};
-use std::{convert::TryFrom, fmt::Debug, str::FromStr, sync::Arc, time::Duration};
+use std::{
+    collections::VecDeque, convert::TryFrom, fmt::Debug, str::FromStr, sync::Arc, time::Duration,
+};
 use tracing::trace;
 use tracing_futures::Instrument;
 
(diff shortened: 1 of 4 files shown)


ethers-providers/src/provider.rs
```diff
@@ -1102,9 +1102,29 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+                } else {
+                    self.get_logs(filter).await
+                }
+            }
+            FilterBlockOption::AtBlockHash(_block_hash) => self.get_logs(filter).await,
+        };
+
+        if logs.is_err() {
+            return Err(logs.err().unwrap())
+        }
```
mattsse, line 1118: can't we use the `?` here?

ethers-providers/src/pubsub.rs
```diff
@@ -31,6 +31,8 @@ pub struct SubscriptionStream<'a, P: PubsubClient, R: DeserializeOwned> {
     /// The subscription's installed id on the ethereum node
     pub id: U256,
 
+    loaded_elements: Vec<R>,
```
mattsse, line 34: looks like we can use a VecDeque here to easily `pop_front`?

mattsse (changes requested):
smol nits

ethers-providers/src/pubsub.rs
meetmangukiya: good catch

ethers-providers/src/provider.rs
```diff
@@ -1102,9 +1104,24 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
     where
         P: PubsubClient,
     {
+        let loaded_logs = match filter.block_option {
+            FilterBlockOption::Range { from_block, to_block: _ } => {
+                if from_block.is_none() {
+                    vec![]
+                } else {
+                    self.get_logs(filter).await?
```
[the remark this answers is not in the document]
gakonst: There is an issue for this already I believe, #954 cc @meetmangukiya

PR #1265, opened 2022-05-14
fix: all request ids start at 1

oliver-giersch:
## Motivation

I came across an issue in foundry trying to deploy a contract to fantom using the official RPC `https://rpc.ftm.tools/`. A call to `eth_chainId` fails to deserialize, because the sent out request has the ID `zero`, to which the response is `{"jsonrpc":"2.0","id":null,"result":"0xfa"}`. For some reason the server fails to unmarshall the request ID and responds with a `null` ID. When sending a request with a non-zero ID, everything works fine. Strictly speaking, this is an error on their side, since nothing in the JSON-RPC forbids a zero ID and I am not entirely sure why this occurs, since go-opera uses the `rpc` package straight from the go-ethereum repo and I don't think geth exhibits this behaviour. Anyways, the fix is simple and the alternative behaviour is consistent with other eth libraries like ethers-js.

## Solution

All request IDs for all transports start at 1.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-providers/src/transports/common.rs
+++ b/ethers-providers/src/transports/common.rs
@@ -212,6 +212,8 @@ impl fmt::Display for Authorization {
 
 #[cfg(test)]
 mod tests {
+    use ethers_core::types::U64;
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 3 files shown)


gakonst (approved):
Seems OK to do

PR #1266, opened 2022-05-15
fix(ethers-solc/ConfigurableArtifacts): output methodIdentifiers by default

meetmangukiya:
Related https://github.com/foundry-rs/foundry/issues/1497

## Motivation

As per https://github.com/gakonst/foundry/blob/0062360706e77a155856e128c3e250264c75f48c/config/src/lib.rs#L199-L214 `methodIdentifiers` should be enabled as extra output by default as well in addition to `abi`, `bytecode` and `deployedBytecode`. Making it default in ethers as other fields, but can also inject `methodIdentifiers` as `extra_output` in `Config::extra_output`.

 

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifact_output/configurable.rs
+++ b/ethers-solc/src/artifact_output/configurable.rs
@@ -113,7 +113,8 @@ impl<'a> From<&'a ConfigurableContractArtifact> for CompactContractBytecodeCow<'
 ///  {
 ///    "abi": [],
 ///    "bytecode": {...},
-///    "deployedBytecode": {...}
+///    "deployedBytecode": {...},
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

meetmangukiya:
Yep, thought of adding it there, but the other 3 defaults were made default in ethers, so went that way for consistency

gakonst (approved):
ACK - let's merge this then to unblock the Foundry issue, and think about the caching issue and replacing `Option<BTreeMap>` with `BTreeMap` in a follow up

Issue #1154, opened 2022-04-18
Some RPC methods incorrectly us `Option<BlockId>` instead of `Option<BlockNumber>`


prestwich:
**Version**
master

**Platform**
osx

**Description**
Some RPC methods take `Option<BlockNumber>` by spec, but the Middleware trait has them taking `Option<BlockId>` instead. This prevents users from calling with `pending` or `earliest` (`latest` can be accessed, as `None` maps to `latest`)

https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/provider.rs#L478
https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/provider.rs#L478

Issue #969, opened 2022-02-26
A version of `PendingTransaction` that owns the provider


meetmangukiya:
**Is your feature request related to a problem? Please describe.**

I think currently it is not possible for a function to return a pending transaction because it refers to the provider that sent the tx. Provider maybe behind an `Arc` that gets passed to the function as param. So maybe we need a version where `PendingTransaction` can hold a ref to the `Arc` as well so it can be retuned?

I was trying to write something like:
```rs
pub async fn simulate_and_send_tx_min<'a, M, D>(
    contract_call: ContractCall<M, D>,
    provider: Arc<M>,
    simulate: bool,
) -> eyre::Result<PendingTransaction<'a, M::Provider>>
where
    M: Middleware + 'static + Clone,
    M::Provider: 'a,
    D: Detokenize + Clone,
{
    if simulate {
        let res = contract_call.call().await;
        if res.is_err() {
            warn!("simulation failed, not sending tx");
        }
        res?;
    }

    let tx = contract_call.tx;
    let res = provider.send_transaction(tx, None).await?;
    Ok(res)
}
```

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Issue #876, opened 2022-02-07
Generic function to get a provider for any type of url / transport


meetmangukiya:
**Is your feature request related to a problem? Please describe.**
Right now if you want to develop something that can work with different transports, you would have to rewrite it or configure it somehow. Basically an equivalent to ethers-js' `new JsonRpcProvider(url: string)`

**Describe the solution you'd like**
`pub fn get_provider<P>(url: String) -> Provider<P>` signature function would be good to have.

**Describe alternatives you've considered**
I see a workaround / solution implemented on yield-liquidator where you create another generic main function that you will provide the provider to https://github.com/yieldprotocol/yield-liquidator/blob/master/src/main.rs#L74

meetmangukiya:
I have thought of a different way we could this that might be a smaller refactor compared to object safety in case we want to do this before that.

```rs
pub struct Provider<P> {
    inner: P,
    ens: Option<Address>,
    interval: Option<Duration>,
    from: Option<Address>,
    /// Node client hasn't been checked yet = `None`
    /// Unsupported node client = `Some(None)`
    /// Supported node client = `Some(Some(NodeClient))`
    _node_client: Arc<Mutex<Option<NodeClient>>>,
}
```
This is the current `Provider` structure which is generic over the underlying transports: http, ws, ipc, etc.
I suggest we create a new transport that is an enumeration over existing transports:
```rs
#[non_exhaustive]
pub enum CommonProviderTransport {
    Ws(Ws),
    Http(Http),
    #[cfg(all(target_family = "unix", feature = "ipc"))]
    Ipc(Ipc),
}

impl JsonRpcClient for CommonProviderTransport { .. }
```
and creating a provider that is not generic and the transport is set to `CommonProviderTransport` like so:
```rs
pub struct Provider<P> {
    transport: CommonProviderTransport,
    ens: Option<Address>,
    interval: Option<Duration>,
    from: Option<Address>,
    /// Node client hasn't been checked yet = `None`
    /// Unsupported node client = `Some(None)`
    /// Supported node client = `Some(Some(NodeClient))`
    _node_client: Arc<Mutex<Option<NodeClient>>>,
}
```

[...]

PR #1267, opened 2022-05-15
Feat object safe transports

oliver-giersch:
This is the realization of the design proposed in #592.

## Motivation

The current `Transport` trait is not object-safe, meaning it is not easily possible to dynamically abstract over different connection types at runtime. It is also heavily entangled with the `Middleware` trait, obscuring in some cases what happens, when API calls are made (or even, if API calls are made at all).

## Solution

Rewrite `Transport` trait and implementations to be object-safe, untangle `Middleware` trait and move entirely to `ethers-middleware`, making it optional for certain use-cases.

### Issues to be Resolved by this PR

- #1154 - the rewritten `Provider` type will expose the Ethereum JSON-RPC in type-safe manner, with mostly a 1:1 mapping between RPC calls and exposed methods
- #969 - with the new `Provider` API, the `PendingTransaction` type could/should be made generic over owned & borrowed providers
- #876 - this is already implemented, the signature is `async fn [Provider::]connect(url: &str) -> Result<Provider<Arc<dyn Connection>>, ProviderError>`
- #592

## Points of Contention

Some points I would like to discuss before moving forward on them:

1. There is a bunch of types in `ethers-core` that only exist due to the way the JSON-RPC API is structured, such as `SyncStatus` (models the return type of `eth_syncing`) or `BlockNumber` (allows calling certain methods with either a block number or a generic tag like "latest"). I would like to move all those types to `ethers_providers::types`.
2. The "meta-transport" implementations `Quorum` and `ReadWrite`. The latter in particular feels like something that could be easily handwritten by each user requiring such functionality and is not really worth the cost of maintaining it at a library level, but feel free to voice your concerns in the comments. Both could probably be "hackily" implemented as `Connection` impls by inspecting the "raw" request strings and searching for method call names.
3. To me, `ethers-providers` should provide an API that fairly closely matches the Ethereum JSON-RPC API, so `send_raw_transaction` should just return a hash, `get_block_by_number`/`get_block_by_hash` should be distinct functions, etc. Convenience should be added on top of that at a higher level (i.e., for now, the `Middleware` trait). 

## PR Checklist

- [x] Implemented new `Connection` & `DuplexConnection` traits 
- [x] Implemented HTTP connection
- [x] Implemented IPC connection
- [ ] Implemented WS connection
- [ ] Implemented `Quorum`/`ReadWrite` (discussion re potential removal?)
- [ ] Implement full JSON-RPC API
- [ ] Moved `Middleware` trait to `ethers-middleware`
- [ ] Moved types that are only RPC-related from `ethers-core` to `ethers-providers` (discussion?) 
- [ ] Merge with `ethers-providers` 
- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse:
looks like the core `Connection` is well defined now, so what's left is porting the existing Provider impls

do you have the bandwidth to bring (parts of) it over the line @oliver-giersch? 

Otherwise, I can take over the provider migration

oliver-giersch:
> looks like the core `Connection` is well defined now, so what's left is porting the existing Provider impls, right?
> 
> do you have the bandwidth to bring (parts of) it over the line @oliver-giersch?
> 
> Otherwise, I can take over the provider migration

I've done a lot of further API experimentation in a separate repo that I am pretty happy with, including a pretty neat way to handle batch request, but I am indeed strapped for time currently. Notably, the WS provider (especially for WASM) is still missing as well as some less important API calls and ofc tests and documentation, so help would definitely be appreciated. I will port this over to the PR branch in the next days and report back

oliver-giersch:
Apologies for the delay, but as I said, I am a bit short on free time atm. The latest commit ports the remaining stuff from my experimental repo over to the PR branch and fixes almost all currently implemented tests. Basic HTTP and IPC connections work, (non-WASM) WS is halfway implemented and a large portion of the Ethereum JSON-RPC API is implemented (the rest can be fairly easily extended). Batch requests work, but could be extended to give more options (i.e. an additional API as in #1339, which should be compatible with the current system).

PR #1277, opened 2022-05-17
fix(solc): follow symlinks in source files

mattsse:
## Motivation
Closes https://github.com/foundry-rs/foundry/issues/1244
by following symlinks in source.
I'd consider this an unusual setup, nevertheless following symlinks in src/test folder is possible

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
indeed i'd also consider unusual but makes sense to me

PR #1278, opened 2022-05-17
feat(core): add  helpers to unwrap `TypedTransaction`

joshieDo:
## Motivation
ref: https://github.com/foundry-rs/foundry/pull/1208#discussion_r872909522

Helpers to unwrap `TypedTransaction` into whichever type the user requires.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/types/transaction/eip2718.rs
+++ b/ethers-core/src/types/transaction/eip2718.rs
@@ -448,7 +448,8 @@ impl TypedTransaction {
     fn into_legacy(self) -> TransactionRequest {
         match self {
             Legacy(tx) => tx,
-            _ => TransactionRequest {
+            Eip2930(tx) => tx.tx,
+            Eip1559(_) => TransactionRequest {
                 from: self.from().copied(),
                 to: self.to().cloned(),
                 nonce: self.nonce().copied(),
--- a/ethers-core/src/types/transaction/eip2718.rs
+++ b/ethers-core/src/types/transaction/eip2718.rs
@@ -402,33 +402,30 @@ impl From<&Transaction> for TypedTransaction {
 }
 
 impl TypedTransaction {
-    #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


ethers-core/src/types/transaction/eip2718.rs
```diff
@@ -398,6 +401,143 @@ impl From<&Transaction> for TypedTransaction {
     }
 }
 
+impl TypedTransaction {
+    #[allow(dead_code)]
+    fn as_legacy_ref(&self) -> Result<&TransactionRequest, TypedTransactionError> {
```
mattsse, line 406: these are all private, and marked as dead?

I think they should rather return a simple `Option<>`

ethers-core/src/types/transaction/eip2718.rs
```diff
@@ -398,6 +401,143 @@ impl From<&Transaction> for TypedTransaction {
...
+}
+
+impl From<TypedTransaction> for Eip1559TransactionRequest {
+    fn from(src: TypedTransaction) -> Eip1559TransactionRequest {
+        src.into_eip1559()
+    }
+}
+
+impl TypedTransaction {
+    fn as_legacy(&self) -> TransactionRequest {
```
mattsse, line 466: since we basically clone every field, I think we should take ownership here so the caller can decide if a clone is necessary, otherwise, converting always requires a clone

gakonst: yeah seems like we should not have `as_*` methods then, let's remove them and replace with `into_*`

ethers-core/src/types/transaction/eip2718.rs
gakonst: agree

gakonst (approved):
gg

PR #1285, opened 2022-05-19
feat: add paginated logs

### LINKED ISSUE #954: Pagination / Streaming for Event Logs

meetmangukiya:
Closes #954 

## Solution

Implement a new stream that breaks requested block range into smaller pages that are loaded one at a time and streamed to the user.

Note: this only solves block limitations, if an rpc provider limits to 1000 blocks for instance, you could set page size to 1000 to solve it. However, if the limitation is on the number of logs(infura), this can't solve it automatically or completely, can try smaller page sizes but because number of logs in arbitrary block ranges is also arbitrary it can be hard to decide on a page size.  

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -18,6 +18,9 @@ pub use pending_transaction::PendingTransaction;
 mod pending_escalator;
 pub use pending_escalator::EscalatingPending;
 
+mod log_query;
+pub use log_query::LogQuery;
+
 mod stream;
 pub use futures_util::StreamExt;
 pub use stream::{interval, FilterWatcher, TransactionStream, DEFAULT_POLL_INTERVAL};
@@ -419,6 +422,15 @@ pub trait Middleware: Sync + Send + Debug {
 
     async fn get_logs(&self, filter: &Filter) -> Result<Vec<Log>, Self::Error> {
         self.inner().get_logs(filter).await.map_err(FromErr::from)
+    }
+
+    /// Returns a stream of logs are loaded in pages of given page size
+    fn get_logs_paginated<'a>(
+        &'a self,
+        filter: &Filter,
+        page_size: u64,
+    ) -> LogQuery<'a, Self::Provider> {
+        self.inner().get_logs_paginated(filter, page_size)
     }
 
     async fn new_filter(&self, filter: FilterKind<'_>) -> Result<U256, Self::Error> {
--- a/ethers-providers/src/log_query.rs
+++ b/ethers-providers/src/log_query.rs
@@ -1,0 +1,132 @@
+use super::{JsonRpcClient, Middleware, PinBoxFut, Provider};
+use ethers_core::types::{Filter, Log, U64};
+use futures_core::stream::Stream;
+use std::{
+    collections::VecDeque,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+pub struct LogQuery<'a, P> {
+    provider: &'a Provider<P>,
+    filter: Filter,
+    from_block: Option<U64>,
+    page_size: u64,
+    current_logs: VecDeque<Log>,
+    last_block: Option<U64>,
+    state: LogQueryState<'a>,
+}
+
+enum LogQueryState<'a> {
+    Initial,
+    LoadLastBlock(PinBoxFut<'a, U64>),
+    LoadLogs(PinBoxFut<'a, Vec<Log>>),
+    Consume,
+}
+
+impl<'a, P> LogQuery<'a, P>
+where
+    P: JsonRpcClient,
+{
+    pub fn new(provider: &'a Provider<P>, filter: &Filter) -> Self {
+        Self {
+            provider,
+            filter: filter.clone(),
+            from_block: filter.get_from_block(),
+            page_size: 10000,
+            current_logs: VecDeque::new(),
+            last_block: None,
+            state: LogQueryState::Initial,
+        }
+    }
+
+    /// set page size for pagination
+    pub fn with_page_size(mut self, page_size: u64) -> Self {
+        self.page_size = page_size;
+        self
+    }
+}
+
+macro_rules! rewake_with_new_state {
+    ($ctx:ident, $this:ident, $new_state:expr) => {
+        $this.state = $new_state;
+        $ctx.waker().wake_by_ref();
+        return Poll::Pending
+    };
+}
+
+impl<'a, P> Unpin for LogQuery<'a, P> {}
+
+impl<'a, P> Stream for LogQuery<'a, P>
+where
+    P: JsonRpcClient,
+{
+    type Item = Log;
+
+    fn poll_next(mut self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        match &mut self.state {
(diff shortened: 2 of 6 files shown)


ethers-providers/src/lib.rs
```diff
@@ -421,6 +424,15 @@ pub trait Middleware: Sync + Send + Debug {
...
     }
 
+    /// Returns a stream of logs are loaded in pages of given page size
+    fn get_logs_paginated<'a>(
+        &'a self,
+        filter: &Filter,
+        page_size: u64,
+    ) -> LogQuery<'a, Self::Provider> {
+        self.inner().get_logs_paginated(filter, page_size)
+    }
```
gakonst, line 434: @prestwich wdyt about unifying this logic in the default get_logs behavior?

gakonst (commented):
looks good - i think we need a test to ensure it works appropriately, either with a local node or with a large mainnet/testnet query?

meetmangukiya:
> looks good - i think we need a test to ensure it works appropriately, either with a local node or with a large mainnet/testnet query?

Ah actually i tested with an example seems like i missed committing.

ethers-providers/src/log_query.rs
```diff
@@ -0,0 +1,132 @@
...
+
+macro_rules! rewake_with_new_state {
+    ($ctx:ident, $this:ident, $new_state:expr) => {
+        $this.state = $new_state;
+        $ctx.waker().wake_by_ref();
+        return Poll::Pending
+    };
+}
+
+impl<'a, P> Unpin for LogQuery<'a, P> {}
```
mattsse, line 58: is this sound? adding a type bound to P should work

ethers-providers/src/log_query.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub fn with_page_size(mut self, page_size: u64) -> Self {
+        self.page_size = page_size;
+        self
+    }
+}
+
+macro_rules! rewake_with_new_state {
+    ($ctx:ident, $this:ident, $new_state:expr) => {
+        $this.state = $new_state;
+        $ctx.waker().wake_by_ref();
```
mattsse, line 53: can we make this work without continuously waking the waker? because this will immediately re-poll the query

mattsse (changes requested):

ethers-providers/src/lib.rs
prestwich: i think it's a pretty reasonable way to smooth over inconsistent RPC paging requirements 🤔

ethers-providers/src/log_query.rs
prestwich: this macro should basically not be used unless the new state blocks on another future

meetmangukiya: yeah only using it in case of polling another future

meetmangukiya: I suppose, could also just set the new state and return `Poll::Pending`?

ethers-providers/src/log_query.rs
meetmangukiya: It should be, right? since we are storing a ref to provider, not a raw pointer?

meetmangukiya: Removing this impl, seems unnecessary

ethers-providers/src/log_query.rs
mattsse: > yeah only using it in case of polling another future

got it, that works!

gakonst (approved):
LGTM. Separately, I wonder if we should unify this with `get_logs`. Would be a breaking change, requiring to change the function signature to return `LogQuery` instead of `Vec<Log>` but that could be navigable as it'd basically only introduce an additional `await` when iterating over the vector of logs.

Is there any benefit to using `get_log` over always using paginated? Probably not?

meetmangukiya:
Yeah, no real benefits of using `get_log`, didn't change it to not break things

gakonst:
Would give the follow-up a try, I think this is a good enough change that it's worth breaking, and we can include in the changelog

ethers-providers/src/log_query.rs
```diff
@@ -0,0 +1,130 @@
...
+
+                let filter = self.filter.clone().from_block(from_block).to_block(to_block);
+                let provider = self.provider;
+                // load first page of logs
+                let fut = Box::pin(async move { provider.get_logs(&filter).await });
+                rewake_with_new_state!(ctx, self, LogQueryState::LoadLogs(fut));
+            }
+            LogQueryState::LoadLogs(fut) => {
+                let logs = futures_util::ready!(fut.as_mut().poll(ctx))
+                    .expect("error occurred loading logs");
```
recmo, line 97: Panicking on failed RPCs is not very nice. I'd much prefer if this was a TryStream instead. Will submit PR when I get to it.

ethers-providers/src/log_query.rs
```diff
@@ -0,0 +1,130 @@
...
+                        // load new logs if there are still more pages to go through
+                        let from_block = self.from_block.unwrap();
+                        let to_block = from_block + self.page_size;
+
+                        // no more pages to load, and everything is consumed
+                        if from_block > self.last_block.unwrap() {
+                            return Poll::Ready(None)
+                        }
+                        // load next page
+                        self.from_block = Some(to_block);
```
philsippl, line 117: This should be `to_block + 1`, otherwise you double-sync blocks on the boundaries.

gakonst: Cc @meetmangukiya

meetmangukiya: will verify this

meetmangukiya: @philsippl is correct, will fix this

meetmangukiya: https://github.com/gakonst/ethers-rs/pull/1328/commits/86c349701b13dd2a248395a0b38e6094752d089f

ethers-providers/src/log_query.rs
meetmangukiya: Implemented `TryStream` along with other fixes here https://github.com/gakonst/ethers-rs/pull/1328/commits/0143cafc9f187feab49559e107e894c5725854e6

Issue #1227, opened 2022-05-05
Replace ganache with anvil


mattsse:
**Is your feature request related to a problem? Please describe.**
now that foundryup ships anvil, we can replace ganache

**Describe the solution you'd like**
- [ ] replace `Ganache` with `Anvil` in examples
- [ ] install `Anvil` in ci

**Describe alternatives you've considered**

**Additional context**

PR #1229, opened 2022-05-05
test: replace ganache with anvil

onbjerg:
Closes #1227

PR #1286, opened 2022-05-19
test: replace ganache with anvil

mattsse:
## Motivation
Closes https://github.com/gakonst/ethers-rs/issues/1227

supersedes #1229

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst:
I also think we may want to replace all `Geth` instances with `Anvil` in the tests? Found a mempool API which we do not implement in Anvil:

```
running 1 test
test txpool ... FAILED

failures:

---- txpool stdout ----
thread 'txpool' panicked at 'called `Result::unwrap()` on an `Err` value: JsonRpcClientError(JsonRpcError(JsonRpcError { code: -32601, message: "Method not found", data: None }))', ethers-providers/tests/txpool.rs:28:49
```

for the call `provider.txpool_status()`

gakonst:
I also think we may want to replace all `Geth` instances with `Anvil` in the tests? Found a mempool API which we do not implement in Anvil:

```
running 1 test
test txpool ... FAILED

failures:

---- txpool stdout ----
thread 'txpool' panicked at 'called `Result::unwrap()` on an `Err` value: JsonRpcClientError(JsonRpcError(JsonRpcError { code: -32601, message: "Method not found", data: None }))', ethers-providers/tests/txpool.rs:28:49
```

for the call `provider.txpool_status()`, supect it may apply to the `txpool_inspect()` too

gakonst:
@mattsse the commit above highlights some new issue in the mempool implementation I think

mattsse:
> @mattsse the commit above where I replaced Geth txpool with Anvil highlights some new issue in the mempool implementation I think

used the current gas price rather than the gas price set by the tx, fixed here https://github.com/foundry-rs/foundry/pull/1756 which should unblock this once included in nightly

PR #1287, opened 2022-05-20
return instead of ignore error

WyseNynja:
## Motivation

I noticed that occasionally my program would get stuck with one thread spinning at near 100% CPU. After a bunch of investigation with rust-gdb and tokio-console and stripping it down to a much [smaller program](https://github.com/SatoshiAndKin/web3-proxy/tree/c3d1d14f1628e415729e55809e3037b422ef16ad/web3-proxy-minimal), I narrowed it down to a websocket task being very active doing nothing.

Here you can see the spawned future stuck running:

IMAGE 2022-05-19 18:30:24

It is being woken up ~253k times per second, so that explains the 100% CPU:

IMAGE 2022-05-19 18:29:44

After some digging, I found this related code:

https://github.com/gakonst/ethers-rs/blob/3df1527cef7a7de8cbb26a8c98de74348203176b/ethers-providers/src/transports/ws.rs#L402-L408

Dropping that error is bad because it might be an `Io(Kind(UnexpectedEof))`. If that's the case, the websocket needs to reconnect.

## Solution

Quick fix is to just return an error when the socket gives an error.

I think a more robust fix might be to use something like https://docs.rs/stubborn-io/latest/stubborn_io/.

## PR Checklist

I'm not sure how to write a test for this.

- [ ] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-providers/src/transports/ws.rs
+++ b/ethers-providers/src/transports/ws.rs
@@ -401,8 +401,9 @@ where
             // Handle ws messages
             resp = self.ws.next() => match resp {
                 Some(Ok(resp)) => self.handle(resp).await?,
-                // TODO: Log the error?
-                Some(Err(_)) => {},
+                Some(Err(_)) => {
+                    return Err(ClientError::UnexpectedClose);
+                }
                 None => {
                     return Err(ClientError::UnexpectedClose);
                 },
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@
 - Add `TransactionReceipt::to` and `TransactionReceipt::from`
   [#1184](https://github.com/gakonst/ethers-rs/pull/1184)
 - Add `From<H160>` and From<Vec<H160>> traits to `ValueOrArray<H160>` [#1199](https://github.com/gakonst/ethers-rs/pull/1200)
+- Fix handling of Websocket connection errors [#1287](https://github.com/gakonst/ethers-rs/pull/1287)
 
 ## ethers-contract-abigen
 


gakonst:
Thanks for the in-depth investigation!

> Quick fix is to just return an error when the socket gives an error.

How does returning the error fix the CPU issue? Is this error a temporary error, which means that it made sense to ignore it? If we're terminating on a temporary error that doesn't seem right?

WyseNynja:
The error I'm seeing (and I *think* all of them), is not a temporary error. Once you get it, checking that future again always gives back an Eof immediately. The CPU issue is because it gets stuck in `tick` just getting an Eof back immediately every time it polls. Changing it to return an error causes the websocket thread to panic higher up in the code. I think there is another issue (or at least some discussion in telegram) to make that give a nicer error instead of panic.

Panic works fine for me though because I detect that my subscription exited and reconnect.

ethers-providers/src/transports/ws.rs
```diff
@@ -401,8 +401,9 @@ where
             // Handle ws messages
             resp = self.ws.next() => match resp {
                 Some(Ok(resp)) => self.handle(resp).await?,
-                // TODO: Log the error?
-                Some(Err(_)) => {},
+                Some(Err(_)) => {
+                    return Err(ClientError::UnexpectedClose);
```
meetmangukiya, line 405: maybe instead of dropping the error we can add a new field to enum `UnknownError(Err)` that wraps the error?

WyseNynja: That's a good idea. I'll try to get to that tomorrow.

WyseNynja:
with this PR, my proxy saw an error and reconnected:

image

Having the full error would be even better though.

ethers-providers/src/transports/ws.rs
```diff
@@ -401,8 +401,9 @@ where
             // Handle ws messages
             resp = self.ws.next() => match resp {
                 Some(Ok(resp)) => self.handle(resp).await?,
-                // TODO: Log the error?
-                Some(Err(_)) => {},
+                Some(Err(_)) => {
```
mattsse, line 404: that makes sense, but let's also log here the error

mattsse (changes requested):

ethers-providers/src/transports/ws.rs
gakonst: +1 let's log the error and otherwise looks good

PR #1292, opened 2022-05-20
fix(solc): flatten random statement order

rkrasiuk:
## Motivation

If any pragma statement is placed after the first import statement, the application will panic.

## Solution

Replace import and pragma statements in correct order to preserve the offset

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse (approved):
nice

gakonst (approved):

PR #1293, opened 2022-05-20
Linkable and DynamicLinkable

meetmangukiya:
Related https://github.com/foundry-rs/foundry/issues/975

## Todos

`Linkable` is to be implemented for:

- [x] `Artifact`: A proxy for linking the artifact's bytecodes
- [x] `Bytecode`: Links the bytecode directly
- [x] `CompactBytecode`: Links the bytecode directly
- [x] `BytecodeObject`: Links the bytecode if it's unlinked
- [x] `Contract`: A proxy for linking the contract's bytecodes
- [x] `CompactContract`: A proxy for linking the contract's bytecodes
- [x] `ProjectCompileOutput`: A proxy for linking many artifacts at the same time
- [x] `Artifacts`: A proxy for linking many artifacts at the same time

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


ethers-solc/src/artifacts/bytecode.rs
```diff
@@ -15,21 +15,15 @@ pub struct Bytecode {
...
-    pub source_map: Option<String>,
     /// Array of sources generated by the compiler. Currently only contains a
     /// single Yul file.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub generated_sources: Vec<GeneratedSource>,
-    /// If given, this is an unlinked object.
-    #[serde(default)]
-    pub link_references: BTreeMap<String, BTreeMap<String, Vec<Offsets>>>,
+    #[serde(default, flatten)]
+    pub compact_bytecode: CompactBytecode,
```
mattsse, line 26: this is a great change!

mattsse (commented):
as discussed, very much in favor of this change

gakonst (commented):
Love the direction

ethers-solc/src/artifact_output/mod.rs
```diff
@@ -475,7 +529,7 @@ where
 /// relationship (1-N+).
 pub trait ArtifactOutput {
     /// Represents the artifact that will be stored for a `Contract`
-    type Artifact: Artifact + DeserializeOwned + Serialize + fmt::Debug;
+    type Artifact: Artifact + DeserializeOwned + Serialize + fmt::Debug + Linkable;
```
mattsse, line 532: can we instead make `trait Artifact: Linkable` I think that'd be more fitting

ethers-solc/src/artifacts/mod.rs
```diff
@@ -1828,6 +1829,81 @@ impl SourceFiles {
...
+pub type LinkerOutput = BTreeMap<ArtifactId, Dependencies>;
+
+/// A linker function that is provided an [Artifact], a library's [Artifact],
+/// the `index` of the library denoting the deployment order
+/// for the library in respect to the provided contract.
+///
+/// The function should return the address at which the library is deployed.
+pub trait LinkerFn:
+    FnMut((&ArtifactId, &dyn Artifact), (&ArtifactId, &dyn Artifact), usize) -> Option<Address>
+{
```
mattsse, line 1897: this looks pretty wild

not sure I understand what that's supposed to do

can't we do this

`trait Linker { fn link(....)}`

?

ethers-solc/src/compile/output/mod.rs
```diff
@@ -29,6 +31,122 @@ pub struct ProjectCompileOutput<T: ArtifactOutput = ConfigurableArtifacts> {
...
+    }
+
+    fn is_unlinked(&self) -> bool {
+        // if any of the compiled or cached artifacts are not linked, the project is not linked
+        self.compiled_artifacts.is_unlinked() || self.cached_artifacts.is_unlinked()
+    }
+}
+type DependencyTree<T> = BTreeMap<T, Vec<T>>;
+
+fn link_dep<T: std::cmp::Ord + std::clone::Clone + std::fmt::Display>(
```
mattsse, line 51: needs docs

mattsse (changes requested):
nice progress,

some linker abstractions need some additional work I think

unclear at this point how this would be integrated in foundry

ethers-solc/src/artifacts/mod.rs
onbjerg: it's needed (or something like it) for e.g. `forge test` where we basically just want to get the artifact that should be linked, what library we are linking with, and what order it has. using this we can compute the address of the library before it is deployed

if we don't have something like this we have to keep pretty much all linker logic we have in foundry rn

ethers-solc/src/compile/output/mod.rs
```diff
@@ -471,3 +594,23 @@ impl<'a> fmt::Display for OutputDiagnostics<'a> {
...
+        dependency_tree.insert("Contract", vec!["Library", "ThirdLibrary"]);
+        dependency_tree.insert("Library", vec!["SecondLibrary"]);
+        dependency_tree.insert("SecondContract", vec!["Library"]);
+        dependency_tree.insert("ThirdContract", vec!["ThirdLibrary", "FourthLibrary"]);
+        dependency_tree.insert("FourthLibrary", vec!["Library"]);
+        dependency_tree.insert("SecondLibrary", vec![]);
+        dependency_tree.insert("ThirdLibrary", vec![]);
+
+        let link_order = get_link_order(&dependency_tree);
+        assert_eq!(link_order, vec!["SecondLibrary", "Library", "ThirdLibrary"]);
```
onbjerg, line 614: would be better if this was `Vec<(target, lib)>` so we can just iterate over those pairs, rn we iterate over all artifacts in the project in `link_all_dynamic` even if they dont need to be linked

PR #1294, opened 2022-05-20
relax thiserror semver patch number constraint

rakita:
Relax semver constraint in `thiserror` patch number so that it can mingle better with crates that want to use it.


mattsse (approved):
perhaps we should consider doing this for all of the major crates

rakita:
In most cases it is desirable. and the dependency bot is still going to spam you about the patch version update :)

Issue #1291, opened 2022-05-20
Add Backoff/Retry functionality to Http Transport


mattsse:
we're encountering more and more test failures on foundry due to ratelimited requests that look like

```console
thread 'fork::can_deploy_greeter_on_fork' panicked at 'called `Result::unwrap()` on an `Err` value: JsonRpcClientError(JsonRpcError(JsonRpcError { code: 429, message: "Your app has exceeded its compute units per second capacity. If you have retries enabled, you can safely ignore this message. If not, check out https://docs.alchemyapi.io/guides/rate-limits", data: None }))', /home/runner/work/foundry/foundry/anvil/src/config.rs:478:57

```

this is alchemyapi https://docs.alchemyapi.io/guides/rate-limits

the request should succeed if we just wait a little, it'd be nice if the Http Provider/Transport had an option to configure this with a backoff interval or something.

@meetmangukiya would this be something you would like to take on?

cc @oliver-giersch

meetmangukiya:
Can look into this next

oliver-giersch:
I think a `RateAwareHttp` or something should be rather easy to realize. I think I'd give it a `min_delay` and (optional) `max_delay` duration and internally keep track of a deadline before sending out any requests. Only ugly part is detecting rate limit errors. The  429 error code looks rather unique, but I bet each provider uses its own code.

danielattilasimon:
I started following this issue because we're also currently being hit by Alchemy rate limits when running Foundry in forking mode.

> Only ugly part is detecting rate limit errors. The 429 error code looks rather unique, but I bet each provider uses its own code.

FWIW status 429 is at least somewhat standard:
https://www.rfc-editor.org/rfc/rfc6585#section-4

In addition to Alchemy, it is also used by Infura to signal being rate limited:
https://docs.infura.io/infura/networks/ethereum/how-to/avoid-rate-limiting#noticing-rate-limiting-behavior

Therefore, IMHO it's not necessarily that ugly to handle 429 by backing off then retrying. If there _is_ a provider out there that doesn't use 429, now _that_ would be really ugly to detect.

oliver-giersch:
Ok, in this case I would propose something simple like the code below.

This is using the #1267 PR I am currently working on and I would just include it in that PR, if that is Ok with everyone.
[...]

PR #1302, opened 2022-05-23
feat(providers): add `RetryClient`

meetmangukiya:
## Motivation

https://github.com/gakonst/ethers-rs/issues/1291

## Solution

Instead of relying on non-standard status codes and errors for rate limiting, maintain the current state of rate limits and enforce the limits on the client side such that the rate limit is never hit on the server side. This requires for the client to know the rate limit i.e. the `quota` that renews every `duration` amount of time. 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse:
> This requires for the client to know the rate limit i.e. the quota that renews every duration amount of time.

this info is not always available and will also result in errors if the server simply decides we're getting rate limited even if we follow the limits.
adding support for retries should be more convenient

gakonst (changes requested):
Yeah not sure about this approach, we cannot really always know the rate limit. Agree with @mattsse comment that the 'right' way is some exponential backoff, or some buffering of requests to be sent at a steady rate.

Maybe as a general point, let's try to discuss/align on the 'right' solution in the issue, before we move on to a PR, so we don't spend cycles on things we do not agree yet?

meetmangukiya:
So one of the challenges while implementing backoff retry or anything that would require us to make multiple rpc calls wouldn't work unless we make `param` either `Copy` or `Clone` which would require bounding `JsonRpcClient`'s `request` further and would be a breaking change.

```
error[E0276]: impl has stricter requirements than trait
  --> ethers-providers/src/transports/retry.rs:73:56
   |
73 |           A: std::fmt::Debug + Serialize + Send + Sync + Copy,
   |                                                          ^^^^ impl has extra requirement `A: std::marker::Copy`
   |
  ::: ethers-providers/src/lib.rs:63:5
   |
63 | /     async fn request<T, R>(&self, method: &str, params: T) -> Result<R, Self::Error>
64 | |     where
65 | |         T: Debug + Serialize + Send + Sync,
66 | |         R: DeserializeOwned;
   | |____________________________- definition of `request` from trait
```

ethers-providers/src/transports/retry.rs
```diff
@@ -0,0 +1,93 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+use thiserror::Error;
+
+use std::{clone::Clone, fmt::Debug, sync::atomic::AtomicU64};
+
+#[derive(Debug)]
+pub struct RetryClient<T: JsonRpcClient, U: RetryPolicy<T::Error> + Send + Sync + Debug> {
+    inner: T,
+    requests_enqueued: AtomicU64,
+    policy: U,
```
mattsse, line 16: I'd like to make this Box<dyn> to clean up generics

ethers-providers/src/transports/retry.rs
```diff
@@ -0,0 +1,93 @@
...
+    backoff: ExponentialBackoff,
+}
+
+pub trait RetryPolicy<E> {
```
mattsse, line 20: let's add `+ Send +  Sync` here

ethers-providers/src/transports/retry.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    async fn request<A, R>(&self, method: &str, params: A) -> Result<R, Self::Error>
+    where
+        A: std::fmt::Debug + Serialize + Send + Sync + Copy,
+        R: DeserializeOwned,
+    {
+        let mut backoff = self.backoff.clone();
+        loop {
+            let ret: Result<R, T::Error> = self.inner.request(method, params).await;
+            if ret.is_err() {
```
mattsse, line 79: this should probably a match statement

ethers-providers/src/transports/retry.rs
```diff
@@ -0,0 +1,93 @@
...
+where
+    T: JsonRpcClient + 'static,
+    T::Error: Sync + Send + 'static,
+    U: RetryPolicy<T::Error> + Send + Sync + Debug,
+{
+    type Error = RetryClientError<T>;
+
+    async fn request<A, R>(&self, method: &str, params: A) -> Result<R, Self::Error>
+    where
+        A: std::fmt::Debug + Serialize + Send + Sync + Copy,
```
mattsse, line 73: making this copy will cause a ton of problems.

ethers-providers/src/transports/retry.rs
```diff
@@ -0,0 +1,93 @@
...
+        R: DeserializeOwned,
+    {
+        let mut backoff = self.backoff.clone();
+        loop {
+            let ret: Result<R, T::Error> = self.inner.request(method, params).await;
```
mattsse, line 78: you can serialize the params once to a json and use the reference as params

ethers-providers/src/transports/retry.rs
```diff
@@ -0,0 +1,93 @@
...
+{
+    type Error = RetryClientError<T>;
+
+    async fn request<A, R>(&self, method: &str, params: A) -> Result<R, Self::Error>
+    where
+        A: std::fmt::Debug + Serialize + Send + Sync + Copy,
+        R: DeserializeOwned,
+    {
+        let mut backoff = self.backoff.clone();
+        loop {
```
mattsse, line 77: this doesn't actually backoff currently, right?
there's an async example https://github.com/ihrwein/backoff/blob/master/examples/async.rs

but I suppose we can also solve this via loop and sleep this also gives us more flexibility with the `policy`

ethers-providers/src/transports/retry.rs
```diff
@@ -0,0 +1,257 @@
...
+                let next_backoff = self
+                    .backoff
+                    .lock()
+                    .unwrap()
+                    .next_backoff(self.requests_enqueued.load(Ordering::SeqCst), retry_number);
+                if next_backoff.is_none() {
+                    return Err(RetryClientError::TimeoutError)
+                }
+                let next_backoff = next_backoff.unwrap();
+                Delay::new(next_backoff).await;
```
mattsse, line 161: there should be `tokio::time::sleep()`

ethers-providers/src/transports/retry.rs
```diff
@@ -0,0 +1,257 @@
...
+where
+    T: JsonRpcClient,
+    T::Error: Sync + Send + 'static,
+    U: RetryPolicy<T::Error> + Send + Sync + Debug,
+    B: Backoff + Send + Sync,
+{
+    inner: T,
+    requests_enqueued: AtomicU32,
+    policy: U,
+    backoff: Mutex<B>,
```
mattsse, line 53: not a big fan of the mutex but perhaps the easiest way to make this work.
wdyt @gakonst 

nevertheless we should use the tokio::sync::Mutex here

ethers-providers/src/transports/retry.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    async fn request<A, R>(&self, method: &str, params: A) -> Result<R, Self::Error>
+    where
+        A: std::fmt::Debug + Serialize + Send + Sync,
+        R: DeserializeOwned,
+    {
+        self.requests_enqueued.fetch_add(1, Ordering::SeqCst);
+
+        let params =
+            serde_json::to_value(params).map_err(|err| RetryClientError::SerdeJson(err))?;
```
mattsse, line 131: this works,
but now that I see it, I believe serialize to String would be more efficient because then we don't have to serialize Value again and again

ethers-providers/src/transports/retry.rs
```diff
@@ -0,0 +1,257 @@
...
+            if should_retry {
+                let next_backoff = self
+                    .backoff
+                    .lock()
+                    .unwrap()
+                    .next_backoff(self.requests_enqueued.load(Ordering::SeqCst), retry_number);
+                if next_backoff.is_none() {
+                    return Err(RetryClientError::TimeoutError)
+                }
+                let next_backoff = next_backoff.unwrap();
```
mattsse, line 160: let's do a `ok_or(err)?` here to get rid of the unwrap

mattsse (changes requested):
getting there :)

ethers-providers/src/transports/retry.rs
mattsse: I see, I'm actually not sure how the differ, so should be fine

ethers-providers/src/transports/retry.rs
meetmangukiya: it does now

ethers-providers/src/transports/retry.rs
```diff
@@ -0,0 +1,176 @@
...
+            // hack to not hold `R` across an await in the sleep future and prevent requiring
+            // R: Send + Sync
+            {
+                let ret = self.inner.request(method, params.clone()).await;
+                if let Ok(ret) = ret {
+                    self.requests_enqueued.fetch_sub(1, Ordering::SeqCst);
+                    return Ok(ret)
+                }
+
+                err = ret.err().unwrap();
```
mattsse, line 135: nit: let's replace this with a match statement

mattsse (commented):
let's try this on foundry to ensure it works?

meetmangukiya:
@mattsse https://github.com/foundry-rs/foundry/runs/6608273536?check_suite_focus=true

ethers-providers/src/transports/retry.rs
```diff
@@ -126,13 +126,10 @@ where
             // hack to not hold `R` across an await in the sleep future and prevent requiring
             // R: Send + Sync
             {
-                let ret = self.inner.request(method, params.clone()).await;
-                if let Ok(ret) = ret {
-                    self.requests_enqueued.fetch_sub(1, Ordering::SeqCst);
-                    return Ok(ret)
+                match self.inner.request(method, params.clone()).await {
+                    Ok(ret) => return Ok(ret),
```
mattsse, line 130: `self.requests_enqueued.fetch_sub(1, Ordering::SeqCst);`

mattsse:
> @mattsse https://github.com/foundry-rs/foundry/runs/6608273536?check_suite_focus=true

very nice, lots of unrelated failing tests but no rate limit errors

also ref https://github.com/foundry-rs/foundry/pull/1728

ethers-providers/src/transports/retry.rs
meetmangukiya: whoops

ethers-providers/src/transports/retry.rs
```diff
@@ -0,0 +1,176 @@
...
+
+            let should_retry = self.policy.should_retry(&err);
+            if should_retry {
+                let current_queued_requests = self.requests_enqueued.load(Ordering::SeqCst);
+                // using `retry_number + current_queued_requests` for creating back pressure because
+                // of already queued requests
+                let next_backoff =
+                    self.initial_backoff * 2u64.pow(retry_number + current_queued_requests);
+                tokio::time::sleep(Duration::new(next_backoff, 0)).await;
+                continue
```
mattsse, line 151: this is redundant, right?

mattsse (commented):
this looks ready now?

The last suggestion would be to serialize to string instead of `serde_json::Value` which prevents repeated serialization effort

meetmangukiya:
> The last suggestion would be to serialize to string instead of serde_json::Value which prevents repeated serialization effort

if we serialize it to string, it actually gets serialized as a string further down when the json gets serialized, so instead of getting `{"params":["latest",null]}` we'd get `{"params":"[l\"latest\",null]"}`

mattsse:
> > The last suggestion would be to serialize to string instead of serde_json::Value which prevents repeated serialization effort
> 
> if we serialize it to string, it actually gets serialized as a string further down when the json gets serialized, so instead of getting `{"params":["latest",null]}` we'd get `{"params":"[l\"latest\",null]"}`

ah that makes sense, disregard that then!

mattsse (approved):

gakonst (approved):
Very supportive - let's see how this works out in practice.

PR #1308, opened 2022-05-25
fix(signer): use wallet chainid for tx signing

Rjected:
## Motivation

In `sign_transaction_sync`, if a transaction `chain_id` was `None`, the wallet would sign a transaction without serializing the chain id. However, the signature's `v` would encode the proper chain.

## Solution

Use the signer's `chain_id` if the transaction does not specify one, before hashing the transaction in `sign_transaction_sync`.

Adds a test checking that `sign_transaction_sync` outputs a correct `v` when signing a transaction with an empty chain.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
I don't think this is right now a bottleneck, but we should be careful with all these `clone`'s on transactions

PR #1316, opened 2022-05-27
fix: wrong unit for gas_price (ether -> gwei)

naiba:


gakonst (commented):
@recmo wdyt

gakonst (approved):
ok this seems fine - just realized that polygon returns gwei and not wei - cc @recmo

PR #1317, opened 2022-05-27
feat(solc): use relative paths and --base-path option

### LINKED ISSUE #1315: Use solc --base-path to project root by default
currently, source files in the compilerinput are canonic, this can cause some issues with metadata/hashes.

in #1307 we already save the artifacts with relative files, this should be obsolete if we set base-path == root and strip the root from all sources

mattsse:
## Motivation
Closes #1315

working with paths is hard.

Ideally, we want to work with absolute paths because we resolve them in the Graph anyway and `Project::root` is not guaranteed to be cwd.

however, absolute/relative paths result in different solc metadata, so when invoking solc, we want to be consistent, or rather OS and local disk independent.

with 0.6.9, solc supports a `--base-path` option so we use that instead and also set the cwd of the solc process to `Project::root`.

This also introduces the `CompilerInput::with_base_path` function that will strip `Project::root` from:
* source paths
* output selection
* remappings 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,8 @@
 
 ### Unreleased
 
+- Use relative source paths and `solc --base-path`
+  [#1317](https://github.com/gakonst/ethers-rs/pull/1317)
 - Save cache entry objects with relative paths
   [#1307](https://github.com/gakonst/ethers-rs/pull/1307)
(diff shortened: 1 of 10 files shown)


gakonst (approved):
good call

PR #1329, opened 2022-05-31
fix(solc): remove compile_exact restriction

mattsse:
## Motivation
Ref https://github.com/foundry-rs/foundry/issues/1776

this removes the `compile_exact` restriction, which after some investigation can cause some issues in following circumstances:
* paths (or keys) of the contracts in the output differ from the input keys on windows
* if we failed to resolve the complete graph properly (likely due to some parser limitations) this can filter out those contracts we missed but solc includes in the output.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
cool - let's remove `compile_exact` in a follow up if it's non longer necessary

Issue #867, opened 2022-02-05
abigen: ambiguous import of EthCall for inheriting contracts


sveitser:
**Version**
```
├── ethers v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc)
│   ├── ethers-addressbook v0.1.0 (https://github.com/gakonst/ethers-rs#b2b891dc)
│   │   ├── ethers-core v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc)
│   ├── ethers-contract v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc)
│   │   ├── ethers-contract-abigen v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc)
│   │   │   ├── ethers-core v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc) (*)
│   │   ├── ethers-contract-derive v0.6.0 (proc-macro) (https://github.com/gakonst/ethers-rs#b2b891dc)
│   │   │   ├── ethers-contract-abigen v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc)
│   │   │   │   ├── ethers-core v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc)
│   │   │   ├── ethers-core v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc) (*)
│   │   ├── ethers-core v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc) (*)
│   │   ├── ethers-providers v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc)
│   │   │   ├── ethers-core v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc) (*)
│   ├── ethers-core v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc) (*)
│   ├── ethers-etherscan v0.2.0 (https://github.com/gakonst/ethers-rs#b2b891dc)
│   │   ├── ethers-core v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc) (*)
│   │   ├── ethers-solc v0.2.0 (https://github.com/gakonst/ethers-rs#b2b891dc)
│   │   │   ├── ethers-core v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc) (*)
│   ├── ethers-middleware v0.6.0 (https://github.com/gakonst/ethers-rs#b2b891dc)
[...]

PR #1332, opened 2022-05-31
make abigen-ed modules pub

tri-joe:
## Motivation

Allows end-users to work around https://github.com/gakonst/ethers-rs/issues/867

## Solution

This makes it possible to refer directly to generated types when they have the same name (as occurs with event filter types when one contract inherits another).

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (commented):
👍

tomekozlowski:
I’ve updated ethers and still getting this `error[E0616]: field `0` of struct `iuniswapv2pair_mod::IUniswapV2Pair` is private`

PR #1338, opened 2022-06-02
feat(solc): emit build info files if configured

mattsse:
## Motivation
Emit build-info files that include the whole compiler(input, output) and version.

## Solution

* Create bindings for `BuildInfo`
* If configured in the project we create a `BuildInfo` instance for each solc invocation.
* build-infos are stored as `<root>/build-info/<solc>-<timestamp>.json`, this differs from hardhat which uses the content hash as file name, like `0.8.14+commit.80d49f37.Darwin.appleclang-2022-06-02T12:17:16.944469+00:00.json`

wdyt @tynes ?

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -449,6 +449,7 @@ dependencies = [
  "num-integer",
  "num-traits",
  "serde",
+ "time 0.1.43",
  "winapi",
 ]
 
@@ -1403,6 +1404,7 @@ name = "ethers-solc"
 version = "0.3.0"
 dependencies = [
  "cfg-if 1.0.0",
+ "chrono",
  "colored",
  "criterion",
  "dunce",
@@ -3822,6 +3824,16 @@ dependencies = [
 
 [[package]]
 name = "time"
+version = "0.1.43"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ca8a50ef2360fbd1eeb0ecd46795a87a19024eb4b53c5dc916ca1fd95fe62438"
+dependencies = [
+ "libc",
+ "winapi",
+]
+
+[[package]]
+name = "time"
 version = "0.3.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "004cbc98f30fa233c61a38bc77e96a9106e65c88f2d3bef182ae952027e5753d"
@@ -4520,7 +4532,7 @@ dependencies = [
  "signature",
  "subtle",
  "thiserror",
- "time",
+ "time 0.3.7",
  "uuid",
  "zeroize",
 ]
--- a/ethers-solc/Cargo.toml
+++ b/ethers-solc/Cargo.toml
@@ -40,6 +40,7 @@ rayon = "1.5.3"
 rand = { version = "0.8.5", optional = true }
 path-slash = "0.1.4"
 cfg-if = "1.0.0"
+chrono = "0.4.19"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 home = "0.5.3"
--- a/ethers-solc/src/artifacts/serde_helpers.rs
+++ b/ethers-solc/src/artifacts/serde_helpers.rs
@@ -78,7 +78,7 @@ pub mod json_string_opt {
 pub mod empty_json_object_opt {
     use serde::{
         de::{self, DeserializeOwned},
-        ser, Deserialize, Deserializer, Serialize, Serializer,
+        Deserialize, Deserializer, Serialize, Serializer,
     };
 
     pub fn serialize<T, S>(value: &Option<T>, serializer: S) -> Result<S::Ok, S::Error>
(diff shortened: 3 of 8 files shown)


gakonst:
Do we see this as impacting performance?

mattsse:
it's definitely an overhead because a buildinfo object is in the MB and can be 50-100MB or even higher for large projects, so perhaps we should make this opt in via config val in foundry.toml?

gakonst:
Yeah I'd disable it by default given we only really want it for hardhat-like integrations and only enable via foundry toml / CLI. Are there other use cases beyond Hardhat that we'd want it for? I think @tynes is the main feedback source here?

tynes:
> a buildinfo object is in the MB and can be 50-100MB or even higher for large projects

If we do it based on timestamp, this could use a lot of space really quickly. The main consideration is we need to be able to go from qualified name -> build info file. If we use the timestamp methodology, the hh plugin can sort by timestamp and take the most recent buildinfo file. A way around using a lot of disk would be to have foundry delete the buildinfo file if one exists before writing the next one.

hh maps a bulidinfo + solc version to a hash and uses that as the filename, see: https://github.com/NomicFoundation/hardhat/blob/783650c5711f457ad63cc7eee9ac190d0ff264fc/packages/hardhat-core/src/internal/artifacts.ts#L244
This way the same input results in the same buildinfo file. 

It looks like hh solves being able to look up qualified name -> buildinfo file by keeping a json debug file for each artifact that looks like this:

```ts
    const debugFile: DebugFile = {
      _format: DEBUG_FILE_FORMAT_VERSION,
      buildInfo: relativePathToBuildInfo,
    };
```
https://github.com/NomicFoundation/hardhat/blob/783650c5711f457ad63cc7eee9ac190d0ff264fc/packages/hardhat-core/src/internal/artifacts.ts#L295
The name of the debug file is simply the name of the artifact but with `.dbg.json`, see https://github.com/NomicFoundation/hardhat/blob/783650c5711f457ad63cc7eee9ac190d0ff264fc/packages/hardhat-core/src/internal/artifacts.ts#L580

It seems like the options are:
- have a 1:1 mapping of buildinfo files based on what was compiled and ensure that can be dealt with in the plugin
  - the plugin currently doesn't have access to the version of `solc` used
- create a buildinfo each time and prune the old one

I'm ok with either of these solutions, the first seems like the "better" solution as it would save a bunch of i/o but also more complex. 

Note that hh does prune old buildinfo files: https://github.com/NomicFoundation/hardhat/blob/783650c5711f457ad63cc7eee9ac190d0ff264fc/packages/hardhat-core/src/internal/artifacts.ts#L221

> Yeah I'd disable it by default given we only really want it for hardhat-like integrations and only enable via foundry toml / CLI. Are there other use cases beyond Hardhat that we'd want it for? I think @tynes is the main feedback source here?

There are other people that have asked for access to the compiler input/output directly according to the tg group, it should def be gated by a flag in the `foundry.toml`. Another usecase would be `cat artifacts/buildinfo/file.json | jq .compilerOutput | abigen --combined-json` to build golang bindings and I'm sure other existing tooling (slither?) could take advantage of it

mattsse:
> f we do it based on timestamp, this could use a lot of space really quickly. The main consideration is we need to be able to go from qualified name -> build info file. If we use the timestamp methodology, the hh plugin can sort by timestamp and take the most recent buildinfo file. A way around using a lot of disk would be to have foundry delete the buildinfo file if one exists before writing the next one.

I used the timestamp due to lack of understanding how the `buildInfo Name` is created by hh, I guess it's a hash of the `buildInfo` content?

> This way the same input results in the same buildinfo file.

right, I deemed that to be too much overhead, but makes sense now.

> The main consideration is we need to be able to go from qualified name -> build info file.

ah now, this makes sense :)

> I'm ok with either of these solutions, the first seems like the "better" solution as it would save a bunch of i/o but also more complex.

agreed, let's do option 1.

I think what's missing is to get the build info file at least is just

- [ ] content hashing -> filename

The additional dbg.json would also be doable but would require some followup work

tynes:
> I used the timestamp due to lack of understanding how the `buildInfo Name` is created by hh, I guess it's a hash of the `buildInfo` content?

It looks like its literally just the md5 of a json file:
- https://github.com/NomicFoundation/hardhat/blob/20d8dbc33e696bf8900de0221bbc465f8d91e593/packages/hardhat-core/src/internal/util/hash.ts#L10
- https://github.com/NomicFoundation/hardhat/blob/20d8dbc33e696bf8900de0221bbc465f8d91e593/packages/hardhat-core/src/internal/artifacts.ts#L249

mattsse:
file name is now the content hash of `{_format,solcVersion,solcLongVersion,input}` same as hh


still needs a followup for:

- [ ] top level dbg.json
- [ ] foundry companion

smartcontracts:
hardhat-deploy also relies on buildinfo files for etherscan verification

gakonst (approved):
mhm seems fine to use, glad it's off by default :)

tynes:
I can handle the hh plugin implementation of this once forge is updated with the latest ethers + a way to pass through config to trigger the buildinfo being written to disk

PR #1339, opened 2022-06-02
feat(provider): Add batching for JSON-RPC requests

frjnn:
The `BatchRequest` struct is added and it allows for sending multiple JSON-RPC requests in a single batch.

## Motivation

ethers-rs is currently lacking batching of multiple JSON-RPC requests. There are many use-cases for such functionality and hereby an implementation is proposed.

## Solution

A new struct, namely, `BatchRequest`, is introduced.  `BatchRequest` acts a simple, generic, container of JSON-RPC requests. 

Batches are created as follows:

```rust
use ethers::prelude::*;
// Creates a new, empty, batch.
let batch = BatchRequest::new();
```

If the user already knows the number of requests it is going to send, `BatchRequest::with_capacity` can be used instead:

```rust
use ethers::prelude::*;
// Creates a new, empty, batch, with a pre-allocated capacity of 10.
let batch = BatchRequest::with_capacity(10);
```

`BatchRequest` is agnostic of the underlying requests, both in terms of parameters and return types, as long as they can be serialised and deserialised. Adding JSON-RPC requests to batches is as simple as:

```rust
use ethers::prelude::*;

let address1: Address = "0xd5a37dC5C9A396A03dd1136Fc76A1a02B1c88Ffa".parse()?;
let address2: Address = "0x4d7450da3abc67663c6C3989B4c84F35e5D0B5fC".parse()?;

let pos = H256::from_low_u64_be(8);

let mut batch: BatchRequest = BatchRequest::with_capacity(2);
// All kinds of JSON-RPC requests can go here.
batch.add_request("eth_getStorageAt", (address1, pos, BlockNumber::Latest))?;
batch.add_request("eth_getStorageAt", (address2, pos, BlockNumber::Latest))?;
```

Furthermore, batches created with `BatchRequest` can be executed by providers. The same batch can be executed by manyfold providers, as shown in the following example:

```rust
use ethers::prelude::*;

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let address1: Address = "0xd5a37dC5C9A396A03dd1136Fc76A1a02B1c88Ffa".parse()?;
    let address2: Address = "0x4d7450da3abc67663c6C3989B4c84F35e5D0B5fC".parse()?;

    let pos = H256::from_low_u64_be(8);

    let mut batch = BatchRequest::with_capacity(2);
    batch.add_request("eth_getStorageAt", (address1, pos, BlockNumber::Latest))?;
    batch.add_request("eth_getStorageAt", (address2, pos, BlockNumber::Latest))?;

    let http_client = Provider::<Http>::try_from("https://api.avax.network/ext/bc/C/rpc")?;
    let ws_client = Provider::<Ws>::connect(
        "wss://speedy-nodes-nyc.moralis.io/bdc23feb8c3bd2b3e8d6a716/avalanche/mainnet/ws",
    )
    .await?;

    let mut ws_responses: BatchResponse = ws_client.as_ref().execute_batch(&mut batch).await?;
    let mut http_responses: BatchResponse = http_client.as_ref().execute_batch(&mut batch).await?;

    while let Some(Ok(storage)) = ws_responses.next_response::<H256>() {
        println!("{storage:?}")
    }

    while let Some(Ok(storage)) = http_responses.next_response::<H256>() {
        println!("{storage:?}")
    }

    Ok(())
}
```

out:

```text
0x6298ae7d00000000399dcf6078402dab87ab000000000000000000050ae7bd47
0x6298af7d0000000000d41171d7825e495f8e00000421fcd185b36af97a5abe6e
0x6298ae7d00000000399dcf6078402dab87ab000000000000000000050ae7bd47
0x6298af7d0000000000d41171d7825e495f8e00000421fcd185b36af97a5abe6e
```

As shown, batches can be re-used and cached.

The user gets back a `BatchResponse` struct, which implements the `next_response` method. Responses are retrieved and deserialised, as also shown.

I implemented a `.execute_batch()` method for the `Provider` struct, more precisely, such method is implemented for 3 transports: http, ws and icp. I did not put it in the `JsonRpcClient` trait, as to proceed in that direction I felt that someone else's opinion was needed.

Transports implementation needed only minor modifications to accommodate such change.

### Request and response ids

To match requests and responses in the transport layer, and to preserve the order of issued responses with respect to requests, the following solution is adopted, in order:

1. The provider's `id` is incremented by the length of the batch and fetched.
2. `BatchRequest` sequentially populates the ids of the inner requests starting from the latter id.
3. Such id is then attached to the batch instruction, which is sent by the provider to the inner transport.
4. `BatchResponse` sorts the responses in descending order with respect to their ids. 
5. The id of the first response in the batch is used to uniquely determine the channel to send the response into in the transport layer. 

It's important to note that we forbid the user from sending empty batches by raising `BatchError::EmptyBatch`.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Unreleased
 
-- Add batching of JSON-RPC requests via the addition of `BathRequest`.
+- Add batching of JSON-RPC requests via the addition of `BatchRequest`.
   [#1339](https://github.com/gakonst/ethers-rs/pull/1339)
 - Add `as_*_mut` methods on `TypedTransaction`
   [#1310](https://github.com/gakonst/ethers-rs/pull/1310)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Unreleased
 
+- Add batching of JSON-RPC requests via the addition of `BathRequest`.
+  [#1339](https://github.com/gakonst/ethers-rs/pull/1339)
 - Add `as_*_mut` methods on `TypedTransaction`
   [#1310](https://github.com/gakonst/ethers-rs/pull/1310)
 - AWS EIP712 data signing no longer signs with EIP155
--- a/ethers-providers/src/transports/common.rs
+++ b/ethers-providers/src/transports/common.rs
@@ -1,8 +1,9 @@
 // Code adapted from: https://github.com/althea-net/guac_rs/tree/master/web3/src/jsonrpc
-use std::fmt;
+
+use std::{boxed::Box, fmt};
 
 use serde::{
-    de::{self, MapAccess, Unexpected, Visitor},
+    de::{self, DeserializeOwned, MapAccess, Unexpected, Visitor},
     Deserialize, Serialize,
 };
 use serde_json::{value::RawValue, Value};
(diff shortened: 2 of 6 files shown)


mattsse (commented):
defer to @oliver-giersch for this to reduce friction for #1267

just skimmed it, but I believe this will be much easier with #1267 since `request_id` is now part of the Connection trait, right?


also would love to see some tests against `Anvil` which does support batch requests.

frjnn:
@mattsse I have yet to look at #1267.

gakonst:
As a process point, let's please open an issue + discuss the design before getting into opening a PR :) These are low level features, and it's unclear how they'd integrate with the higher level provider + middlewares. How would we integrate this with e.g. higher level abstractions @frjnn?

oliver-giersch:
I thought a bit about this issue, looking at it through the lens of #1267 (which I am still working on btw).

The solution by @frjnn looks generally workable, but suffers from a lot of duplication due to the  way the current `JsonRpcClient` trait is structured. It does also look a bit inconvenient to use, as the caller has to basically insert raw RPC requests and then manually select the correct response type when iterating over each batch response (the ordering by id is clever though!)
The JSON-RPC spec for batches is pretty permissive, though, and some edge cases seem to be not handled (off the top of my head):

- The server may send a single error response instead of an array of responses
- The server may respond with fewer response objects than sent request objects

Anyways, the revised `Connection` trait will probably need an additional method for sending batch requests. How this will look like depends on what the end user API will look like.

Also, here are some alternative approaches:

#### 1. Middleware

I personally hate the `Middleware` trait but this looks like a legit use case for it. It would require a ton of duplication and be cumbersome to implement (basically reimplement every RPC call to put it into a batch first) but be a bit more type-safe.
Response parsing would likely still have to be done by hand.

#### 2. "Meta" `Connection`

This is similar to the design in rust-web3. There would have to be a separate batch connection wrapper, which intercepts any requests from the higher-level `Provider` API and stores them in a batch first. Would look roughly like this:

```rust
let provider = Provider::connect("http://localhost:8545").await?;
let provider: Provider<Batch<_>> = Batch::wrap(provider);

let a = batch.get_block_by_number("earliest".into()); // no await!
let b = provider.get_block_by_number("latest".into());

// calling `a.await` before `provider.send_batch` would either block
// (potentially deadlock) or panic
provider.send_batch().await?

let res_a = a.await?; // both awaits resolve instantly
let res_b = b.await?;
```

Fully type-safe, but with a potential footgun, as the caller must not await any request futures before calling `send_batch`.

#### 3. `Batch` trait and intermediate `RpcCall` type

This one is the most involved tbh. Under the current revised API `Provider` methods look like this:

```rust
pub async fn get_block_by_hash(
    &self,
    hash: &H256,
) -> Result<Option<Block<H256>>, Box<ProviderError>> {
    self.send_request("eth_getBlockByHash", (hash, false)).await
}
```
This does not allow to externally access the raw request after the call. Instead, all methods could return a struct like

```rust
pub struct RpcCall<'a, C, R> {
    connection: &'a C,
    raw_request: (u64, Box<RawValue>),
    _marker: PhantomData<fn() -> R>,
}
```
This could then be either `call`ed and awaited directly or put into a batch like this:
```rust
let a = provider.get_block_by_hash(&hash1);
let b = provider.get_block_by_hash(&hash2);

let (a, b) = provider.send_batch((a, b)).await?;
```
Where `Provider::send_batch` takes a generic `Batch`, which would be implemented for tuples with a certain length (2-16 or so) of `RpcCall`s. The right types for the returned `a` and `b` would be selected automatically, although they probably would still have to be wrapped in an `Option` or a `Result` (or both...).

frjnn:
Thanks for the articulate review. I will quote some extracts.

> "(...) suffers from a lot of duplication due to the way the current JsonRpcClient trait is structured"

I fully agree. I am very much open to integrate it in #1267, should it be ok for you and for everyone here. From what I was able to seen in the actual transport implementation, the redundancy came from the way ethers-rs handles requests and responses.

>  "It does also look a bit inconvenient to use, as the caller has to basically insert raw RPC requests and then manually select the correct response type when iterating over each batch response."

I agree, nonetheless, that's as abstract as  the `JsonRpcClient` API allows things to be done right now. By calling `request` the user is performing the very same action, just only one method at a time. I do believe though that the `BatchRequest` struct would allow for some pretty dope higher level API.

> The JSON-RPC spec for batches is pretty permissive (...).

Thank you ver much for pointing this out, I am not very familiar with the caveats of the JSON-RPC protocol. Referring to [this specs](https://www.jsonrpc.org/specification#batch), to my understanding, the server will return a single error response, without id (i.e. null id) when the batch is either malformed or empty. The implementation in this PR covers both such events.
Quoting the spec I linked, I notice it says "A Response object SHOULD exist for each Request object (...)" are you referring to this? If so, how can we address this issue?

Also I would not suggest to stick with the `Middleware` trait neither. Out of the solutions you proposed, the one I like the most and that I think provides the user with the better experience is the number **2**.

oliver-giersch:
> I fully agree. I am very much open to integrate it in #1267, should it be ok for you and for everyone here. From what I was able to see in the current transport implementation, the redundancy came from the way ethers-rs handles requests and responses.

I am not entirely sure how this would be organized, but I think you could do a PR-in-PR if I gave you access to my fork of the repo and it eventually all gets merged as #1267. 🤔 

> Thank you ver much for pointing this out, I am not very familiar with the caveats of the JSON-RPC protocol. Referring to [this specs](https://www.jsonrpc.org/specification#batch), to my understanding, the server will return a single error response, without id (i.e. null id) when the batch is either malformed or empty. The implementation in this PR covers both such events. Quoting the spec I linked, I notice it says "A Response object SHOULD exist for each Request object (...)" are you referring to this? If so, how can we address this issue?

Sorry, I must have overlooked this, if it is already handled. But your summary sounds basically correct. I think it could make sense to let the entire batch operation fail, if any of them fail, so as to make handling the happy case nice and easy. So a batch error could either be list of `JsonRpcError`s (collecting all error responses), a `IncompleteError`, if any of the requests is missing a response or just a single `JsonRpcError`.

> Also I would not suggest to stick with the `Middleware` trait neither. Out of the solutions you proposed, the one I like the most and that I think provides the user with the better experience is the number **2**.
> 
> Edit: also number 3 looks good, but I would get rid of the limit of 16, which would not be of much usefulness in my opinion. If you think about id, it's very similar to what I proposed, and such intermediate objects would also greatly simplify the logic behind `BatchRequest`.

The more I think of it, I like option 3 the most. The intermediate `RpcCall` objects add some flexibility for the caller and they can implement `Future` directly and easily, so there is no additonal `call` method required. The arbitrary limit for the tuple length is because rust doesn't have variadic generics, so these trait implementations would have to be made by hand (macro), or do you mean something else? Additionally, impls for `Vec`s, arrays and slice can also be made for requests with the same output type. Ordering by ID would no longer be possible, though, because requests could be added to the batch in any order and the output order would have to match the input order.

gakonst:
(Recommend we move the design discussion to an issue about what the right way to implement this is :D)

Issue #749, opened 2021-12-29
Ability to set state override when calling a contract function


alex88:
**Is your feature request related to a problem? Please describe.**
It would be nice to be able to add the state override argument to eth_call. [docs here](https://geth.ethereum.org/docs/rpc/ns-eth#3-object---state-override-set)

**Describe the solution you'd like**
Be able to pass a state override struct to this function, after the bloock number

https://github.com/gakonst/ethers-rs/blob/e24117a1e1d6301d80101c7cc83c93ac4a8b1258/ethers-providers/src/provider.rs#L512-L520

gakonst:
Good idea.

While we're at it, maybe instead of `Result<Bytes, ProviderError>` we could return `ResponseBuilder<Result<Bytes, ProviderError>`.

Then, the user could do `provider.call(tx).block(BlockNumber::Latest).state(...).await?`. This would allow us to have a nicer interface for overriding function calls, without introducing a bunch of `Option`al arguments to it.

alex88:
I'm currently doing something similar (sorry for the late response but I had to get it to work quick and dirty), I've created:

```rust
use std::collections::HashMap;

use crate::types::{Address, Bytes, U256};
use serde::{Deserialize, Serialize, Serializer};

#[derive(Clone, Debug, Deserialize, Default)]
/// A Block Hash or Block Number
pub struct StateOverride {
    map: HashMap<Address, StateOverrideItem>,
}

impl StateOverride {
    pub fn add(mut self, address: Address, item: StateOverrideItem) -> Self {
        self.map.insert(address, item);
        self
    }
}

#[derive(Clone, Debug, Serialize, Deserialize, Default)]
pub struct StateOverrideItem {
    #[serde(skip_serializing_if = "Option::is_none")]
    pub balance: Option<U256>,

    #[serde(skip_serializing_if = "Option::is_none")]
    pub code: Option<Bytes>,
}

impl Serialize for StateOverride {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        self.map.serialize(serializer)
    }
}
```

to handle the state override parameter (I just need balance and code in my use case), then:

```rust
pub struct ContractCall<M, D> {
    /// The raw transaction object
    pub tx: TypedTransaction,
    /// The ABI of the function being called
    pub function: Function,
    /// Optional block number to be used when calculating the transaction's gas and nonce
    pub block: Option<BlockId>,
[...]

PR #1340, opened 2022-06-03
Add call builder and eth_call state overrides

gio256:
## Motivation

See #749.

Support for `eth_call` state [overrides](https://geth.ethereum.org/docs/rpc/ns-eth#3-object---state-override-set).

## Solution

Adds a `Provider::call_builder()` method, which returns a `Caller` struct. This `Caller` implements `Future`, so it can be `.await`ed the same way the async `call()` method is now. Additionally, it provides builder methods to set the block number and state overrides.

This might be useful as-is, but I think the ultimate goal would be to have what is currently named `call_builder()` replace the `Middleware::call()` method. It seems like this will require some signficant refactoring in addition to a breaking change, though.

Naming is all up for grabs, though "override" is a reserved keyword, which is the reason for the current choice of "spoof".

## Example

```rust
use ethers_core::{
    types::{Address, TransactionRequest, H256},
    utils::{parse_ether},
};
use ethers_providers::{Provider, Http, Middleware, spoof};
use std::convert::TryFrom;

let provider = Provider::<Http>::try_from("http://127.0.0.1:8545").unwrap();

let adr1: Address = "0x6fC21092DA55B392b045eD78F4732bff3C580e2c".parse().unwrap();
let adr2: Address = "0x295a70b2de5e3953354a6a8344e616ed314d7251".parse().unwrap();
let pay_amt = parse_ether(1u64)?;

// Not enough ether to pay for the transaction
let tx = TransactionRequest::pay(adr2, pay_amt).from(adr1).into();

// would fail (insufficient balance)
provider.call_builder(&tx).await.unwrap();

// override storage, nonce, etc. with the call
let mut state = spoof::state();
state
    .account(adr1)
    .store(H256::default(), 1.into())
    .nonce(2.into())
    .balance(pay_amt * 2);
provider.call_builder(&tx).block(100.into()).spoof(&state).await.unwrap();

// override just the balance
let state = spoof::balance(adr1, pay_amt * 2);
provider.call_builder(&tx).spoof(&state).await.unwrap();
```
 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


ethers-providers/src/provider.rs
```diff
@@ -1161,6 +1161,278 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+use serde::ser::SerializeTuple;
+use std::{
+    collections::HashMap,
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+impl<P: JsonRpcClient> Provider<P> {
+    pub fn call_builder<'a>(&'a self, tx: &'a TypedTransaction) -> Caller<'a, P> {
```
mattsse, line 1174: needs docs and perhaps an example

I think we should move this to the existing `impl Provider` section

ethers-providers/src/provider.rs
```diff
@@ -1161,6 +1161,278 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct CallInput<'a> {
+    tx: &'a TypedTransaction,
+    block: Option<BlockId>,
+    spoof: Option<&'a spoof::State>,
+}
+impl<'a> CallInput<'a> {
+    pub fn new(tx: &'a TypedTransaction) -> Self {
+        Self { tx, block: None, spoof: None }
+    }
```
mattsse, line 1188: let's move all of this to a separate module

ethers-providers/src/provider.rs
```diff
@@ -1161,6 +1161,278 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+        let pin = self.get_mut();
+        match pin {
+            Caller::Build(ref call) => {
+                let fut = Box::pin(call.execute());
+                *pin = Caller::Wait(fut);
+                ctx.waker().wake_by_ref();
+                Poll::Pending
+            }
+            Caller::Wait(ref mut fut) => fut.as_mut().poll(ctx),
+        }
```
mattsse, line 1273: this can be a loop instead of transitioning via a wake?

ethers-providers/src/provider.rs
```diff
@@ -1161,6 +1161,278 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+            }
+        }
+    }
+
+    #[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+    #[serde(transparent)]
+    pub struct State(#[serde(skip_serializing_if = "HashMap::is_empty")] HashMap<Address, Account>);
+
+    impl State {
+        pub fn new() -> Self {
```
mattsse, line 1348: `fn new` with default are redundant imo

ethers-providers/src/provider.rs
```diff
@@ -1161,6 +1161,278 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+                ctx.waker().wake_by_ref();
+                Poll::Pending
+            }
+            Caller::Wait(ref mut fut) => fut.as_mut().poll(ctx),
+        }
+    }
+}
+
+/// Provides types and methods for "spoofing" state overrides for eth_call
+pub mod spoof {
```
mattsse, line 1278: let's move this as well

ethers-providers/src/provider.rs
```diff
@@ -1161,6 +1161,278 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
...
+    input: CallInput<'a>,
+}
+
+impl<'a, P> CallBuilder<'a, P> {
+    fn new(provider: &'a Provider<P>, tx: &'a TypedTransaction) -> Self {
+        Self { provider, input: CallInput::new(tx) }
+    }
+}
+impl<'a, P: JsonRpcClient> CallBuilder<'a, P> {
+    fn execute(&self) -> impl Future<Output = Result<Bytes, ProviderError>> + 'a {
```
mattsse, line 1223: this is not async because this is the state of the caller?

mattsse (changes requested):
some nits :)

ethers-providers/src/provider.rs
gio256: The short answer is I tried async briefly and upset the borrow checker.

```rust
async fn execute(&self) -> Result<Bytes, ProviderError>;
```

As I understand it, the problem is that this causes the returned future to capture the anonymous lifetime `'self`. `Caller::Wait(PinBoxFut<'a, Bytes>)` needs to hold a future tied to the `'a` lifetime of the reference to the `Provider`. So making `execute()` async and attempting to store this future in `Future::poll` pulls in the `Pin<&mut Self>` lifetime which can't match the `'a` provider lifetime.

ethers-providers/src/provider.rs
gio256: Agreed, loop is much better here.

gio256:
@mattsse thanks for the feedback. Do you have any thoughts on keeping this as an auxiliary method vs. using it to replace `Middleware::call()`?

If the latter sounds like too big a lift or an untenable breaking change, I can clean things up with the intention of making this usable in its own right.

gakonst:
>  Do you have any thoughts on keeping this as an auxiliary method vs. using it to replace Middleware::call()?

i think we can merge without the breaking change, given this is a nice new feature, and we can then investigate the effort it'd take to do the breaking change in a follow on pr?

mattsse:
sgtm!

gakonst:
Main change I see needed is refactoring the spoofing stuff to a separate file. LGTM otherwise.

gio256:
Still todo:
- Docs
- Update the existing doc tests, which are broken.
- U256 and U64 values need to be serialized to their full length instead of truncated as the existing Serialize impl does.
- Move the values for storage overrides back to U256 (they are currently H256 to work around the above).
- Tests. Anvil doesn't support state overrides, and I don't know if there are currently any geth-based unit tests. But `examples/call_override.rs` can at least be extended to better test the implementation.

gio256:
I think this is approximately there, but happy to make any changes.

Some notes on the above:
- I added one significant test that relies on geth. Maybe this is best put behind a feature flag so you don't need geth to run all the tests? It looks like the example is also failing looking for a geth executable.
- Geth does actually accept U256 and U64 values that are not serialized to their full length. The only values it rejects for this are storage overrides.
- Thinking about it more, using H256 for storage override values seems to make more sense anyway, so I left this as-is. Assuming that you might need to pack multiple values into storage slot overrides to mimic solidity's behavior, the internal type of `H256([u8; 32])` seems more logical here than `U256([u64; 4])`.


You should now be able to use `.call_raw().block(n).state(state).await` anywhere that you would normally use `.call().await`, including contract interactions and deploy transactions. Let me know if I am missing any use cases here.

gakonst:
> I added one significant test that relies on geth. Maybe this is best put behind a feature flag so you don't need geth to run all the tests? It looks like the example is also failing looking for a geth executable.

You can use Anvil which we install on CI! See `ethers_core::utils::Anvil`

gakonst (approved):
great work on the API, very user friendly + excellent docs/examples

PR #1344, opened 2022-06-03
Make RetryClient cloneable

joshieDo:
## Motivation

Make `RetryClient` cloneable.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse (commented):
since the `RetryPolicy` is not object-safe, adding clone support is a bit more complicated,

I'd recommend we hold off on this for because I think, there'll be additional changes, with #1267

joshieDo:
Closing since I've gone full circle:

`RetryClient` shouldn't be cloneable since `requests_enqueued` cannot be shared

PR #1353, opened 2022-06-07
test: update test with changed state on forked tests

meetmangukiya:
Fix tests


meetmangukiya:
```
---- eth_tests::watch_events stdout ----
thread 'eth_tests::watch_events' panicked at 'assertion failed: `(left == right)`
  left: `"initial value"`,
 right: `"0"`', ethers-contract/tests/contract.rs:291:13
```
I think there is some bug in anvil filters that are returning logs from blocks before the filter was created, ws stream seems to be working fine. /cc @mattsse

mattsse (approved):

PR #1359, opened 2022-06-08
feat(solc): add script/ to project paths

joshieDo:
## Motivation

Currently, our convention on foundry is to have scripts in the `script/` folder. However, the folder does not actually belong to the project. It may lead to weird issues, if there's contracts declared inside when verifying contracts.

Alternatively, we could also add a more generic `additional_sources` where by default we add the `script/` folder.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/config.rs
+++ b/ethers-solc/src/config.rs
@@ -511,7 +511,7 @@ impl ProjectPathsConfigBuilder {
         self.tests = Some(utils::canonicalized(tests));
         self
     }
-    
+
     pub fn scripts(mut self, scripts: impl Into<PathBuf>) -> Self {
         self.scripts = Some(utils::canonicalized(scripts));
         self
--- a/ethers-solc/src/config.rs
+++ b/ethers-solc/src/config.rs
@@ -29,6 +29,8 @@ pub struct ProjectPathsConfig {
     pub sources: PathBuf,
     /// Where to find tests
     pub tests: PathBuf,
+    /// Where to find scripts
+    pub scripts: PathBuf,
     /// Where to look for libraries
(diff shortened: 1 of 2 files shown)


ethers-solc/src/config.rs
```diff
@@ -420,7 +437,8 @@ impl Default for ProjectPaths {
         Self {
             artifacts: "out".into(),
             sources: "src".into(),
-            tests: "tests".into(),
+            tests: "test".into(),
```
joshieDo, line 440: I changed here, because it seems like `tests` was a mistake here. Should I revert it?

ethers-solc/src/config.rs
```diff
@@ -538,7 +562,8 @@ impl ProjectPathsConfigBuilder {
                 .artifacts
                 .unwrap_or_else(|| ProjectPathsConfig::find_artifacts_dir(&root)),
             sources: self.sources.unwrap_or_else(|| ProjectPathsConfig::find_source_dir(&root)),
-            tests: self.tests.unwrap_or_else(|| root.join("tests")),
+            tests: self.tests.unwrap_or_else(|| root.join("test")),
```
joshieDo, line 565: I changed here, because it seems like `tests` was a mistake here. Should I revert it?

ethers-solc/src/config.rs
mattsse: that makes sense, is also the default of `foundry::Config`

mattsse (approved):
lgtm

mattsse (approved):

PR #1365, opened 2022-06-09
feat(solc): emit additional raw metadata field

### LINKED ISSUE #1325: Compilation artifacts are altered by JSON deserialization - resulting in a mismatch with the bytecode's appended IPFS hash
**Version**
master branch, building from source

**Platform**
Darwin joaquims-mbp.lan 21.3.0 Darwin Kernel Version 21.3.0: Wed Jan  5 21:37:58 PST 2022; root:xnu-8019.80.24~20/RELEASE_ARM64_T6000 arm64

**Description**

When running `forge build --extra-output metadata`, with `bytecode_hash: "ipfs"`, I encountered an unexpected mismatch when comparing the metadata IPFS hash from the bytecode and the IPFS hash of the metadata outputted by forge.

I first thought it might be because of absolute vs relative paths, filed https://github.com/foundry-rs/foundry/issues/1738 which got resolved, but even with that fix the IPFS don't match.

I kept investigating and found the root of the issue: the raw solc compilation output, particularly the [metadata](https://docs.soliditylang.org/en/latest/metadata.html) is being deserialized by `ethers-rs` and then serialized again when writing the final artifacts to disk.

This causes multiple alterations to the original metadata output from solc. Here's the ones I've confirmed:

## 1. The abi in `metadata.output.abi` is getting all empty inputs stripped, and `indexed` is missing from all event properties.

The empty inputs being stripped is caused by the json deserialization policy:

```rust
#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
pub struct SolcAbi {
  #[serde(default, skip_serializing_if = "Vec::is_empty")]
  pub inputs: Vec<Item>,
  ...
 ```
This can easily be verified by compiling a contract with an empty constructor with forge. The abi in the metadata will be missing the inputs property for the constructor, whereas the top level abi (the one that always get outputted by forge) will have it.

I believe that `SolcAbi` model is missing the `indexed` optional property as well, causing wrong event ABIs.

## 2. When the contract has no libraries, the `libraries: {}` entry in the metadata gets stripped.

Similarly, when deserializing the `settings` entry in the metadata, the JSON policy strips out any empty objects:

```rust
#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
pub struct Settings {
  ...
   #[serde(default, skip_serializing_if = "Libraries::is_empty")]
   pub libraries: Libraries,
}
```
Because of that, the metadata output is altered from its original content.

## 3. The metadata output contains the key `outputSelection` which is not in the solidity metadata spec

This is due to the re-use of `pub struct Settings` for representing both solc inputs AND outputs. But in the spec, `outputSelection` is only an input field, and should not be in the finale metadata output.

## 4. Deserializing and serializing the compiler output shuffles the keys in the final output JSON

The compiler outputs metadata with keys in **alphabetical** order, but because of the deserialization and serialization process, those keys get shuffled, resulting in a different JSON than the one that was generated by the compiler.

---

These are the immediate issues I found while testing with a simple contract, I believe there might be more.

My recommendation would be to avoid deserializing / serializing the compiler metadata output, and instead just output the raw string that is produced by the compiler without altering it. This ensures that the produced bytecode IPFS hash matches the exact metadata contents that gets written in the final artifact files.

I believe that if end users want to do something with the content of the metadata, it is trivial for them to parse the raw string into JSON and do what they need to do. It should not be the responsibility of `ethers-rs` or `forge` to format that into JSON, especially since it breaks the guarantee that this metadata matches the bytecode.

For what it's worth, both hardhat and truffle do exactly this, in their build output there is a `metadata` entry that is just a raw, escaped JSON string, taken straight from the solc output. Without any alteration or formatting, like this:

```
{
  ... rest of the artifact output
  "metadata": "{\"compiler\":{\"version\":\"0.8.13+commit.abaa5c0e\"},\"language\":\"Solidity\",...}"
 }
```

Let me know if that makes sense, I'd be happy to try and do these changes in a PR.

mattsse:
## Motivation
Closes #1325

## Solution
add an additional `rawMetadata:String` field to artifact output which preserves solc's metadata string

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts/serde_helpers.rs
+++ b/ethers-solc/src/artifacts/serde_helpers.rs
@@ -66,8 +66,8 @@ pub mod json_string_opt {
             if s.is_empty() {
                 return Ok(None)
             }
-
-            serde_json::from_str(&s).map_err(de::Error::custom).map(Some)
+            let value = serde_json::Value::String(s);
+            serde_json::from_value(value).map_err(de::Error::custom).map(Some)
         } else {
             Ok(None)
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,8 @@
 
 ### Unreleased
 
+- Add `rawMetadata:String` field to configurable contract output 
+  [#1365](https://github.com/gakonst/ethers-rs/pull/1365)
 - Use relative source paths and `solc --base-path`
   [#1317](https://github.com/gakonst/ethers-rs/pull/1317)
 - Save cache entry objects with relative paths
--- a/ethers-solc/src/artifact_output/configurable.rs
+++ b/ethers-solc/src/artifact_output/configurable.rs
@@ -9,7 +9,8 @@ use crate::{
             EwasmOutputSelection,
         },
         Ast, CompactContractBytecodeCow, DevDoc, Evm, Ewasm, FunctionDebugData, GasEstimates,
-        GeneratedSource, LosslessAbi, Metadata, Offsets, Settings, StorageLayout, UserDoc,
+        GeneratedSource, LosslessAbi, LosslessMetadata, Metadata, Offsets, Settings, StorageLayout,
+        UserDoc,
     },
     sources::VersionedSourceFile,
     ArtifactOutput, SolcConfig, SolcError, SourceFile,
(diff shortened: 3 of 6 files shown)


mattsse:
> Is the unit test failure a flake?

forgot that solc emits empty string for metadata regardless...

improved the json_opt deserialization

mattsse:
> The --all-features tests seem to be hanging for some reason..

this is super annoying because for some reason there are no logs available for the job...

perhaps this is due to some hanging request, but testing locally works...

mattsse:
@gakonst after looking at jobs that timed out, 

it looks like `can_stream_pending_transactions` test hangs

https://github.com/gakonst/ethers-rs/runs/6801222172?check_suite_focus=true

ignored it for now until I debugged this on anvil

ethers-solc/src/artifacts/serde_helpers.rs
```diff
@@ -66,8 +66,8 @@ pub mod json_string_opt {
             if s.is_empty() {
                 return Ok(None)
             }
-
-            serde_json::from_str(&s).map_err(de::Error::custom).map(Some)
+            let value = serde_json::Value::String(s);
```
gakonst, line 69: Why is this needed?

gakonst (approved):

PR #1377, opened 2022-06-14
fix(middleware): use signer chain when tx is None

Rjected:
## Motivation

The previous behavior never changes the tx chain id if it is `None`, since the tx is cloned inside the match statement.

## Solution

Move the tx clone outside the match statement.

Adds a test setting the tx chain id to `None`, with a `1337` chain id on the signer, ensuring that the transaction encoding and signature are changed properly.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-middleware/src/signer.rs
+++ b/ethers-middleware/src/signer.rs
@@ -126,7 +126,7 @@ where
     /// id.
     async fn sign_transaction(
         &self,
-        tx: TypedTransaction,
+        mut tx: TypedTransaction,
     ) -> Result<Bytes, SignerMiddlewareError<M, S>> {
         // compare chain_id and use signer's chain_id if the tranasaction's chain_id is None,
         // return an error if they are not consistent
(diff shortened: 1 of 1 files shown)


ethers-middleware/src/signer.rs
```diff
@@ -131,12 +131,12 @@ where
         // compare chain_id and use signer's chain_id if the tranasaction's chain_id is None,
         // return an error if they are not consistent
         let chain_id = self.signer.chain_id();
+        let mut tx = tx.clone();
```
gakonst, line 134: do we need to clone? let's just make `mut tx` instead in the function declaration?

ethers-middleware/src/signer.rs
```diff
@@ -382,6 +382,42 @@ mod tests {
         assert_eq!(tx, expected_rlp);
     }
 
+    #[tokio::test]
```
gakonst, line 385: great!

gakonst (changes requested):

ethers-middleware/src/signer.rs
Rjected: oh right, yeah we can do that, that makes more sense

gakonst (approved):

PR #1396, opened 2022-06-20
Use H64 for Block Nonce

jpopesculian:
## Motivation

`geth` and other libraries represent the block nonce as a 8 byte fixed array. `ethers-rs` uses an unsigned 64 bit integer to represent the nonce. This causes the nonce to be serialized without leading 0's as specified by the JSON-RPC spec, which therefore may cause deserialization problems with other libraries that expect a fixed byte array. This is the case with `web3` for example which fails to deserialize the nonce [because it uses `H64`](https://docs.rs/web3/0.18.0/web3/types/struct.Block.html#structfield.nonce)

## Solution

Use `H64` instead of `U64` for nonce

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/types/block.rs
+++ b/ethers-core/src/types/block.rs
@@ -1,5 +1,5 @@
 // Taken from <https://github.com/tomusdrw/rust-web3/blob/master/src/types/block.rs>
-use crate::types::{Address, Bloom, Bytes, Transaction, TxHash, H256, U256, U64};
+use crate::types::{Address, Bloom, Bytes, Transaction, TxHash, H256, H64, U256, U64};
 use chrono::{DateTime, TimeZone, Utc};
 #[cfg(not(feature = "celo"))]
 use core::cmp::Ordering;
@@ -81,7 +81,7 @@ pub struct Block<TX> {
     pub mix_hash: Option<H256>,
     /// Nonce
     #[cfg(not(feature = "celo"))]
-    pub nonce: Option<U64>,
+    pub nonce: Option<H64>,
     /// Base fee per unit of gas (if past London)
     #[serde(rename = "baseFeePerGas")]
     pub base_fee_per_gas: Option<U256>,
--- a/ethers-core/src/types/mod.rs
+++ b/ethers-core/src/types/mod.rs
@@ -5,7 +5,7 @@ pub type Selector = [u8; 4];
 /// A transaction Hash
 pub use ethabi::ethereum_types::H256 as TxHash;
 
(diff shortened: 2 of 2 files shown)


ethers-core/src/types/block.rs
```diff
@@ -81,7 +81,7 @@ pub struct Block<TX> {
     pub mix_hash: Option<H256>,
     /// Nonce
     #[cfg(not(feature = "celo"))]
-    pub nonce: Option<U64>,
+    pub nonce: Option<H64>,
```
mattsse, line 84: how does H64 serialize, deserialize compared to U64?

H64 is a hash type, and since nonce is an uint, this change doesn't seem necessary,

also H64 lacks common conversion trait impls for primitive types and I'm not sure it even as a `as_u64` function.

mattsse (changes requested):
> This causes the nonce to be serialized without leading 0's as specified by the JSON-RPC spec, which therefore may cause deserialization problems with other libraries that expect a fixed byte array. 

I don't think we should deviate from the spec just because other libraries do that.

> When encoding quantities (integers, numbers): encode as hex, prefix with "0x", the most compact representation (slight exception: zero should be represented as "0x0").

ethers-core/src/types/block.rs
jpopesculian: H64 will encode as a byte array, and not like number so it won't serialize to the most compact representation and it will expect the length to be 16 on deserialization. The reason this change is necessary is because a client of anvil (in my experience here graph-node) may expect the the value to have a length of 16 (or 18 with 0x prefix) and will fail to deserialize something like 0x0. Because the specs say this value is Bytes8 https://github.com/ethereum/execution-specs/blob/master/src/ethereum/muir_glacier/eth_types.py#L118 and geth also implements it as such, I think failing to deserialize 0x0 for the nonce is reasonable

gakonst:
Did we encounter this as a bug in someone's code, or is this a proactive change?

mattsse:
I just double checked the RPC docs

for blocks, nonce is indeed `DATA` whereas only for tx it is `QUANTITY`...

so @jpopesculian you're totally right, and we should change this to H64, 
geth also returns block nonces like that and so does hardhat
` "nonce": "0x0000000000000042",`

ref https://github.com/foundry-rs/foundry/issues/2088

mattsse:
@gakonst we can merge this, the failing tests are due to installed Anvil which uses the old nonce format

PR #1401, opened 2022-06-21
fix(docs): bump trezor client 0.0.6

joshieDo:
## Motivation
Fix current issue with docs.
`docs.rs failed to build ethers-0.13.0`

## Solution
bump  `hidapi-rs` to `v1.3.2` on the trezor client. Also, created https://github.com/summa-tx/bitcoins-rs/pull/99 , but I don't think it's necessary to wait.

ref https://github.com/joshieDo/hidapi-rs/pull/5

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse (approved):
ty

Issue #1400, opened 2022-06-21
Turn TypedTransaction functions into builder style functions


mattsse:
`TypedTransaction` functions like `TypedTransaction::set_value` return nothing.

they either should return `&mut Self` or consume self and return `Self`

if we don't want to break existing functions we should return `&mut Self`

h00p30:
Tell me if I'm wrong but from what I understand you want `TypedTransaction` to be the builder for different types of transaction like in [this](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html), so we can do things like `tx.set_from(addr1).set_to(addr2).set_value(1).build()` ?

mattsse:
landed with #1414

PR #1414, opened 2022-06-24
TypedTransaction methods into builder style

h00p30:
## Motivation
#1400

## Solution

Make methods like `set_from()` return  `&mut Self` in TypedTransaction

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog


mattsse (approved):
ty

PR #1423, opened 2022-06-28
feat: capture unknown fields

### LINKED ISSUE #1420: Using Celo and Ethereum at the same time
I've tried to improve an application that can connect to the Ethereum network, to be able to connect to the Celo network as well, at the same time. For this I had to add the `celo` feature to my dependencies. Now the Ethereum connection (which was working fine before) is failing with the following ethers::providers::ProviderError error:
```
JsonRpcClientError(SerdeJson { err: Error("missing field `randomness`", line: 1, column: 1563), text: "{\"difficulty\":\"0x2\",......
```

As I found out "randomness" is a Celo specific field in the Block, so it's normal that it will be missing from an Ethereum block.

I'd like to use Celo and Ethereum together, at the same time, in the same binary. I think the Celo specific fields should be optional, or there should be a different type for Celo blocks, by using something like Either or a generic solutions. These are just my rough ideas, I'm not well-versed enough in library development to determine the best solution for this, I hope the community can come up with something easy to use and maintain.

I tried to import the library twice, with different feature-sets, but unfortunately Rust doesn't support that. I tried to look it up and it seems to be a deliberate choice, because features should be "additive" in concept. I think this means that enabling a feature should not break an already working code. A feature could be enabled by any dependency downstream, even without the main program depending on it, so the "celo" feature being used should not break Ethereum support.

**Additional context**
If this issue takes a long time to fix it, or if it will be considered "unfixable" (which I hope is not the case), then at the minimum this would require a better error message, the sooner the better. Debugging the cryptic message `missing field` could be an issue for newcommers, if they accidentally mess up the initial configuration of the features.

mattsse:
## Motivation
Closes #1420

See docs

## Solution
*  Add `OtherFields` types that will capture unknown fields and provide util functions to deserialize values

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Unreleased
 
+- capture unknown fields in `Block` and `Transaction` type via new `OtherFields` type [#1423](https://github.com/gakonst/ethers-rs/pull/1423)
 - Methods like `set_to()` from `TypedTransaction` can be chained
 - Use H64 for Block Nonce [#1396](https://github.com/gakonst/ethers-rs/pull/1396)
 - Add `as_*_mut` methods on `TypedTransaction`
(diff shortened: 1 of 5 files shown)


gakonst (approved):
beautiful - let's follow that up in `cast`

PR #1424, opened 2022-06-28
fix: remove 0x bytecode object prefix for CompilerOutput

### LINKED ISSUE #1422: JSON Representation of `CompilerOutput` Incorrect
**Version**
`forge 0.2.0 (9dcc095 2022-06-28T00:03:46.153858156Z)`

**Platform**
`Linux 5.10.114-1-MANJARO x86_64 GNU/Linux`

**Description**
The `CompilerOutput` JSON that is included in the `BuildInfo` is causing Hardhat to crash because it is serializing the `bytecode` and `deployedBytecode` with a `0x` prefix while `solc` does not prefix those hex strings with `0x` prefixes. This difference causes the conversion from a string to buffer in hardhat to fail and subsequently crash.

If the build info files generated by foundry are present, it will crash hardhat due to this reason.

mattsse:
## Motivation
Closes #1422

## Solution
adds custom serialization for `Bytecode` objects that omits the `0x` prefix. This way the artifact types used by Forge (ConfigurableArtifact) are not affected and will still be output with 0x prefix

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/artifacts/bytecode.rs
+++ b/ethers-solc/src/artifacts/bytecode.rs
@@ -6,7 +6,7 @@ use crate::{
     utils,
 };
 use ethers_core::{abi::Address, types::Bytes};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Serializer};
 use std::collections::BTreeMap;
 
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


tynes:
Thank you!

gakonst (approved):

PR #1426, opened 2022-06-29
Propagate subscription deserialize errors

### LINKED ISSUE #1425: Deserialization of subscriptions fail silently
I don't get the reason behind this in principle, and in practice it makes implementing middleware with non-standard subscriptions annoying as you don't get a nice deser error, but instead you get *nothing*

Code: https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/pubsub.rs#L93

joleeee:
Would fix #1425

## Motivation

See #1425 

## Solution

Return a Result. I'm not sure if i did it the best way, feedback very welcome!

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-providers/src/pubsub.rs
+++ b/ethers-providers/src/pubsub.rs
@@ -78,20 +78,17 @@ where
     P: PubsubClient,
     R: DeserializeOwned,
 {
-    type Item = R;
+    type Item = Result<R, serde_json::Error>;
 
     fn poll_next(self: Pin<&mut Self>, ctx: &mut Context) -> Poll<Option<Self::Item>> {
         if !self.loaded_elements.is_empty() {
-            let next_element = self.get_mut().loaded_elements.pop_front();
+            let next_element = self.get_mut().loaded_elements.pop_front().map(|s| Ok(s));
             return Poll::Ready(next_element)
         }
 
         let this = self.project();
         match futures_util::ready!(this.rx.poll_next(ctx)) {
-            Some(item) => match serde_json::from_str(item.get()) {
-                Ok(res) => Poll::Ready(Some(res)),
-                _ => Poll::Pending,
-            },
+            Some(item) => Poll::Ready(Some(serde_json::from_str(item.get()))),
             None => Poll::Ready(None),
         }
     }


joleeee:
This will probably break compatability by changing return type (but ethers-rs does compile fine). an alternative is to panic, but I'm not sure about that...

ethers-providers/src/pubsub.rs
```diff
@@ -78,20 +78,17 @@ where
...
-            let next_element = self.get_mut().loaded_elements.pop_front();
+            let next_element = self.get_mut().loaded_elements.pop_front().map(|s| Ok(s));
             return Poll::Ready(next_element)
         }
 
         let this = self.project();
         match futures_util::ready!(this.rx.poll_next(ctx)) {
-            Some(item) => match serde_json::from_str(item.get()) {
-                Ok(res) => Poll::Ready(Some(res)),
-                _ => Poll::Pending,
```
mattsse, line 93: I think the reasoning behind this is, that deserializing events (in most cases) works, because this is supposed to be used with correct bindings anyway so returning Result here is not really necessary (in most cases)

mattsse (changes requested):

ethers-providers/src/pubsub.rs
```diff
@@ -78,20 +78,17 @@ where
     P: PubsubClient,
     R: DeserializeOwned,
 {
-    type Item = R;
+    type Item = Result<R, serde_json::Error>;
```
mattsse, line 81: this will break a lot

instead we should rather create a new `TrySubscriptionStream` type that returns results.

this would require some refactoring but so we can reuse lots of code but is the cleaner solution imo

ethers-providers/src/pubsub.rs
joleeee: I mean if it is the case that it silently stops after a single bad message, one could argue it'd be better to panic than to leave the program in a non functional state. And since this doesn't seem to be a normal issue, i would suspect it is exceedingly rare in normal operation... If it becomes a problem one could say to use (the theoretical) `TrySubscriptionStream` instead to avoid panics/catch errors.

mattsse: > it'd be better to panic

hard no on this

> it will never recover

wdym, it just continues with the next item.

> It constantly tries deserializing the same garbage and fails over and over

if your sub type does not match the object the node returns, yes, but imo this is more convenient in most cases, but I agree there should be an option to get the error

gakonst:
+1 on `TrySubscriptionStream` which returns Result, not supportive of making a breaking change on the default subscription API

joleeee:
Gonna close this for now

PR #1428, opened 2022-06-29
perf(providers): replace wake_by_ref with loop

mattsse:
## Motivation
instead of waking up the task when we do a state transition, we poll the next state directly.

## Solution
Replace `cx.waker().wake_by_ref();` with a loop. This way, the next state is polled immediately and we don't have to rely on the task being polled again right away.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
i thought that removing the loops was the right way to do it, and defer to the runtime to drive each loop step?

PR #1433, opened 2022-06-30
feat(ethers-solc): configurable build-info output dir

tynes:
## Motivation
I would like to be able to configure the output directory of the hh style bulidinfo files so that tooling that expects the build info files to be in a specific directory can just work.

## Solution

Allows for the build info output directory to be configured. It should be backwards compatible by using the same directory that it currently uses by default. Will require a follow up PR in foundry to pass through the config value

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Unreleased
 
+- Allow configuration of the output directory of the generated `BuildInfo` [#1433](https://github.com/gakonst/ethers-rs/pull/1433)
 - capture unknown fields in `Block` and `Transaction` type via new `OtherFields` type [#1423](https://github.com/gakonst/ethers-rs/pull/1423)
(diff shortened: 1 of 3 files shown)


tynes:
> this is great!
> 
> we can follow up with a corresponding config value in foundry.toml
> 
> the way that works is, the `Config` struct contains all settings variables: https://github.com/foundry-rs/foundry/blob/master/config/src/lib.rs#L90-L108
> 
> which are picked up by foundry.toml, env, cli args ...
> 
> and are then used to create the ethers-solc::Project here
> 
> https://github.com/foundry-rs/foundry/blob/master/config/src/lib.rs#L528

Thanks for the tips! Will open PRs in foundry and the hh plugin once this version of ethers is pulled into foundry

gakonst (approved):

Issue #1405, opened 2022-06-22
Use arbigen bindings without a provider


joleeee:
Is it possible to encode and decode contracts calls using bindings generated with abigen, but without actually instantiating them?

I know you can use the `MYCONTRACT_ABI.function("myFunction")`, and then decode/encode on that, but that's quite ugly, unergonomic and error-prone.

I'd be interested in submitting a PR myself but i wanna make sure there isn't some way that I've missed.

joleeee:
Thanks, that seems to get me further. Since they implement `EthCall`, i can easily encode and decode the input, but how would i go about parsing the output?

mattsse:
have a look at `EthAbiCodec` and `EthAbiType` derive macros

joleeee:
As far as i can tell, the structs only derive `EthCall` and `EthDisplay`, and the methods actually used when you instantiate don't use these structs.

Since there isn't any complementary `MyMethodReturn` to `MyMethodCall`, I'm not sure how you're suggesting i go about this. I see I can quite easily make my own struct that derives `EthAbiCodec` to get the `AbiEncode` and `AbiDecode` traits, but i still need the actual `MyMethodReturn` struct, which i was hoping Abigen would generate for me, like it generates `MyMethodCall`.

PR #1440, opened 2022-06-30
Generate structs for the return data of abigen-erated contracts

joleeee:
## Motivation
I explain the usecase in #1405, the short version is that you cannot easily decode the return of calls without an instatiated contract (`::new()` with a provider). Since abigen already generates structs representing calls, which AbiEncode into bytes, it only makes sense it should also have structs that AbiDecode the return data into a usable response representation.

## Solution
I hope the code is mostly self explanitory. Since I am new to contributing here, I'd love some feedback on the stuff i did so far.

I'm wondering if you'd want there to be an enum for return structs when there is more than one, like with the calls? I can put it back in but i initially removed it.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst:
like the direction - could you also add some tests for this?

joleeee:
I made a test now, let me know if it is satisfactory or what else should be tested. Note that this will not generate return bindings for functions that do not return anything.

ethers-contract/ethers-contract-abigen/src/contract/methods.rs
```diff
@@ -135,7 +141,47 @@ impl Context {
...
+    /// Expands to the corresponding struct type based on the inputs of the given function
+    fn expand_return_struct(
+        &self,
+        function: &Function,
+        alias: Option<&MethodAlias>,
+    ) -> Result<TokenStream> {
+        let struct_name = expand_return_struct_name(function, alias);
+        let fields = self.expand_output_params(function)?;
+        // no point in having structs when there is no data returned
+        if function.outputs.len() < 1 {
```
mattsse, line 153:
```suggestion
        if function.outputs.is_empty() {
```

mattsse (changes requested):
this only _generates_ structs, right?

this still does not replace the return type of the contract calls, right? those are still tuples?

joleeee:
That is correct, this only adds additional structs, it does not change the call-bindings in any way.

mattsse (approved):
lgtm

this should not break anything, since this merely generates new types

PR #1443, opened 2022-07-01
Implement fastrlp traits for Bytes

Rjected:
## Motivation

Many types in ethers and foundry contain the `Bytes` type, implementing fastrlp traits on any of them would be more convenient if `Bytes` implemented fastrlp traits.

## Solution

`Encodable` and `Decodable` are already implemented for `bytes::Bytes`, we can just defer to those implementations.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/Cargo.toml
+++ b/ethers-core/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/gakonst/ethers-rs"
 keywords = ["ethereum", "web3", "celo", "ethers"]
 
 [dependencies]
-fastrlp = { git = "https://github.com/vorot93/fastrlp", features = ["std", "derive", "ethereum-types"] }
+fastrlp = { version = "0.1.2", features = ["std", "derive", "ethereum-types"] }
 rlp = { version = "0.5.0", default-features = false, features = ["std"] }
 ethabi = { version = "17.1.0", default-features = false, features = ["full-serde", "rlp"] }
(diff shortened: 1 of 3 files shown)


ethers-core/Cargo.toml
```diff
@@ -10,6 +10,7 @@ repository = "https://github.com/gakonst/ethers-rs"
 keywords = ["ethereum", "web3", "celo", "ethers"]
 
 [dependencies]
+fastrlp = { git = "https://github.com/vorot93/fastrlp", features = ["std", "derive", "ethereum-types"] }
```
gakonst, line 13: Does this have a release?

gakonst (changes requested):

ethers-core/Cargo.toml
Rjected: No, not yet. `0.1.1` does not have the changes, a new release needs to be made

gakonst: @vorot93 could we cut a new release good sir?

vorot93: https://crates.io/crates/fastrlp/0.1.2

Rjected: 🚀🚀🚀 thanks! bumped to `0.1.2`

gakonst (approved):

PR #1444, opened 2022-07-01
Implement fastrlp traits for Signature

Rjected:
## Motivation

Deriving fastrlp `Encodable` and `Decodable` traits for types that contain `Signature` would not be possible without implementing `Encodable` and `Decodable` for `Signature`.

## Solution

 * Add `fastrlp` as a dependency to `ethers-core`, meaning this conflicts with #1443 for the time being
 * Implement `Hash` on `Signature` (just a convenience impl, not related to fastrlp traits)
 * Implement RLP encoding and decoding for `Signature`. This does not encode or decode a RLP header for a signature type, since types that would contain a signature (for example a [signed transaction](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#transaction-encoding-and-validity)) are usually encoded as `[...rest of object, v, r, s]` rather than `[...rest of object, [v, r, s]]`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/Cargo.toml
+++ b/ethers-core/Cargo.toml
@@ -10,6 +10,7 @@ repository = "https://github.com/gakonst/ethers-rs"
 keywords = ["ethereum", "web3", "celo", "ethers"]
 
 [dependencies]
+fastrlp = { version = "0.1.2", features = ["std", "derive", "ethereum-types"] }
 rlp = { version = "0.5.0", default-features = false, features = ["std"] }
 ethabi = { version = "17.1.0", default-features = false, features = ["full-serde", "rlp"] }
 arrayvec = { version = "0.7.2", default-features = false }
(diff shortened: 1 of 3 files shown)


ethers-core/Cargo.toml
```diff
@@ -10,6 +10,7 @@ repository = "https://github.com/gakonst/ethers-rs"
 keywords = ["ethereum", "web3", "celo", "ethers"]
 
 [dependencies]
+fastrlp = { git = "https://github.com/vorot93/fastrlp", features = ["std", "derive", "ethereum-types"] }
```
gakonst, line 13: Is there a released version we can use?

gakonst (changes requested):

ethers-core/Cargo.toml
Rjected: bumped to `0.1.2`

gakonst (approved):

PR #1446, opened 2022-07-03
chore(deps): solang 0.1.16

rkrasiuk:
## Motivation

upgrade solang for use in foundry formatter


mattsse (approved):

gakonst (approved):
Failure is a flake I think

PR #1458, opened 2022-07-06
ethers-solc: fix build info default directory

tynes:
## Motivation

This commit fixes the default behavior for the `build-info`
directory when it is unconfigured. It should end up in
the configured artifacts directory, taking into account
whatever the config for the artifacts directory is.
If the build info is specifically configured, then that
will take precendence over the default.

## Solution

Take into account the configured artifacts directory instead of using the default value

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


tynes:
Would super appreciate a merge and pull into foundry because this will fix our toolchain :)

PR #1477, opened 2022-07-13
fix unused warning

joleeee:
## Motivation
Authorization is current always `use`d when the target is not `wasm32`, but it's only actually used in the code when the `ws` feature is also enabled.  This causes an annoying error when i `cargo clippy`.

```rs
warning: unused import: `Authorization`
  --> <redacted>/ethers-rs/ethers-providers/src/provider.rs:11:25
   |
11 | use crate::transports::{Authorization, HttpRateLimitRetryPolicy, RetryClient};
   |                         ^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

```

## Solution
See the code, make it only import when needed

Have not used `#[cfg(...)]` a lot, might've done it a bit more verbose than necessary?

### DIFF
--- a/ethers-providers/src/provider.rs
+++ b/ethers-providers/src/provider.rs
@@ -8,7 +8,9 @@ use crate::{
 };
 
 #[cfg(not(target_arch = "wasm32"))]
-use crate::transports::{Authorization, HttpRateLimitRetryPolicy, RetryClient};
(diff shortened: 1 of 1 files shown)


mattsse (approved):
ty

PR #1490, opened 2022-07-20
nit(ethers-solc): set URL to binaries.soliditylang.org

paulrberg:


ethers-solc/src/compile/mod.rs
```diff
@@ -431,7 +431,7 @@ impl Solc {
     }
 
     /// Verify that the checksum for this version of solc is correct. We check against the SHA256
-    /// checksum from the build information published by binaries.soliditylang
+    /// checksum from the build information published by binaries.soliditylang.org
```
mattsse, line 434:
```suggestion
    /// checksum from the build information published by [binaries.soliditylang.org](https://binaries.soliditylang.org/)
```

paulrberg:
Thanks for the suggestion to use a hyperlink @mattsse.

roynalnaruto (approved):

PR #1500, opened 2022-07-22
set gas for eip 1559 txs on contract calls

chikko80:
Feature implementation

- Set max_fee_per_gas and max_priority_fee_per_gas individually on contract calls instead of setting both to the same value like the gas_price() method


ethers-contract/src/call.rs
```diff
@@ -125,6 +125,15 @@ impl<M, D: Detokenize> ContractCall<M, D> {
         self
     }
 
+    pub fn gas_price_for_eip_1559<T: Into<U256>>(
```
mattsse, line 128: please add docs

ethers-core/src/types/transaction/eip2718.rs
```diff
@@ -204,6 +204,25 @@ impl TypedTransaction {
         self
     }
 
+    pub fn set_gas_price_for_eip_1559<T: Into<U256>>(
```
mattsse, line 207: don't like the naming because this indicates it is only  eip1559 type tx

imo it would make more sense to add a designated setter for `max_priority_fee_per_gas`

mattsse (changes requested):

PR #1501, opened 2022-07-23
fix: use correct model for txpool_content endpoint

mattsse:
## Motivation
The previously used type didn't match the type returned by https://geth.ethereum.org/docs/rpc/ns-txpool#txpool_content a 100%

especially the tx hash

## Solution
use new type that represents it completely

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-providers/tests/txpool.rs
+++ b/ethers-providers/tests/txpool.rs
@@ -49,6 +49,6 @@ async fn txpool() {
     for i in 0..10 {
         tx = tx.nonce(i);
         let req = content.get(&i.to_string()).unwrap();
-        // assert_eq!(req, &tx);
+        assert_eq!(req, &tx);
     }
 }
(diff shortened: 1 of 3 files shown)


mattsse:
txpool test fails because changes not in anvil yet

ethers-providers/tests/txpool.rs
```diff
@@ -49,6 +49,6 @@ async fn txpool() {
     for i in 0..10 {
         tx = tx.nonce(i);
         let req = content.get(&i.to_string()).unwrap();
-        assert_eq!(req, &tx);
+        // assert_eq!(req, &tx);
```
gakonst, line 52:
```suggestion
        assert_eq!(req, &tx);
```

gakonst (approved):

PR #1516, opened 2022-07-26
feature: marker trait for contract events in abigen

prestwich:
## Motivation

Would like to easily write generics over events belonging to a specific contract

## Solution

Add a marker trait and impl it for all filter objects

## PR Checklist

This is a WIP PR. It has a simple approach, but hasn't been tested yet. Would love love love feedback as I haven't messed with proc macros much before

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/contract/events.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract/events.rs
@@ -26,7 +26,11 @@ impl Context {
             quote! {}
         };
 
+        let marker_trait = self.expand_events_marker_trait();
+
         Ok(quote! {
+            #marker_trait
+
             #( #data_types )*
 
             #events_enum_decl
@@ -49,6 +53,14 @@ impl Context {
 
             #events_method
         })
+    }
+
+    fn expand_events_marker_trait(&self) -> TokenStream {
+        let trait_name = self.expand_events_marker_trait_name();
+
+        quote! {
+            pub trait #trait_name {}
+        }
     }
 
     /// Generate an enum with a variant for each event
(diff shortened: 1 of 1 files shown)


mattsse (commented):
afaik, we don't have a `Contract` trait rather all generated are standalone structs. so worth thinking about what kind of abstraction we can add.

if a contract has more than one events we currently generate an enum that works like `serde(untagged)`, 
so perhaps something like this:

```rust
trait Contract {
  type Events: EthLogDecode;
  type Calls;
}
```

prestwich:
in this case, the thing I'd like to do is write a `struct Querier<T: ContractEvent>`. having a contract trait doesn't fill that need exactly

although I agree it would be useful to use `T::Events` for contracts in a lot of situations too

mattsse:
hmm hard to say. 
I guess the main benefit would be to simplify event queries. 

Querying is currently supported via `EthLogDecode`, unsure how `ContractEvent` would differ?

prestwich:
> hmm hard to say.
> I guess the main benefit would be to simplify event queries.

the reason i wanted this was to simplify event indexing pipelines and related things. it's common for a user to want to query ALL events from uniswap (e.g), and then index or otherwise handle each event. As it is, they need to write 1 type for each event that they care about. it'd be nice to be able to write a single type for common lifecycle events

gakonst:
makes sense @prestwich thx - supportive then and open to merging whatever design you think is right. if we want bigger changes like Matt suggested, recommend we investigate for 2.0 alongside the whole middleware redesign

PR #1523, opened 2022-07-28
ethers-solc: add immutableReferences output selector

tynes:
## Motivation

It is a property on the deployed bytecode object
on the compiler output. This is the precursor for
`forge inspect <contract-name> immutableReferences`.

See https://docs.soliditylang.org/en/v0.8.15/using-the-compiler.html?highlight=immutable#output-description

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Unreleased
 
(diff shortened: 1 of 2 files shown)


mattsse (approved):
ty

gakonst (approved):

PR #1530, opened 2022-07-28
Add a simple constructor for Wallet

moh-eulith:
## Motivation
I implemented `Signer`, but realized that I had to copy a large amount of code (Wallet and more) just to get a working implementation.

## Solution
A plain constructor for `Wallet` allows other implementation of `Signer` to use this code.

## PR Checklist

- [ ] Added Tests: correctness is by inspection of the one liner
- [x] Added Documentation: yes
- [ ] Updated the changelog: I don't understand the changelog. It's from a long time ago?

### DIFF
--- a/ethers-signers/src/wallet/mod.rs
+++ b/ethers-signers/src/wallet/mod.rs
@@ -70,7 +70,7 @@ pub struct Wallet<D: DigestSigner<Sha256Proxy, RecoverableSignature>> {
 impl<D: DigestSigner<Sha256Proxy, RecoverableSignature>> Wallet<D> {
 
     /// Construct a new wallet with an external Signer
-    pub fn new_plain(signer: D, address: Address, chain_id: u64) -> Self {
+    pub fn new_with_signer(signer: D, address: Address, chain_id: u64) -> Self {
         Wallet { signer, address, chain_id }
     }
 }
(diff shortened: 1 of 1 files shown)


gakonst:
Can you give an example of when you had to implement your own `DigestSigner`? This doesn't make full sense to me on why it'd be needed

moh-eulith:
There are devices (and custodians and what not) out there than can do ECDSA signatures: given a hashed message, spit out a 64 byte signature from a secretly held key.

Of course having a raw signature is far from being able to talk to the Eth chain. `Wallet` and associated code do all that heavy lifting. 

Honestly, the question seems a bit odd given that the code is already generified over the digest signer, for the very practical reason that there are 2 different implementations even in the library itself (private key and yubi key).

gakonst:
Hmm I was thinking that you could do this by implementing a `Signer` and then the `SignerMiddleware` does all the lifting, but I can see the need to expose a constructor anyway

ethers-signers/src/wallet/mod.rs
```diff
@@ -67,6 +67,14 @@ pub struct Wallet<D: DigestSigner<Sha256Proxy, RecoverableSignature>> {
     pub(crate) chain_id: u64,
 }
 
+impl<D: DigestSigner<Sha256Proxy, RecoverableSignature>> Wallet<D> {
+
+    /// Construct a new wallet with an external Signer
+    pub fn new_plain(signer: D, address: Address, chain_id: u64) -> Self {
```
gakonst, line 73:
```suggestion
    pub fn new_with_signer(signer: D, address: Address, chain_id: u64) -> Self {
```

gakonst (approved):

PR #1534, opened 2022-07-29
Add test reproducing panic when requesting unmined block from alchemy

ckoopmann:
## Motivation

Currently the provider panics when trying to get a non-existant block number from alchemy.
This is due to alchemy responding with a payload:

```
{
  "jsonrpc": "2.0",
  "id": 5
}
```

Which results in a panic.

## Solution

WIP: No solution yet

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-providers/src/provider.rs
+++ b/ethers-providers/src/provider.rs
@@ -1999,4 +1999,32 @@ mod tests {
         assert_eq!(tx.gas_price(), Some(gas_price));
         assert!(tx.access_list().is_none());
     }
+
+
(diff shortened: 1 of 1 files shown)


mattsse:
oh thanks for raising this, I think we need to add custom deserialize support for this, so we can return this as `Ok(None)` as intended

mattsse:
after looking at your tests, I don't understand the problem, because in both cases they return `None` as expected

ckoopmann:
> after looking at your tests, I don't understand the problem, because in both cases they return `None` as expected

Yes the issue seemed to have resolved on alchemy side. Today I get a response including `result: null` but, yesterday this was not the case. 

image

I'll close this for now and reopen if this issue reappears.

PR #1536, opened 2022-07-29
perf(providers): get response as bytes

mattsse:
## Motivation
uses `resp.bytes()` instead of `resp.text()` 

[text](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.text) decodes the body and scans the entire body. this is a bit redundant for json since deserialization is expected to fail if anyway

reqwest's [json](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.json) function is also just this:

```rust

pub async fn json<T: DeserializeOwned>(self) -> crate::Result<T> {
        let full = self.bytes().await?;

        serde_json::from_slice(&full).map_err(crate::error::decode)
    }
```

## Solution
get bytes and deserialize from_slice, and only convert to String in error case

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-providers/src/transports/http.rs
+++ b/ethers-providers/src/transports/http.rs
@@ -72,19 +72,24 @@ impl JsonRpcClient for Provider {
         let payload = Request::new(next_id, method, params);
 
         let res = self.client.post(self.url.as_ref()).json(&payload).send().await?;
-        let text = res.text().await?;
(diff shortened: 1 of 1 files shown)


gakonst (approved):
sgtm

PR #1540, opened 2022-07-31
fix(solc): convert source paths on windows

mattsse:
## Motivation
this fixes yet another path issue on windows which can result in duplicate (`src\\File.sol` and `src/File.sol`) artifacts

Ref https://github.com/foundry-rs/foundry/runs/7586758672?check_suite_focus=true

solc uses unix style paths `/`, if it tries to resolve an import `src/File.sol` it checks the VFS for a corresponding source, `src\\File.sol` does not necessarily match so it will resolve from disk which results in duplicated entries in the output. 

## Solution
convert source paths to `/` on windows before invoking solc

foundry tests: https://github.com/foundry-rs/foundry/pull/2531

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,8 +111,7 @@ jobs:
       - name: Install toolchain
         uses: actions-rs/toolchain@v1
         with:
-          # nightly regression https://github.com/rust-lang/rust/issues/95267
-          toolchain: nightly-2022-03-20
+          toolchain: nightly
           profile: minimal
           components: rustfmt, clippy
           override: true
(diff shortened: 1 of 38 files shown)


gakonst:
gotta love windows. i wonder if we've been doing things the wrong way all this time, or if it's just full of footguns? how can we avoid these in the future?

mattsse:
I guess the main problem is/was I've never worked with windows and just didn't know any better lol

since windows also supports `/` I think we can just convert all of them when ethers-solc spits out the Artifact set

trying to sort things out here https://github.com/foundry-rs/foundry/pull/2531 and make sure everything passes now, getting some weird failures that only happen in the cross compile tests...

gakonst (approved):

gakonst:
smol lint issue otherwise merging

mattsse:
lint issue is fixed here https://github.com/gakonst/ethers-rs/pull/1541

I'd like to keep this open until https://github.com/foundry-rs/foundry/pull/2531 passes

this is incredible tedious lol for some reason a random fuzzing test keeps failing randomly...

mattsse:
@gakonst this is ready now, it will spit out `/` paths on windows now

gakonst (approved):

PR #1549, opened 2022-08-02
feat: add EthError trait and derive

mattsse:
## Motivation
Add support for custom errors

## Solution
* introduce new `EthError` trait, basically same as `EthCall` with derive impl
  * unify some derive code for `EthCall` and `EthError` in `calllike` 
* support human readable error parsing
* generate bindings for custom solidity errors, if there are more than 1 error in the abi, we generate an error enum `enum <name>Errors {..` see Seaport test

 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,7 @@
 
 ### Unreleased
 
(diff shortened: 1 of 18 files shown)


gakonst (approved):
pretty reasonable

PR #1553, opened 2022-08-02
chore: seal extension traits

mattsse:
## Motivation
seal extension traits so they can't be implemented since these are only intended for `ethabi`.

this prevents accidental misuse where `EthCall` or `EthEvent` traits should actually be used

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst:
Smol lint

PR #1559, opened 2022-08-03
feat: add ProviderExt trait

mattsse:
## Motivation
moved from https://github.com/foundry-rs/foundry/pull/2559

## Solution
* add `ProviderExt` trait for `HttpProvider`

this is currently sealed to prevent impl outside of ethers until we finalize https://github.com/gakonst/ethers-rs/pull/1267

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/types/chain.rs
+++ b/ethers-core/src/types/chain.rs
@@ -4,6 +4,7 @@ use std::{
     convert::{TryFrom, TryInto},
     fmt,
     str::FromStr,
+    time::Duration,
 };
 use strum::EnumVariantNames;
 use thiserror::Error;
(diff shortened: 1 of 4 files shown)


gakonst (approved):

ethers-core/src/types/chain.rs
```diff
@@ -60,6 +61,39 @@ pub enum Chain {
...
+            Chain::BinanceSmartChain | Chain::BinanceSmartChainTestnet => 3_000,
+            Chain::Avalanche | Chain::AvalancheFuji => 2_000,
+            Chain::Fantom | Chain::FantomTestnet => 1_200,
+            Chain::Cronos | Chain::CronosTestnet => 5_700,
+            Chain::Evmos | Chain::EvmosTestnet => 1_900,
+            Chain::Aurora | Chain::AuroraTestnet => 1_100,
+            Chain::Oasis => 5_500,
+            Chain::Emerald => 6_000,
+            Chain::Dev | Chain::AnvilHardhat => 200,
+            _ => return None,
```
mfornet, line 90: I think it is better to explicitly catch other variants, otherwise it will be very easy for others contributor that are adding new networks to miss this match!

gakonst: What should we do? panic?

mfornet: sorry I wasn't clear. Still return None is fine, but instead of using `_` as a pattern, write down all other enum variants.

This way if a new network is added, compiler will help developer noticing that this is something to fill in.

mattsse: reasonable, do you want to add this?

mfornet: https://github.com/gakonst/ethers-rs/pull/1635

PR #1580, opened 2022-08-09
solc: fix bug in basic lib and basic contract generation

derch28:
## Motivation

In `add_basic_lib` and `add_basic_source` functions, the `format!` parameters are written in the wrong order. This creates contracts with "pragma solidity Contract.sol" and declared "contract ^0.8.0 {}".
## Solution

Swap the two parameters (after the string literal) passed to the `format!` macro in the mentioned functions.
## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

This is just a bug fix, so I don't think adding docs and updating changelog is necessary. If they are, please let me know.

### DIFF
--- a/ethers-solc/src/project_util/mod.rs
+++ b/ethers-solc/src/project_util/mod.rs
@@ -191,8 +191,8 @@ impl<T: ArtifactOutput> TempProject<T> {
 pragma solidity {};
 contract {} {{}}
             "#,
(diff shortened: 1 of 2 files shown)


mattsse (approved):
nice catch! ty

gakonst (approved):

Issue #1576, opened 2022-08-08
Update Multicall to Multicall3


DaniPopes:
### Problem
[Multicall](https://github.com/makerdao/multicall) supports very few chains and with different contract addresses, as well as always reverting on a single call failure. Another problem is that the `Multicall` struct is only initializable by either having the address beforehand or asynchronously with a call to `get_chain_id`.

### Proposed solution
Upgrading to the [maintained fork Multicall](https://github.com/mds1/multicall). It includes `Multicall3` with a higher chain coverage, always deployed at the same address (`0xcA11bde05977b3631167028862bE2a173976CA11`), retro-compatible and more advanced contract methods (ex. specify `allowFailure` or `value` for each call).

I have already made some changes for my own project and can easily port them over in a PR.

Please let me know what you think about this 😃

gakonst:
SGTM - PR is welcome

PR #1584, opened 2022-08-09
feat(contract): update Multicall to Multicall3

DaniPopes:
## Motivation

Closes: #1576 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-contract/src/multicall/mod.rs
+++ b/ethers-contract/src/multicall/mod.rs
@@ -234,6 +234,7 @@ impl TryFrom<u8> for MulticallVersion {
 /// [`add_call`]: #method.add_call
 /// [`call`]: #method.call
 /// [`send`]: #method.send
+#[must_use = "Multicall does nothing unless you use `call` or `send`"]
 pub struct Multicall<M> {
     version: MulticallVersion,
     legacy: bool,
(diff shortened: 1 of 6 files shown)


gakonst:
@mds1 @mattsse wdyt?

mds1:
Supportive, I agree with all the rationale in #1576 which was the motivation for Multicall3

DaniPopes:
Looks like we need to lock the entire `Multicall` module behind the `abigen` feature.. unless there's a way to generate the contract binding without the `derive`s?

DaniPopes:
^ CC @mattsse @gakonst - new contract binding requires the derive macros from `ethers-contract-derive` which needs the `abigen` feature to be enabled

gakonst:
Hmm should we just stop feature gating abigen as a feature altogether? @mattsse i always seem to need it

ethers-contract/src/multicall/mod.rs
```diff
@@ -330,28 +569,91 @@ impl<M: Middleware> Multicall<M> {
...
+                                        "Illegal revert.\n{:?}\n{:?}",
+                                        call, res
+                                    )),
+                                ))
+                            }
+
+                            // Decode with "Error(string)" (0x08c379a0)
+                            if ret.len() >= 4 && ret[..4] == [0x08, 0xc3, 0x79, 0xa0] {
+                                decode(&[ParamType::String], &ret[4..])?.remove(0)
+                            } else if ret.is_empty() {
```
mattsse, line 630: can use `String::decode(&ret[4..])` here
with `AbiDecode`

ethers-contract/src/multicall/multicall_contract.rs
```diff
@@ -1,97 +1,911 @@
...
-    });
-    pub struct MulticallContract<M>(Contract<M>);
-    impl<M> Clone for MulticallContract<M> {
+    // Some macros like EthAbiType and EthAbiCodec expand into using "ethers_contract" which is not
+    // defined here
+    mod ethers_contract {
+        pub use crate::*;
+        // re-export derive macros to build this multicall in case abigen feature is not enabled
+        pub use ethers_contract_derive::*;
```
mattsse, line 12: this reexport seems wrong

this reexports the derive macros not the traits

mattsse (changes requested):
sorry, this one slipped through.

some nits, but lgtm in general

mattsse (approved):
docs + test lgtm

gakonst (approved):
LGTM - it does add an extra param on add_call but should be ok

PR #1586, opened 2022-08-10
feat(TraceType+PartialEq): added PartialEq to TraceType

DrawnWren:
Forge also needs PartialEq on TraceType.

## Motivation

I was too dumb to see this the first time, so I had to make two PRs. Sorry.

## Solution
Add PartialEq to TraceType

### DIFF
--- a/ethers-core/src/types/trace/mod.rs
+++ b/ethers-core/src/types/trace/mod.rs
@@ -11,7 +11,7 @@ pub use filter::*;
 mod geth;
 pub use geth::*;
 
-#[derive(Debug, PartialEq, Clone, Deserialize, Serialize)]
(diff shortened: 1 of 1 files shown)


ethers-core/src/types/trace/mod.rs
```diff
@@ -11,7 +11,7 @@ pub use filter::*;
 mod geth;
 pub use geth::*;
 
-#[derive(Debug, Clone, Deserialize, Serialize)]
+#[derive(Debug, PartialEq, Clone, Deserialize, Serialize)]
```
mattsse, line 14: clippy will suggest +Eq
```suggestion
#[derive(Debug, PartialEq, Eq, Clone, Deserialize, Serialize)]
```

mattsse (changes requested):
ty, smol nit, since clippy will suggest this

PR #1590, opened 2022-08-11
feat(solc): resolve absolute imports in libraries

mattsse:
## Motivation
Closes https://github.com/foundry-rs/foundry/issues/2706
Resolve absolute imports like `src/interfaces/File.sol` in nested libs.

## Solution
try to find the existing file be scanning the `cwd` until we reach the `lib` folder, see docs

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/config.rs
+++ b/ethers-solc/src/config.rs
@@ -256,6 +256,14 @@ impl ProjectPathsConfig {
                         return Ok(import)
                     }
                 }
+                // also try to resolve absolute imports from the project paths
+                for path in [&self.root, &self.sources, &self.tests, &self.scripts] {
+                    if cwd.starts_with(path) {
+                        if let Ok(import) = utils::canonicalize(path.join(import)) {
+                            return Ok(import)
+                        }
+                    }
+                }
             }
 
             resolved.ok_or_else(|| {
--- a/ethers-solc/src/config.rs
+++ b/ethers-solc/src/config.rs
@@ -88,7 +88,7 @@ impl ProjectPathsConfig {
     ///
     /// See [IncludePaths]
     pub fn include_paths(&self) -> Vec<PathBuf> {
-        vec![self.sources.clone(), self.tests.clone(), self.scripts.clone()]
+        vec![self.root.clone(), self.sources.clone(), self.tests.clone(), self.scripts.clone()]
     }
 
     /// Creates all configured dirs and files
--- a/ethers-solc/src/config.rs
+++ b/ethers-solc/src/config.rs
@@ -219,6 +219,7 @@ impl ProjectPathsConfig {
             .components()
             .next()
             .ok_or_else(|| SolcError::msg(format!("Empty import path {}", import.display())))?;
+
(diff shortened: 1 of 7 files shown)


ethers-solc/src/config.rs
```diff
@@ -227,7 +228,24 @@ impl ProjectPathsConfig {
...
+            let resolved = self.resolve_library_import(import.as_ref());
+
+            if resolved.is_none() {
+                // absolute paths in solidity are a thing for example `import
+                // "src/interfaces/IConfig.sol"` which could either point to `cwd +
+                // src/interfaces/IConfig.sol`, or make use of a remapping (`src/=....`)
+                // While checking the former first would be safer, it is also less likely and more
+                // expensive, therefor we only check this if resolving the library import resulted
+                // in no path. However, this leaves open an unlikely edge case where a file
+                // `root + import` exists, which is neglected here.
```
gakonst, line 240: let's also add a unit test for this edge case?

ethers-solc/tests/project.rs
```diff
@@ -2162,3 +2162,52 @@ fn can_add_basic_contract_and_library() {
...
+            r#"
+    pragma solidity ^0.8.10;
+    import "myDepdendency/src/Config.sol";
+
+    contract Greeter {}
+   "#,
+        )
+        .unwrap();
+
+    let _compiled = project.compile().unwrap();
```
gakonst, line 2212: probably want a assert no compiler errors check here?

ethers-solc/src/config.rs
mattsse: I realized, this is actually not accurate when using `--include-path`

mattsse:
## Update

absolute import paths in nested libs require `--include-path`

This also adds `IncludePaths` support and enables it if `solc > 0.8.8` which introduced --include-path.

By default, includepaths will be src,test,script, this now makes forge `src` remappings obsolete.

when resolving libraries we also track additional dirs that should be added to the `IncludePaths` set

ethers-solc/tests/project.rs
mattsse: updated test

mattsse:
foundry tests also pass https://github.com/foundry-rs/foundry/runs/7809997586?check_suite_focus=true

with https://github.com/foundry-rs/foundry/pull/2747

ethers-solc/tests/project.rs
```diff
@@ -2211,3 +2211,110 @@ fn can_add_basic_contract_and_library() {
     assert!(compiled.find_first("Foo").is_some());
     assert!(compiled.find_first("Bar").is_some());
 }
+
+// <https://github.com/foundry-rs/foundry/issues/2706>
+#[test]
+fn can_handle_nested_absolute_imports() {
+    let mut project = TempProject::dapptools().unwrap();
+
+    let remapping = project.paths().libraries[0].join("myDepdendency");
```
gakonst, line 2220: typo but let's merge asap

gakonst (approved):

PR #1603, opened 2022-08-16
fix: sanitize absolute paths from etherscan

onbjerg:
## Motivation

Some paths from Etherscan may be absolute and just doing `output_directory.join(sanitized_path)` replaces the entire output path with the absolute path from Etherscan.

## Solution

Strip the leading `/` from absolute paths, forcing them to be relative.

See https://github.com/foundry-rs/foundry/issues/2792

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse (approved):
lgtm

PR #1604, opened 2022-08-16
fix(ethers-core): Deserialize a Filter request with `topics == null`

ngotchac:
## Motivation

I was getting some errors when using the `go-ethereum/ethclient` with Anvil/Foundry ; the issue was that when calling `eth_getLogs` with no `topics` set, `go-ethereum` would still pass the `topics` set to `null`: cf. https://github.com/ethereum/go-ethereum/blob/c4ab7d229117fdaa24292d84c1d450a727cb3e9f/ethclient/ethclient.go#L393

## Solution

Wrapping the topics in the Filter deserialize method in another `Option<..>` fixes the issue ; an added test ensures it does.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/types/filter.rs
+++ b/ethers-core/src/types/filter.rs
@@ -492,7 +492,7 @@ impl<'de> Deserialize<'de> for Filter {
                 let to_block = to_block.unwrap_or_default();
                 let block_hash = block_hash.unwrap_or_default();
                 let address = address.unwrap_or_default();
-                let topics_vec = topics.unwrap_or_default().unwrap_or_default();
+                let topics_vec = topics.flatten().unwrap_or_default();
 
                 // maximum allowed filter len
                 if topics_vec.len() > 4 {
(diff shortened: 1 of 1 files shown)


ethers-core/src/types/filter.rs
```diff
@@ -492,7 +492,7 @@ impl<'de> Deserialize<'de> for Filter {
                 let to_block = to_block.unwrap_or_default();
                 let block_hash = block_hash.unwrap_or_default();
                 let address = address.unwrap_or_default();
-                let topics_vec = topics.unwrap_or_default();
+                let topics_vec = topics.unwrap_or_default().unwrap_or_default();
```
mattsse, line 495:
```suggestion
                let topics_vec = topics.flatten().unwrap_or_default();
```

mattsse (approved):
nice catch!

PR #1605, opened 2022-08-16
fix(providers): validate address resolver

joshieDo:
## Motivation
ref https://github.com/foundry-rs/foundry/issues/2764

> seems like the domain ox63616e.eth has its resolver set to the WETH contract, ultimately leading to decoding a '0x' address

## Solution

Make sure that the resolver set is a valid one that supports `name(bytes32)` or any other selector that might be passed.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse (approved):
test lgtm!

gakonst (approved):

PR #1607, opened 2022-08-18
feat: add events function to set multiple event filter

mattsse:
## Motivation
Add `Filter::events(iter)` to set [A,B,..] for topic0 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (approved):

gakonst:
we should probably add an example of this

PR #1612, opened 2022-08-18
chore: make proof response fields pub

mattsse:
## Motivation
make EIP1186ProofResponse fields public 
change nonce type to U64, which matches geth

https://github.com/ethereum/go-ethereum/blob/23ac8df15302bbde098cab6d711abdd24843d66a/ethclient/gethclient/gethclient.go#L93-L94

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst:
dupe https://github.com/gakonst/ethers-rs/pull/1614

PR #1613, opened 2022-08-18
fix(providers): QuorumProvider zero-parameter json Value handling

mattiecnvr:
## Motivation

Currently the `QuorumProvider` fails for the very simple case of `provider.request("eth_blockNumber", ())`.

Example code which fails:
```rs
use ethers::prelude::{Http, JsonRpcClient, Quorum, QuorumProvider, WeightedProvider, U64};

#[tokio::main(flavor = "current_thread")]
async fn main() {
    let provider1 = "http://127.0.0.1:8545".parse::<Http>().unwrap();
    let provider2 = "http://127.0.0.1:8545".parse::<Http>().unwrap();
    let provider3 = "http://127.0.0.1:8545".parse::<Http>().unwrap();
    let quorum_provider = QuorumProvider::builder()
        .add_providers(
            [&provider1, &provider2, &provider3]
                .into_iter()
                .cloned()
                .map(WeightedProvider::new),
        )
        .quorum(Quorum::Majority)
        .build();

    let block_number: U64 = quorum_provider
        .request("eth_blockNumber", ())
        .await
        .expect("Failed to get block number");
    println!("Quorum block number {block_number}");
}
```

Ran against the hardhat node.
```
$ yarn hardhat node
```

Bit about my env:
```
$ yarn hardhat --version
2.9.9
```

The error message observed from the above code is:
```
thread 'main' panicked at 'Failed to get block number: JsonRpcClientError(NoQuorumReached { values: [], errors: [JsonRpcClientError(SerdeJson { err: Error("invalid type: null, expected u64", line: 1, column: 26), text: "{\"jsonrpc\":\"2.0\",\"id\":null,\"error\":{\"code\":-32600,\"message\":\"Invalid request\",\"data\":{\"message\":\"Invalid request\"}}}" }), JsonRpcClientError(SerdeJson { err: Error("invalid type: null, expected u64", line: 1, column: 26), text: "{\"jsonrpc\":\"2.0\",\"id\":null,\"error\":{\"code\":-32600,\"message\":\"Invalid request\",\"data\":{\"message\":\"Invalid request\"}}}" }), JsonRpcClientError(SerdeJson { err: Error("invalid type: null, expected u64", line: 1, column: 26), text: "{\"jsonrpc\":\"2.0\",\"id\":null,\"error\":{\"code\":-32600,\"message\":\"Invalid request\",\"data\":{\"message\":\"Invalid request\"}}}" })] })'
```

## Solution

After much digging into why this was an issue with `QuorumProvider` and not also with `Http`, I noticed that the http provider is using the following to serialize the body.
```rust
fn is_zst<T>(_t: &T) -> bool {
    std::mem::size_of::<T>() == 0
}

#[derive(Serialize, Deserialize, Debug)]
/// A JSON-RPC request
pub struct Request<'a, T> {
    id: u64,
    jsonrpc: &'a str,
    method: &'a str,
    #[serde(skip_serializing_if = "is_zst")]
    params: T,
}
```

What happens instead with `QuorumProvider` is it is using `serde_json::Value` types to support multiple different backends through the `JsonRpcClientWrapper` which defines something like:
```rust
impl<C: JsonRpcClient> JsonRpcClientWrapper for C {
    async fn request(&self, method: &str, params: Value) -> Result<Value, ProviderError> {
        Ok(JsonRpcClient::request(self, method, params).await.map_err(C::Error::into)?)
    }
}
```

the consequence is that when `params` is `()` it will be serialized as `serde_json::Value::Null` instead of being excluded from the body. Some providers may accept this as valid, but since it is not working with the hardhat node we need it changed at a minimum for our local code tests. The two solutions I saw for this are to:

1) Make it an Option<Value> which can be excluded from the body
2) Use an empty array

~~This PR opts for option (2) as it seems to be the minimal change.~~
See discussion below.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-providers/src/transports/quorum.rs
+++ b/ethers-providers/src/transports/quorum.rs
@@ -166,7 +166,7 @@ impl<T: JsonRpcClientWrapper> QuorumProvider<T> {
     /// This is the minimum of all provider's block numbers
     async fn get_minimum_block_number(&self) -> Result<U64, ProviderError> {
         let mut numbers = join_all(self.providers.iter().map(|provider| async move {
-            let block = provider.inner.request("eth_blockNumber", serde_json::json!(())).await?;
+            let block = provider.inner.request("eth_blockNumber", serde_json::json!([])).await?;
             serde_json::from_value::<U64>(block).map_err(ProviderError::from)
         }))
         .await
@@ -437,7 +437,12 @@ where
         method: &str,
         params: T,
     ) -> Result<R, Self::Error> {
-        let mut params = serde_json::to_value(params)?;
+        let mut params = if std::mem::size_of::<T>() == 0 {
+            // we don't want `()` to become `"null"`.
+            serde_json::json!([])
+        } else {
+            serde_json::to_value(params)?
+        };
         self.normalize_request(method, &mut params).await;
 
         let requests = self
--- a/ethers-providers/src/transports/quorum.rs
+++ b/ethers-providers/src/transports/quorum.rs
@@ -1,5 +1,4 @@
 use std::{
-    fmt,
     fmt::Debug,
     future::Future,
     pin::Pin,
@@ -166,7 +165,7 @@ impl<T: JsonRpcClientWrapper> QuorumProvider<T> {
     /// This is the minimum of all provider's block numbers
     async fn get_minimum_block_number(&self) -> Result<U64, ProviderError> {
         let mut numbers = join_all(self.providers.iter().map(|provider| async move {
-            let block = provider.inner.request("eth_blockNumber", serde_json::json!([])).await?;
+            let block = provider.inner.request("eth_blockNumber", QuorumParams::Zst).await?;
             serde_json::from_value::<U64>(block).map_err(ProviderError::from)
         }))
         .await
@@ -181,7 +180,13 @@ impl<T: JsonRpcClientWrapper> QuorumProvider<T> {
     }
 
     /// Normalizes the request payload depending on the call
-    async fn normalize_request(&self, method: &str, params: &mut Value) {
+    async fn normalize_request(&self, method: &str, q_params: &mut QuorumParams) {
+        let params = if let QuorumParams::Value(v) = q_params {
+            v
+        } else {
+            // at this time no normalization is required for calls with zero parameters.
+            return
+        };
         match method {
             "eth_call" |
             "eth_createAccessList" |
@@ -364,8 +369,8 @@ impl From<QuorumError> for ProviderError {
 
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
-pub trait JsonRpcClientWrapper: Send + Sync + fmt::Debug {
-    async fn request(&self, method: &str, params: Value) -> Result<Value, ProviderError>;
+pub trait JsonRpcClientWrapper: Send + Sync + Debug {
+    async fn request(&self, method: &str, params: QuorumParams) -> Result<Value, ProviderError>;
 }
 type NotificationStream =
     Box<dyn futures_core::Stream<Item = Box<RawValue>> + Send + Unpin + 'static>;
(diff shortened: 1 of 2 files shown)


ethers-providers/src/transports/quorum.rs
```diff
@@ -437,7 +437,12 @@ where
         method: &str,
         params: T,
     ) -> Result<R, Self::Error> {
-        let mut params = serde_json::to_value(params)?;
+        let mut params = if std::mem::size_of::<T>() == 0 {
+            // we don't want `()` to become `"null"`.
+            serde_json::json!([])
+        } else {
+            serde_json::to_value(params)?
+        };
```
mattsse, line 445: I don't think this is the right, fix.

as you've pointed out the params field is skipped by `Request` if it's a zst, which is not the case for `serde_json::json!([])`.

we had the same issue with the RetryProvider that we fixed like this:

https://github.com/gakonst/ethers-rs/blob/92b51b6d9fbec759857a5af01437540bee5edde2/ethers-providers/src/transports/retry.rs#L143-L157

I suggest we extract this solution and make it applicable for the quorum provider as well.

mattsse (changes requested):
this is a great find, and we should put a better emphasis on the ZST condition.

I don't think that this is the right fix though

ethers-providers/src/transports/quorum.rs
mattiecnvr: Made the changes, hopefully this is what you were thinking. Was debating between pulling it out fully and just kinda duplicating as I did and this felt okay given that in most places we can hide the details. If you want it fully abstracted please let me know where you think it should go.

ethers-providers/src/transports/quorum.rs
```diff
@@ -556,6 +572,15 @@ where
     }
 }
 
+/// Helper type that can be used to pass through the `params` value.
+/// This is necessary because the wrapper provider is supposed to skip the `params` if it's of
+/// size 0, see `crate::transports::common::Request`
+#[derive(Clone)]
+pub enum QuorumParams {
```
mattsse, line 579: don't need to expose probably
```suggestion
enum QuorumParams {
```

ethers-providers/src/transports/quorum.rs
```diff
@@ -181,7 +180,13 @@ impl<T: JsonRpcClientWrapper> QuorumProvider<T> {
...
 
     /// Normalizes the request payload depending on the call
-    async fn normalize_request(&self, method: &str, params: &mut Value) {
+    async fn normalize_request(&self, method: &str, q_params: &mut QuorumParams) {
+        let params = if let QuorumParams::Value(v) = q_params {
+            v
+        } else {
+            // at this time no normalization is required for calls with zero parameters.
+            return
+        };
```
mattsse, line 189: this is correct, since no params

ethers-providers/src/transports/quorum.rs
```diff
@@ -381,22 +386,28 @@ pub trait PubsubClientWrapper: JsonRpcClientWrapper {
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 impl<C: JsonRpcClient> JsonRpcClientWrapper for C {
-    async fn request(&self, method: &str, params: Value) -> Result<Value, ProviderError> {
-        Ok(JsonRpcClient::request(self, method, params).await.map_err(C::Error::into)?)
+    async fn request(&self, method: &str, params: QuorumParams) -> Result<Value, ProviderError> {
+        let fut = if let QuorumParams::Value(params) = params {
+            JsonRpcClient::request(self, method, params)
+        } else {
+            JsonRpcClient::request(self, method, ())
```
mattsse, line 393: that's the critical part here

the `()` ensures  we end up with a ZST type

mattsse (approved):
thanks!

duplicating is fine here imo.

lgtm

ethers-providers/src/transports/quorum.rs
mattiecnvr: Sadly we do because the wrapper is also `pub`. Oh, BTW, I have a bone to pick about this because if you `use ethers::prelude::*` it currently means the `JsonRpcClientWrapper` creates a conflict on the `request` function forcing you to call it via `JsonRpcClient::request(...)` or just not `use ethers::preulde::*`,

mattsse: I see,

that's actually a good point.

imo quorum should not be reexported via `prelude::*`

PR #1615, opened 2022-08-18
Export LogQueryError

recmo:
## Motivation

It is hard to create a wrapper error type containing errors returned from a `LogQuery`. The `LogQuery` type is public and accessible through `ethers::providers::LogQuery`, but the `LogQueryError` type is not. It can in principle be accesses through the `Stream` trait as such:

```
type LogQueryError<'a> = <LogQuery<'a> as Stream>::Item::Error;
```

but that introduces an unnecessary lifetime parameter and feels pretty hacky.

## Solution

```
pub use log_query::{LogQuery, LogQueryError};
``` 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -20,7 +20,7 @@ mod pending_escalator;
 pub use pending_escalator::EscalatingPending;
 
 mod log_query;
-pub use log_query::LogQuery;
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I can't see any reason why we shouldn't export this

PR #1619, opened 2022-08-19
Don't auto-generate an AccessList when sending/filling a tx

ngotchac:
## Motivation

Cf. discussion in https://github.com/foundry-rs/foundry/pull/2839 

I think that the automatic generation of access-list isn't a wanted feature, for a couple of reasons:
1. There is no way to disable it when using a compatible tx type (if it's set to empty it's always queried)
2. There is no good solution to compare the gas usages. As per other implementations (Geth, Nevermind), the returned GasUsed is the actual transaction gas usage, and not the gas estimation (which wouldn't include saved gas from using AL).

Thus I think it should be a manual step.

## Solution

Disable the automatic fetching of AL when filling a transaction.

I also added an optional block parameter to the gas estimation methods, since it's available on every node implementation. It also rightly use it when a block is passed to `send_transaction` or `fill_transaction`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-contract/src/call.rs
+++ b/ethers-contract/src/call.rs
@@ -150,7 +150,7 @@ where
 
     /// Returns the estimated gas cost for the underlying transaction to be executed
     pub async fn estimate_gas(&self) -> Result<U256, ContractError<M>> {
-        self.client.estimate_gas(&self.tx).await.map_err(ContractError::MiddlewareError)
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm,

as discussed I think that's the right fix here.

should put that in the changelog though

gakonst (approved):
yeah this ended up causing more pain than benefit

PR #1628, opened 2022-08-22
fix(signers): switch to the official AWS SDK for Rust

tomtau:
## Motivation

rusoto is only passively maintained

## Solution

switch to the official SDK: https://docs.aws.amazon.com/sdk-for-rust/latest/dg/welcome.html

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -241,6 +241,7 @@
 
 ### Unreleased
 
+- `AwsSigner` switched to the official AWS SDK for Rust [#1628](https://github.com/gakonst/ethers-rs/pull/1628)
 - `eth-keystore-rs` crate updated. Allow an optional name for the to-be-generated
   keystore file [#910](https://github.com/gakonst/ethers-rs/pull/910)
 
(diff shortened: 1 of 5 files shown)


mattsse (approved):
this makes sense.

lgtm.

cc @prestwich

prestwich:
first line of the docs is that it might be subject to breaking changes. Is this change driven by a specific need? Are we sure the dev preview is the correct place to start depending on it?

tomtau:
@prestwich one specific need on my side is that I'd like to to run the signer over vsock instead of TCP which is possible via: https://docs.rs/aws-sdk-kms/0.17.0/aws_sdk_kms/struct.Client.html#method.from_conf_conn
It's still a developer preview, but it's IMHO on a similar level as Rusoto or better. I've been using it since early alphas and breaking changes took a few minutes of refactoring work from my time so far.

tomtau:
(I'm not blocked by this change; I thought I may just upstream it, but ok to put it off until the SDK has a stable release)

gakonst:
Thanks @tomtau! I think given breakages let's hold off here, but supportive of this directionally eventually.

PR #1632, opened 2022-08-23
Fix 'get_transactions' response.

cobward:
## Motivation

The response to the `get_transactions` method on the etherscan client has a number of issues:
1. `methodId` and `functionName` fields are missing.
2. A number of fields are of the type `GenesisOption`, but from my investigations only `hash` and `from` can have this format.
3. `input` does not need to be wrapped in an option. From my investigation, the response from etherscan is always `"0x"` for empty input, and the serde implementation for the `Bytes` type will process this as an empty byte array.

## Solution

1. Add the missing fields.
2. Replace `GenesisOption` with `Option` on fields that cannot start with `GENESIS`.
3. Remove the option wrapper for the `bytes` field.

## PR Checklist

- [x] Added Tests (no tests needed as the existing tests are sufficient)
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-etherscan/src/account.rs
+++ b/ethers-etherscan/src/account.rs
@@ -1,7 +1,7 @@
 use crate::{Client, EtherscanError, Query, Response, Result};
 use ethers_core::{
     abi::Address,
-    types::{serde_helpers::*, BlockNumber, Bytes, H256, U256},
+    types::{serde_helpers::*, BlockNumber, Bytes, H256, H32, U256},
 };
 use serde::{Deserialize, Serialize};
 use std::{
@@ -17,7 +17,7 @@ pub struct AccountBalance {
     pub balance: String,
 }
 
-mod jsonstring {
+mod genesis_string {
     use super::*;
     use serde::{
         de::{DeserializeOwned, Error as _},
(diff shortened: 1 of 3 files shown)


ethers-etherscan/src/account.rs
```diff
@@ -63,7 +63,82 @@ mod jsonstring {
...
+    where
+        T: DeserializeOwned,
+        D: Deserializer<'de>,
+    {
+        let json = Cow::<'de, str>::deserialize(deserializer)?;
+        if json.is_empty() {
+            Ok(Option::None)
+        } else {
+            let value =
+                serde_json::from_str(&format!("\"{}\"", &json)).map_err(D::Error::custom)?;
```
mattsse, line 96: I believe this could be a one-liner by adding `.map(Some)`

ethers-etherscan/src/account.rs
```diff
@@ -63,7 +63,82 @@ mod jsonstring {
...
+    where
+        T: DeserializeOwned,
+        D: Deserializer<'de>,
+    {
+        let json = Cow::<'de, str>::deserialize(deserializer)?;
+        if json.is_empty() || json == "0x" {
+            Ok(Option::None)
+        } else {
+            let value =
+                serde_json::from_str(&format!("\"{}\"", &json)).map_err(D::Error::custom)?;
```
mattsse, line 132: same here

mattsse (approved):
the etherscan api continues to be a bit tricky -.-

thanks for this, lgtm, smol suggestion

ethers-etherscan/src/account.rs
cobward: Done

ethers-etherscan/src/account.rs
cobward: and done!

ethers-etherscan/src/account.rs
```diff
@@ -135,7 +135,10 @@ mod hex_string {
     }
 }
 
-/// Possible values for some field responses
+/// Possible values for some field responses.
+///
+/// Transactions from the Genesis block may contain fields that do not conform to the expected
+/// types.
```
kyruzic, line 141: Thank you for adding this explanation :)

PR #1635, opened 2022-08-24
nit: Explicitly handle all chains on match

mfornet:
## Motivation

Whenever `Chain` enum is matched, all its variant should be handled explicitly. This will help developers get assistance from the compiler to figure out all places that they need to take care if a new chain is added.

## Solution

Avoid wildcard pattern (`_`) while matching chain.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mfornet:
Is it worth adding this to the Changelog? If every PR goes to Changelog, then probably yes, otherwise, this is small enough to not be added IMO.

mattsse (approved):
ty

gakonst (approved):

PR #1638, opened 2022-08-25
fix(providers): Quorum provider resilience to out-of-sync and failing nodes

mattiecnvr:
Wanted some thoughts on this before I go through and fix up tests and the like. Currently tests are not passing; as this PR does change the behavior, I do not think the failing tests indicate an issue at this time, but I have not really dug into it.

## Motivation

This PR addresses a couple of main issues:
1) Currently `get_minimum_block_number`, which is used for a number of request pathways to replace `latest` block parameters, is not resistent to
1a) any provider failing, even if a quorum of providers successfully return their block numbers
1b) a provider falling out-of-sync and returning a (very?) low block number (technically a possible attack vector as well)
2) eth_getBlockByNumber still passes through `"latest"` instead of replacing with a specific block number

The spirit of the changes made is to bring the resilience of a quorum to (1) and patch (2) since it was probably just forgotten.

## Solution

A few different options were considered to address (1), namely to either wait for any quorum's worth of responses and then use the minimum of that response (which does not protect against (1b)) or to wait for all responses, separate out errors, and take the minimum block number of the maximum quorum. The latter was chosen.

The algorithm for this is to simply
1) request block numbers from all providers
2) wait for all responses (using a timeout on providers is a must for this to be resilient to providers dropping requests)
3) split responses into Ok/Err lists (returns the errors if a quorum is not reached)
4) sort Ok responses by descending block order
5) find the minimum number of providers required from the head of the list to reach a quorum
6) take the minimum block number of this list of providers

By doing this, we should be resilient to both (1a) and (1b).

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-providers/src/transports/quorum.rs
+++ b/ethers-providers/src/transports/quorum.rs
@@ -9,7 +9,7 @@ use crate::{provider::ProviderError, JsonRpcClient, PubsubClient};
 use async_trait::async_trait;
 use ethers_core::types::{U256, U64};
 use futures_core::Stream;
-use futures_util::{stream, FutureExt, StreamExt};
+use futures_util::{future, FutureExt, StreamExt};
 use serde::{de::DeserializeOwned, Serialize};
 use serde_json::{value::RawValue, Value};
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


mattsse:
sorry this one slipped through.

this makes a lot of sense, perhaps it would be useful to support customizing this behavior via a trait that we can `Box<dyn>` and use the described logic as default

gakonst:
@mattiecnvr is this rfr? can take a pass if so, else let us know on how to handle the next steps

mattiecnvr:
@gakonst 
This was made a while ago... We have been tweaking a bunch of things tbh so if you have some time might make sense for you to poke through a full diff and see if there is anything else that should get pulled in. I would have to spend some time I think to get this fully ready but the point of the draft was mostly to see if you like the idea.

This is our current version.
https://github.com/hyperlane-xyz/ethers-rs/blob/abacus/ethers-providers/src/transports/quorum.rs

gakonst:
Got it - thansk! Closing then and we can take a pass in the future.

PR #1647, opened 2022-08-30
fix: set chain id explicitly

mattsse:
## Motivation
anvil now enforces chain id https://github.com/foundry-rs/foundry/pull/2995, whereas ganache does not.
this broke the tests

hardhat also enforces correct chain id, https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md 

implemented [_validateTxV](https://github.com/ethereumjs/ethereumjs-monorepo/blob/2d60261422437c2353e5b547dbde610494d37c5e/packages/tx/src/legacyTransaction.ts#L361)

added test to confirm that

## Solution
* set chain id where required
* expose `Anvil::chain_id`

it'd be nice if the `Anvil` instance could just return a Wallet directly, but this is currently not possible since `Wallet` is not part of the core crate...
perhaps moving  + feature gating would be possible, but a bit of a hassle...

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-contract/tests/it/abigen.rs
+++ b/ethers-contract/tests/it/abigen.rs
@@ -4,7 +4,7 @@
 use ethers_contract::{abigen, Abigen, EthCall, EthEvent};
 use ethers_core::{
     abi::{AbiDecode, AbiEncode, Address, Tokenizable},
-    types::{transaction::eip2718::TypedTransaction, Eip1559TransactionRequest, U256},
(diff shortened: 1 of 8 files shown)


mattsse:
remaining errors seem to be limited to etherscan

gakonst (approved):

PR #1653, opened 2022-08-31
fix(abigen): only derive default of no arrays len > 32

### LINKED ISSUE #1640: Support `FixedArray` with larger than 32 elements in `abigen!`
Rust does not implement the `Default` for `[T; N]` where `N` > 32.  So if a solidity function has an input parameter with type `uint256[40]`, `abigen!` can not work well. 

Introduce a custom trait `CustomDefault`, which has the same behavior as the `Default` trait, except the way to initialize an array `[T; N]`. It simply assumes `T: Clone` and initializes `[T; N]` by `[T::custom_default(); N]`. Then implement `Default` for solidity structs/inputs by `CustomDefault`.

**Alternative 1:**
Manually implementing the `Default` trait in the procedure macro. 

**Alternative 2:**
Use a wrapped type for large fixed arrays. Like what you did for `Uint8`.

mattsse:
## Motivation
Closes #1640

this adds a check whether default can be derived for a call/output struct
rust-std auto derives default only for arrays len <=32

I think it's fine to skip the derive if not possible, since this generates code, the user still can implement it manually

## Solution
skip `derive(Default)` of a type contains a fixed array with more than 32 elements

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/contract/methods.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract/methods.rs
@@ -1,8 +1,9 @@
 use std::collections::{btree_map::Entry, BTreeMap, HashMap};
 
 use super::{types, util, Context};
-use crate::contract::common::{
-    expand_data_struct, expand_data_tuple, expand_param_type, expand_params,
+use crate::{
+    contract::common::{expand_data_struct, expand_data_tuple, expand_param_type, expand_params},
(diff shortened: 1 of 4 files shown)


gakonst (approved):
Reasonable

PR #1657, opened 2022-09-01
fix(providers): don't default to "latest" block ID for `eth_estimateGas`

tkporter:
## Motivation

Upon testing using a newer version of `ethers-rs`, I ran into issues estimating gas for Optimism chains (i.e. mainnet & optimism kovan). After some digging, it seem like this PR https://github.com/gakonst/ethers-rs/pull/1619 made a change to the way `eth_estimateGas` RPCs are constructed that breaks gas estimation for Optimism chains.

Previously, no block ID was provided when estimating gas, and this resulted params length of 1, where the element at index 0 was the tx to estimate. E.g. this works as expected against Optimism / OptimismKovan (this is just estimating a DAI transfer on mainnet Optimism):
```
$ curl -H "Content-Type: application/json" --data "{\"jsonrpc\":\"2.0\",\"method\":\"eth_estimateGas\",\"params\":[{\"from\":\"0x82e64f49ed5ec1bc6e43dad4fc8af9bb3a2312ee\",\"to\":\"0xDA10009cBd5D07dd0CeCc66161FC93D7c9000da1\",\"data\":\"0xa9059cbb000000000000000000000000dead00000000000000000000000000000000beef0000000000000000000000000000000000000000000000000000000000000069\"}],\"id\":67}" https://mainnet.optimism.io
{"jsonrpc":"2.0","result":"0xc760","id":67}
```

However, following https://github.com/gakonst/ethers-rs/pull/1619, the `eth_estimateGas` RPC will always specify the block ID, and will default to `"latest"`, resulting in a params length of 2. This works for most chains, but evidently not Optimism / Optimism Kovan :(.

E.g. this doesn't work-- it's the same RPC as above but with the block ID specified:
```
$ curl -H "Content-Type: application/json" --data "{\"jsonrpc\":\"2.0\",\"method\":\"eth_estimateGas\",\"params\":[{\"from\":\"0x82e64f49ed5ec1bc6e43dad4fc8af9bb3a2312ee\",\"to\":\"0xDA10009cBd5D07dd0CeCc66161FC93D7c9000da1\",\"data\":\"0xa9059cbb000000000000000000000000dead00000000000000000000000000000000beef0000000000000000000000000000000000000000000000000000000000000069\"},\"latest\"],\"id\":67}" https://mainnet.optimism.io
{"jsonrpc":"2.0","error":{"code":-32602,"message":"too many arguments, want at most 1"},"id":67}
```

From what I can tell, the behavior that https://github.com/gakonst/ethers-rs/pull/1619 introduced is spec compliant and Optimism isn't, but in any case it seems that ethers-rs should work correctly when using Optimism.

## Solution

If the block ID is `None`, instead of defaulting to `"latest"` in the RPC, instead the block ID is omitted.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-providers/src/provider.rs
+++ b/ethers-providers/src/provider.rs
@@ -574,8 +574,14 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
         block: Option<BlockId>,
     ) -> Result<U256, ProviderError> {
         let tx = utils::serialize(tx);
-        let block = utils::serialize(&block.unwrap_or_else(|| BlockNumber::Latest.into()));
-        self.request("eth_estimateGas", [tx, block]).await
+        // Some chains (e.g. Optimism) don't support a block ID being passed as a param,
+        // so refrain from defaulting to BlockNumber::Latest.
+        let params = if let Some(block_id) = block {
+            vec![tx, utils::serialize(&block_id)]
+        } else {
+            vec![tx]
+        };
+        self.request("eth_estimateGas", params).await
     }
 
     async fn create_access_list(
--- a/ethers-providers/src/provider.rs
+++ b/ethers-providers/src/provider.rs
@@ -574,7 +574,7 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
         block: Option<BlockId>,
     ) -> Result<U256, ProviderError> {
         let tx = utils::serialize(tx);
-        // Some chains (e.g. Optimism) don't support a block ID being passed as a param,
+        // Some nodes (e.g. old Optimism clients) don't support a block ID being passed as a param,
         // so refrain from defaulting to BlockNumber::Latest.
         let params = if let Some(block_id) = block {
             vec![tx, utils::serialize(&block_id)]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Unreleased
 
+- Stop defaulting to the `"latest"` block in `eth_estimateGas` params [#1657](https://github.com/gakonst/ethers-rs/pull/1657)
 - Fix geth trace types for debug_traceTransaction rpc
 - Fix RLP decoding of legacy `Transaction`
 - Fix RLP encoding of `TransactionReceipt` [#1661](https://github.com/gakonst/ethers-rs/pull/1661)


mattsse:
@ngotchac wdyt here?

ref https://github.com/foundry-rs/foundry/pull/3032

ngotchac:
I think this is the way to go, however I would argue that this should be done for other methods as well: 
- `get_transaction_count`
- `get_balance`
- `call`
- `create_access_list`
- `get_storage_at`
- `get_code`
- `get_proof`

tynes:
Fix in Optimism is here: https://github.com/ethereum-optimism/optimism/pull/3378

Its because Optimism is currently based off of an older geth fork that did not support the blocktag for `eth_estimateGas`. Once we upgrade to bedrock it will be very easy to keep up to date with upstream geth

tkporter:
Awesome!

Because it takes time to get RPC providers out there to upgrade, imo this change is probably still worth it. @mattsse @ngotchac wdyt? I'm happy to also make the change for those other methods if you'd like

ethers-providers/src/provider.rs
```diff
@@ -574,8 +574,14 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
         block: Option<BlockId>,
     ) -> Result<U256, ProviderError> {
         let tx = utils::serialize(tx);
-        let block = utils::serialize(&block.unwrap_or_else(|| BlockNumber::Latest.into()));
-        self.request("eth_estimateGas", [tx, block]).await
+        // Some chains (e.g. Optimism) don't support a block ID being passed as a param,
+        // so refrain from defaulting to BlockNumber::Latest.
```
mattsse, line 578: this is related to the optimism goerli bug and not actually accurate, right?

ngotchac:
> Awesome!
> 
> Because it takes time to get RPC providers out there to upgrade, imo this change is probably still worth it. @mattsse @ngotchac wdyt? I'm happy to also make the change for those other methods if you'd like

I think it just makes sense to do it for the other methods, yes, as they all have more or less the same format :)

gakonst:
given Optimism fixed this, should we close this?

gakonst (approved):
I think let's go with this given that it keeps happening on forks of Optimism etc. @tkporter do u mind rebasing to latest master

tkporter:
@gakonst yeah should be good now

gakonst (approved):

PR #1658, opened 2022-09-02
feat(core): improved debug format for Bytes

clifton:
## Motivation

Debug output for `Bytes` (eg calldata) currently looks like `Bytes(b"|\x02R\0\0\x93")` while debug format for other core types (eg Address) show a hex literal. This can make for a painful debugging experience when using external tools.

## Solution

Use hex literal output in `Bytes` `Debug` implementation eg `Bytes(0xdead)`

## PR Checklist

- [x] Added Tests
- [x] Updated the changelog


mattsse (approved):
reasonable,

there's also a display impl for bytes

https://github.com/foundry-rs/foundry/blob/4b9cca1790096c8bfdd50922bf22244983ad6e4f/anvil/src/eth/util.rs#L31-L68

which truncates based after certain len

clifton:
yeah i like that logic in foundry of truncating for `Display` but printing all the data for `Debug`

mattsse:
failing clippy lints: https://github.com/gakonst/ethers-rs/pull/1659

PR #1661, opened 2022-09-04
fix: incorrect encoding on TransactionReceipt

ncitron:
## Motivation
RLP encoding for `TransactionReceipt` is broken, since it directly encoded the `status` field, which is an optional.

## Solution
Use `rlp_opt` instead of `rlp_append` for the `status` field.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Unreleased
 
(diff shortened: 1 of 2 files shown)


mattsse (approved):
nice catch!

gakonst (approved):

PR #1665, opened 2022-09-04
perf(solc): read artifacts in parallel

mattsse:
## Motivation
manually identified that reading cached artifacts is a big part of pre-processing.

```console
2022-09-04T23:21:01.777551Z TRACE compile: ethers_solc::cache: reading artifacts from cache...
2022-09-04T23:21:03.309518Z TRACE compile: ethers_solc::cache: read 68 artifacts from cache
```

## Solution
* read artifacts in parallel
* Add Send+Sync restriction to `trait Artifact`, which shouldn't have any downsides

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


gakonst (approved):
Very nice

PR #1666, opened 2022-09-05
feat(contract): add extra Multicall helper methods

DaniPopes:
## Motivation

`Multicall` has some extra helper methods such as `getBlockHash` or `getBlockNumber`.

## Solution

Add these methods to `Multicall` in the same style of the already present `eth_balance_of`. Decided against adding the `allow_failure` param on most of these helper calls as they are base opcodes included in all EVM implementations (except `BASEFEE`) so they shouldn't fail if we're using the base Multicall contract.
Also did some small docs changes and refactoring.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Unreleased
 
-- Fix RLP encoding of `TransactionReceipt`
+- Fix RLP encoding of `TransactionReceipt` [#1661](https://github.com/gakonst/ethers-rs/pull/1661)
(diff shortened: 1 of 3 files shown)


mattsse (approved):
reasonable, lgtm

but should we prefix these calls with `add_`?

DaniPopes:
> reasonable, lgtm
> 
> but should we prefix these calls with `add_`?

Makes sense, following `add_call`, but changing `eth_balance_of` would be a breaking change

PR #1672, opened 2022-09-05
fix: legacy transaction rlp decoding

ncitron:
## Motivation
`Transaction` is improperly decoding legacy transactions as it assumes that transactions are always inside of 2718 envelopes.

## Solution
Check if the `transaction_type` field is `None` (signifying a legacy tx). In this case the full tx bytes should be passed to the decoder without snipping the first byte as is done for 2718 style txs.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-core/src/types/transaction/response.rs
+++ b/ethers-core/src/types/transaction/response.rs
@@ -344,9 +344,13 @@ impl Decodable for Transaction {
             true => Ok(Some(rlp.data()?.into())),
             false => Ok(None),
         }?;
-        let rest = rlp::Rlp::new(
-            rlp.as_raw().get(1..).ok_or(DecoderError::Custom("no transaction payload"))?,
-        );
+
+        let rest = match txn.transaction_type {
+            Some(_) => rlp::Rlp::new(
+                rlp.as_raw().get(1..).ok_or(DecoderError::Custom("no transaction payload"))?,
+            ),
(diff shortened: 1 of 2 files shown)


ethers-core/src/types/transaction/response.rs
```diff
@@ -344,9 +344,13 @@ impl Decodable for Transaction {
...
-        let rest = rlp::Rlp::new(
-            rlp.as_raw().get(1..).ok_or(DecoderError::Custom("no transaction payload"))?,
-        );
+
+        let rest = match txn.transaction_type {
+            Some(_) => rlp::Rlp::new(
+                rlp.as_raw().get(1..).ok_or(DecoderError::Custom("no transaction payload"))?,
+            ),
+            None => rlp.to_owned(),
+        };
```
mattsse, line 353: since none of the arms are used here I'd recommend a simple `if.txn.transaction_type.is_some(){}else{};` here

mattsse (approved):
ty,

smol style nit.

cc @Rjected

ethers-core/src/types/transaction/response.rs
ncitron: Updated!

Rjected (approved):
nice, lgtm!

gakonst (approved):
good catch.

PR #1683, opened 2022-09-09
fix(core): add derives, impls for Units

DaniPopes:
## Motivation

Currently it's not possible to pass `ethers_core::utils::Units` to its own `parse_units` due to the Copy trait bound

## Solution

Add derives for Units as well as some missing impls

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/utils/units.rs
+++ b/ethers-core/src/utils/units.rs
@@ -1,63 +1,122 @@
 use super::ConversionError;
+use std::{convert::TryFrom, fmt, str::FromStr};
 
 /// Common Ethereum unit types.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum Units {
-    /// Ether corresponds to 1e18 Wei
-    Ether,
-    /// Gwei corresponds to 1e9 Wei
-    Gwei,
-    /// Wei corresponds to 1 Wei
+    /// Wei is equivalent to 1 wei.
     Wei,
-    /// Use this for other less frequent unit sizes
+    /// Gwei is equivalent to 1e9 wei.
+    Gwei,
+    /// Ether is equivalent to 1e18 wei.
+    Ether,
+    /// Other less frequent unit sizes, equivalent to 1e{0} wei.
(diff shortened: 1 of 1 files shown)


ethers-core/src/utils/units.rs
```diff
@@ -1,63 +1,122 @@
...
             "ether" => Units::Ether,
             "gwei" => Units::Gwei,
             "wei" => Units::Wei,
-            _ => return Err(ConversionError::UnrecognizedUnits(src.to_string())),
+            _ => return Err(ConversionError::UnrecognizedUnits(s.to_string())),
         })
     }
 }
+
+/* ----------------------------------------- From<Units> ---------------------------------------- */
```
mattsse, line 78:
```suggestion
```
this divider is not necessary imo, since this file only contains the `Units` enum

ethers-core/src/utils/units.rs
```diff
@@ -1,63 +1,122 @@
...
+    }
+}
+
+impl TryFrom<&str> for Units {
+    type Error = ConversionError;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        Self::from_str(value)
+    }
+}
```
mattsse, line 63: I can't remember if there was an issue with `TryFrom<T:AsRef<str>>` in general, so not sure if this would work.

if `TryFrom<T:AsRef<str>>` is not possible we should also add `impl<'a> TryFrom<&'a String>`

mattsse (changes requested):
nice!

ethers-core/src/utils/units.rs
DaniPopes: Can't figure out how to bound `T` to `AsRef<str>`, getting T not found or conflicting implementations of TryFrom

mattsse: yeh there's apparently an issue with blanket tryfrom impls which doesn't allow this

mattsse (approved):

gakonst (approved):

PR #1687, opened 2022-09-11
fix: serialize slot as h256 66 byte string

mattsse:
## Motivation
Ref https://github.com/foundry-rs/foundry/issues/2687

hardhat recently enforced 66byte slot argument for `eth_getStorageAt` https://github.com/NomicFoundation/hardhat/pull/2581 because:

> According to [the spec](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/execution-apis/assembled-spec/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false), the storage slot in eth_getStorageAt should have a length of 66 (0x + 32 bytes).

this is a bit confusing to me because this spec https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getstorageat says it's a [QUANTITY](https://ethereum.org/en/developers/docs/apis/json-rpc/#quantities-encoding)...

~~this is not very clear to me~~

resolved here https://github.com/NomicFoundation/hardhat/pull/2581#issuecomment-1242956716

we should merge this to be on par with hardhat

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-providers/src/provider.rs
+++ b/ethers-providers/src/provider.rs
@@ -734,7 +734,7 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
     async fn get_storage_at<T: Into<NameOrAddress> + Send + Sync>(
         &self,
         from: T,
-        location: H256,
+        slot: H256,
         block: Option<BlockId>,
     ) -> Result<H256, ProviderError> {
         let from = match from.into() {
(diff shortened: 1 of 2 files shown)


gakonst:
Suggest we hold off until https://github.com/NomicFoundation/hardhat/pull/2581#issuecomment-1243043983 or https://github.com/ethereum/execution-apis/issues/303 gets some reply? Not obvious to me what's right here.

mattsse:
sgtm, 

reasonable

aathan:
Yes, I would suggest holding off. I just added a comment to https://github.com/NomicFoundation/hardhat/pull/2581#issuecomment-1243043983 where I looked into the geth source and it does *not* enforce a size. I'm pretty sure all of this stems from an incorrect spec at https://github.com/NomicFoundation/hardhat/issues/2230 ... which says {64} instead of {0,64}.

I posted a bug to the spec project also suggesting this should say {1,64} anyway.

aathan:
Perhaps the foundry implementation SHOULD retry the request with 0 padding if the unpadded request fails. I believe older versions of hardhat contained an issue that would *reject* values with leading zeros where QUANTITY is required (as mentioned in https://github.com/NomicFoundation/hardhat/issues/2230). Perhaps foundry could tag each RPC URL provider as either "leading zeros required" or "leading zeros rejected" and act accordingly... this would perhaps allow it to be compatible in all cases, given there seems to be ambiguity around this aspect of the spec.

aathan:
FYI, I tried running foundry's script features vs hardhat 2.9.4 (pre-dating their "fix") and got this error:

```
eth_feeHistory

  Remote node did not answer to eth_feeHistory correctly

eth_feeHistory

  Errors encountered in param 0: Invalid value 10 supplied to : QUANTITY
```

Not sure what about the parameter was considered invalid (formatting, or something about "10" ... I'm pretty certain the chain was longer than 10 blocks given my hardhat config is forking a public chain).

aathan:
If I do `cargo build --release` in a master checkout I end up with a forge artifact. If I do the same out of this PR I do not. No obvious error is emitted (but maybe something scrolled off screen?). Not a Rust guy, happy to report on output if given instructions.

aathan:
Just making sure you saw: https://github.com/NomicFoundation/hardhat/pull/3177 which closes https://github.com/NomicFoundation/hardhat/issues/3164 .

mattsse:
this appears to be solved, thx @aathan

PR #1689, opened 2022-09-11
chore: inherit stderr for anvil

mattsse:
## Motivation
reading from stderr and stdout continuously is surprisingly hard because both are blocking operations.
this will inherit stderr to at least show the panic message if it panics, alternatively we could spawn an stderr reader to another thread and read the stderr output there, but this will also block and hang if anvil never prints to stderr, so unsure if worth it to have a background thread that may live for the entire program just for monitoring the anvil launch...

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/utils/anvil.rs
+++ b/ethers-core/src/utils/anvil.rs
@@ -214,7 +214,7 @@ impl Anvil {
         } else {
             Command::new("anvil")
         };
-        cmd.stdout(std::process::Stdio::piped());
+        cmd.stdout(std::process::Stdio::piped()).stderr(std::process::Stdio::inherit());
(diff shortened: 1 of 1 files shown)


gakonst:
Can you please explain what the current issue is that we're having? Is this performance related on Anvil right now?

mattsse:
launching anvil may not be successful, for example if the port is already in use, in which case it either times out or blocks on reading from stdour indefinitely.

this will make sure that the panic is at least printed to to stderr via inherit so the panic is visibile

gakonst (approved):

PR #1691, opened 2022-09-12
fix(solc): remove default include paths

rkrasiuk:
## Motivation

currently, the default `include_path`s contain the sources, test & script dirs, which lead to the possibility of ambiguous imports like `import 'interface/IContract.sol';` given the project tree
```text
- src
-- Contract.sol
-- interface
--- IContract.sol
```

while it is possible to make the `Graph` resolve such import paths properly, it leads to ambiguity while trying to reconstruct sources for standard json input, since the same node can be imported as `interface/IContract.sol` and `src/interface/IContract.sol`

more info [here](https://docs.soliditylang.org/en/v0.8.13/path-resolution.html#initial-content-of-the-virtual-filesystem)

## Solution

remove default include paths since they are all covered by the base path

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/config.rs
+++ b/ethers-solc/src/config.rs
@@ -88,9 +88,8 @@ impl ProjectPathsConfig {
     ///
     /// See [IncludePaths]
     pub fn include_paths(&self) -> Vec<PathBuf> {
-        // Note: root must not be included, since it will be used as base-path, which would be a
-        // conflict
-        vec![self.sources.clone(), self.tests.clone(), self.scripts.clone()]
+        // Note: sources, tests and scripts are covered by the base-path
+        Default::default()
     }
 
     /// Creates all configured dirs and files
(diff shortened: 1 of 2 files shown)


ethers-solc/src/config.rs
```diff
@@ -266,6 +266,15 @@ impl ProjectPathsConfig {
                         }
                     }
                 }
+                // if still unable to resolve, attempt to find the absolute import
+                // inside the project paths that's possible due to `--include-dir`
+                for path in [&self.sources, &self.tests, &self.scripts] {
```
mattsse, line 271: looping over `include_paths` should be equivalent here, this also has the benefit that we don't add `sources`, `tests` `scripts` as hardcoded include-paths

mattsse (changes requested):

ethers-solc/src/config.rs
rkrasiuk: i'm sus, because `include_paths` are a mutable ref here and get modified on every invocation

mattsse: I see, good point,

I think this is fine then, the worst that can happen is that the graph resolves a file but solc can't in which case we still fail as expected.

mattsse (changes requested):
this is the correct fix, 

absolute import paths like `src/...` are already covered by `--base-path <root>` so we don't need to (or shouldn't) add those paths.

but looks like we can remove this function entirely?

gakonst (changes requested):
yeah let's remove the function and remove the usage of this from foundry?

rkrasiuk:
@mattsse @gakonst removed
> remove the usage of this from foundry?
are you talking about disabling custom `include_paths` altogether?
https://github.com/foundry-rs/foundry/blob/433be903a02be0e49c37deacd4f3fe4d71b4c97c/config/src/lib.rs#L155

mattsse:
this we should keep, since this is a valid solc customization, but remove the default paths in `ProjectPathConfig`

mattsse (approved):

gakonst (approved):

PR #1712, opened 2022-09-18
docs: add MSRV

DaniPopes:
## Motivation

When users try to build Foundry or a crate using this library with an older Rust version than 1.62 (found using [cargo-msrv](https://crates.io/crates/cargo-msrv)), cargo fails the build with "unstable errors" (see examples below), and only after building the rest of the dependencies. These ambiguous errors can be avoided by adding an MSRV (Minimum Supported Rust Version) as cargo will immediatly fail the build if the toolchain version is less than the MSRV.

## Solution

Add `rust-version` key in /**/Cargo.toml and the `msrv` key in /.clippy.toml

The other changes are by my formatter, let me know if you want me to revert them

### Before, using rustc v1.60.0:

image

image

### After:

image

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse (approved):
I think that's reasonable, this should prevent additional "could not compile" issues in the future

gakonst (approved):

PR #1720, opened 2022-09-19
feat(core): add utility methods to NameOrAddress

DaniPopes:
## Motivation

@gakonst https://t.me/ethers_rs/20599

## Solution

- fix `impl Encodable for NameOrAddress` wrong cmp order
- add as_X -> X by cloning to NameOrAddress
- add to_addr to TransactionRequest
- add more impls and cleanup

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/types/ens.rs
+++ b/ethers-core/src/types/ens.rs
@@ -1,10 +1,9 @@
-use std::cmp::Ordering;
-
 use crate::types::Address;
 use rlp::{Decodable, Encodable, RlpStream};
 use serde::{ser::Error as SerializationError, Deserialize, Deserializer, Serialize, Serializer};
+use std::{cmp::Ordering, convert::Infallible, str::FromStr};
 
-/// ENS name or Ethereum Address. Not RLP encoded/serialized if it's a name
+/// ENS name or Ethereum Address. Not RLP encoded/serialized if it's a name.
 #[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


ethers-core/src/types/ens.rs
```diff
@@ -85,8 +72,57 @@ impl<'de> Deserialize<'de> for NameOrAddress {
...
+    /// Maps Address(&a) to Some(a) by cloning the value and Name to None.
+    pub fn as_address(&self) -> Option<Address> {
+        match self {
+            Self::Address(a) => Some(*a),
+            Self::Name(_) => None,
+        }
+    }
+
+    /// Maps Name(&n) to Some(n) by cloning the value and Address to None.
+    pub fn as_name(&self) -> Option<String> {
```
mattsse, line 121: this should return `&str` so caller can decided whether to clone

mattsse (changes requested):
smol nit, otherwise lgtm

ethers-core/src/types/ens.rs
DaniPopes: Ok, same for `as_address`?

mattsse: `Address` is `Copy` so nbd there, but for consistency maybe it makes sense

gakonst (approved):
beautiful - clean

PR #1728, opened 2022-09-23
fix(contract): relax Middleware trait bound for getters

DaniPopes:
## Motivation

`Contract`'s getters are bound to the `Middleware` trait which blocks `Debug` from being derived automatically for structs that hold abigen'd contracts, eg:

```rust
abigen!(MyContract, "...")
// Clone works fine
// Cannot derive Debug: 'M: Middleware is not satisfied'
#[derive(Clone, Debug)]
struct<M> MyStruct<M> {
  contract: MyContract<M>
}

```

## Solution

Move to new impl block


mattsse (approved):
lgtm

gakonst:
reasonable change - thx

Issue #1713, opened 2022-09-18
Cyclic deps


DaniPopes:
Rust Analyzer throws errors due to cyclic deps in `ethers-derive-eip712` and `ethers-middleware`/`ethers-contract`. Already mentioned in https://github.com/gakonst/ethers-rs/pull/481#issuecomment-938045421, although I don't think the follow-up PR happened ^^

Log:

```
[ERROR project_model::workspace] cyclic deps: ethers_derive_eip712(CrateId(152)) -> ethers_contract(CrateId(147)), alternative path: ethers_contract(CrateId(147)) -> ethers_derive_eip712(CrateId(152))
[ERROR project_model::workspace] cyclic deps: ethers_middleware(CrateId(155)) -> ethers_contract(CrateId(147)), alternative path: ethers_contract(CrateId(147)) -> ethers_middleware(CrateId(155))
[ERROR project_model::workspace] cyclic deps: ethers_signers(CrateId(165)) -> ethers_contract(CrateId(147)), alternative path: ethers_contract(CrateId(147)) -> ethers_derive_eip712(CrateId(152)) -> ethers_signers(CrateId(165))
[ERROR project_model::workspace] cyclic deps: ethers_signers(CrateId(165)) -> ethers_derive_eip712(CrateId(152)), alternative path: ethers_derive_eip712(CrateId(152)) -> ethers_signers(CrateId(165))
[ERROR project_model::workspace] cyclic deps: ethers_derive_eip712(CrateId(152)) -> ethers_contract(CrateId(147)), alternative path: ethers_contract(CrateId(147)) -> ethers_derive_eip712(CrateId(152))
[ERROR project_model::workspace] cyclic deps: ethers_middleware(CrateId(155)) -> ethers_contract(CrateId(147)), alternative path: ethers_contract(CrateId(147)) -> ethers_middleware(CrateId(155))
[ERROR project_model::workspace] cyclic deps: ethers_signers(CrateId(165)) -> ethers_contract(CrateId(147)), alternative path: ethers_contract(CrateId(147)) -> ethers_derive_eip712(CrateId(152)) -> ethers_signers(CrateId(165))
[ERROR project_model::workspace] cyclic deps: ethers_signers(CrateId(165)) -> ethers_derive_eip712(CrateId(152)), alternative path: ethers_derive_eip712(CrateId(152)) -> ethers_signers(CrateId(165))
```

DaniPopes:
[...]

PR #1730, opened 2022-09-24
fix(contract, signers): cyclic deps

DaniPopes:
## Motivation

https://github.com/gakonst/ethers-rs/issues/1713#issuecomment-1250400755

## Solution

- contract: Remove middleware, not needed since accounts are unlocked on Anvil
- signers: contract import was used only for the derive macros

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1429,7 +1429,7 @@ dependencies = [
  "coins-ledger",
  "elliptic-curve",
  "eth-keystore",
- "ethers-contract",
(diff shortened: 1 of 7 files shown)


mattsse (approved):
ty, using the derive crate directly, is indeed the fix here

gakonst (approved):
ah beautiful

PR #1731, opened 2022-09-24
contract: return `Arc<M>` instead of `&M`

DaniPopes:
## Motivation

https://t.me/ethers_rs/20735?thread=20735

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


mattsse (approved):
reasonable, also shouldn't break sine `Arc<M>` is also Middleware

PR #1733, opened 2022-09-24
fix: legacy signed rlp decoding

### LINKED ISSUE #1732: Recovering sender of a legacy (type 0) transaction fails
**Version**

development ethers version:
`ethers v0.17.0 (https://github.com/gakonst/ethers-rs?rev=afdab2a555c10d55a32c5991282b4353a4a81256#afdab2a5)`

**Description**

Seems like the code to recover the sender of a transaction doesn't work if it's a legacy transaction.

Let's use https://etherscan.io/tx/0x280cde7cdefe4b188750e76c888f13bd05ce9a4d7767730feefe8a0e50ca6fc4 as an example, we can access the raw RLP in Etherscan: https://etherscan.io/getRawTx?tx=0x280cde7cdefe4b188750e76c888f13bd05ce9a4d7767730feefe8a0e50ca6fc4

Here is a test to reproduce:

```
#[test]
fn test_rlp_decoding() {
    use ethers::types::*;

    let raw_tx = "0xf9015482078b8505d21dba0083022ef1947a250d5630b4cf539739df2c5dacb4c659f2488d880c46549a521b13d8b8e47ff36ab50000000000000000000000000000000000000000000066ab5a608bd00a23f2fe000000000000000000000000000000000000000000000000000000000000008000000000000000000000000048c04ed5691981c42154c6167398f95e8f38a7ff00000000000000000000000000000000000000000000000000000000632ceac70000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000006c6ee5e31d828de241282b9606c8e98ea48526e225a0c9077369501641a92ef7399ff81c21639ed4fd8fc69cb793cfa1dbfab342e10aa0615facb2f1bcf3274a354cfe384a38d0cc008a11c2dd23a69111bc6930ba27a8";
    let mut decoded_tx = ethers::utils::rlp::decode::<Transaction>(
        &raw_tx.parse::<Bytes>().expect("unable to parse raw tx"),
    )
    .expect("unable to decode raw tx");
    decoded_tx.recover_from_mut().unwrap();
    decoded_tx.hash = decoded_tx.hash();
    dbg!(&decoded_tx);
    assert_eq!(
        decoded_tx.from,
        H160(
            b"\xa1\x2e\x14\x62\xd0\xce\xd5\x72\xf3\x96\xf5\x8b\x6e\x2d\x03\x89\x4c\xd7\xc8\xa4"
                .clone()
        )
    );
    assert_eq!(decoded_tx.value, U256::from(884487398604084184u64));
}
```

The test fails, but the `from` address that I check for is the sender of the transaction (Etherscan confirms), so it means that ethers is incorrectly recovering the sender.

mattsse:
## Motivation
Closes #1732

fixes some issues around signed transaction rlp decoding.

* Use conditional stream size depending on chain_id
* fix `TypedTransaction::decoded_signed` to support legacy transactions: replace the wrong `is_data()` check and treat lists as legacy transactions

cc @Rjected 

imo the rlp decode_unsigned/decode_signed are a bit confusing. especially since the unsiged variant is the default, which would lead to wrong chain ids (37 instead 1 for mainnet). Not sure what to do about it.

need to make this very clear in the docs at least.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/types/transaction/eip2718.rs
+++ b/ethers-core/src/types/transaction/eip2718.rs
@@ -320,33 +320,31 @@ impl TypedTransaction {
 
     /// Decodes a signed TypedTransaction from a rlp encoded byte stream
     pub fn decode_signed(rlp: &rlp::Rlp) -> Result<(Self, Signature), TypedTransactionError> {
-        let tx_type: Option<U64> = match rlp.is_data() {
-            true => Ok(Some(rlp.data()?.into())),
-            false => Err(TypedTransactionError::MissingTransactionType),
-        }?;
+        let data = rlp.data()?;
+        let first = *data.first().ok_or(rlp::DecoderError::Custom("empty slice"))?;
+        if rlp.is_list() {
+            // Legacy (0x00)
+            // use the original rlp
+            let decoded_request = TransactionRequest::decode_signed_rlp(rlp)?;
+            return Ok((Self::Legacy(decoded_request.0), decoded_request.1))
+        }
 
         let rest = rlp::Rlp::new(
             rlp.as_raw().get(1..).ok_or(TypedTransactionError::MissingTransactionPayload)?,
         );
 
-        match tx_type {
-            Some(x) if x == U64::from(1u64) => {
-                // EIP-2930 (0x01)
-                let decoded_request = Eip2930TransactionRequest::decode_signed_rlp(&rest)?;
-                Ok((Self::Eip2930(decoded_request.0), decoded_request.1))
-            }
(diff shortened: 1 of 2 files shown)


mattsse:
looks like the celo test wallet is out of funds

gakonst (approved):
funded `0x4B3285236D38a29229975B2D928E103f52cB499D` via https://celo.org/developers/faucet

Issue #1736, opened 2022-09-24
etherscan: deserialize a contract's SourceCode field


DaniPopes:
Sort of implemented for sources already: 
https://github.com/gakonst/ethers-rs/blob/75d7f452318b108f63f459b26c6e89dc6452ce06/ethers-etherscan/src/contract.rs#L162-L165
https://github.com/gakonst/ethers-rs/blob/75d7f452318b108f63f459b26c6e89dc6452ce06/ethers-etherscan/src/contract.rs#L182-L185

But the rest of the options (see example) are also necessary for something like creating a `Solc` project (foundry-rs/foundry#3335)

### Example

Query: https://api.etherscan.io/api?module=contract&action=getsourcecode&address=0xDef1C0ded9bec7F1a1670819833240f027b25EfF

Link: https://etherscan.io/address/0xDef1C0ded9bec7F1a1670819833240f027b25EfF#code

Sources and ABI omitted

Raw:

```json
{
    "SourceCode": "{{\r\n  \"language\": \"Solidity\",\r\n  \"sources\": {...},\r\n  \"settings\": {\r\n    \"remappings\": [\r\n      \"@0x/contracts-utils=/home/cluracan/code/0x-monorepo/node_modules/@0x/contracts-utils\",\r\n      \"@0x/contracts-erc20=/home/cluracan/code/0x-monorepo/node_modules/@0x/contracts-erc20\"\r\n    ],\r\n    \"optimizer\": {\r\n      \"enabled\": true,\r\n      \"runs\": 1000000,\r\n      \"details\": {\r\n        \"yul\": true,\r\n        \"deduplicate\": true,\r\n        \"cse\": true,\r\n        \"constantOptimizer\": true\r\n      }\r\n    },\r\n    \"outputSelection\": {\r\n      \"*\": {\r\n        \"*\": [\r\n          \"evm.bytecode\",\r\n          \"evm.deployedBytecode\",\r\n          \"abi\"\r\n        ]\r\n      }\r\n    },\r\n    \"evmVersion\": \"istanbul\"\r\n  }\r\n}}",
    "ABI": "[...]",
    "ContractName": "ZeroEx",
    "CompilerVersion": "v0.6.8+commit.0bbfe453",
    "OptimizationUsed": "1",
    "Runs": "1000000",
    "ConstructorArguments": "",
    "EVMVersion": "Default",
    "Library": "",
    "LicenseType": "Apache-2.0",
    "Proxy": "1",
[...]

PR #1747, opened 2022-09-27
feat(etherscan): parse SourceCode field

DaniPopes:
## Motivation

#1736

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/ethers-etherscan/Cargo.toml
+++ b/ethers-etherscan/Cargo.toml
@@ -18,8 +18,8 @@ Rust API bindings for the etherscan.io web API
 keywords = ["ethereum", "web3", "etherscan", "ethers"]
 
 [dependencies]
-ethers-solc = { version = "^0.17.0", path = "../ethers-solc", default-features = false }
 ethers-core = { version = "^0.17.0", path = "../ethers-core", default-features = false }
+ethers-solc = { version = "^0.17.0", path = "../ethers-solc", default-features = false }
 reqwest = { version = "0.11.12", default-features = false, features = ["json"] }
 serde = { version = "1.0.124", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.64", default-features = false }
@@ -37,7 +37,6 @@ tempfile = "3.3.0"
 tokio = { version = "1.18", features = ["macros", "rt-multi-thread", "time"] }
 serial_test = "0.9.0"
 tracing-subscriber = { version = "0.3", default-features = false, features = ["env-filter", "fmt"] }
-ethers-solc = { version = "^0.17.0", path = "../ethers-solc", default-features = false }
 
 [package.metadata.docs.rs]
 all-features = true
--- a/ethers-etherscan/Cargo.toml
+++ b/ethers-etherscan/Cargo.toml
@@ -18,6 +18,7 @@ Rust API bindings for the etherscan.io web API
 keywords = ["ethereum", "web3", "etherscan", "ethers"]
 
 [dependencies]
+ethers-solc = { version = "^0.17.0", path = "../ethers-solc", default-features = false }
 ethers-core = { version = "^0.17.0", path = "../ethers-core", default-features = false }
 reqwest = { version = "0.11.12", default-features = false, features = ["json"] }
 serde = { version = "1.0.124", default-features = false, features = ["derive"] }
--- a/ethers-etherscan/Cargo.toml
+++ b/ethers-etherscan/Cargo.toml
@@ -19,7 +19,7 @@ keywords = ["ethereum", "web3", "etherscan", "ethers"]
 
 [dependencies]
 ethers-core = { version = "^0.17.0", path = "../ethers-core", default-features = false }
-ethers-solc = { version = "^0.17.0", path = "../ethers-solc", default-features = false }
+ethers-solc = { version = "^0.17.0", path = "../ethers-solc", default-features = false, optional = true }
 reqwest = { version = "0.11.12", default-features = false, features = ["json"] }
 serde = { version = "1.0.124", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.64", default-features = false }
@@ -43,6 +43,7 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-default = ["rustls"]
+default = ["rustls", "solc"]
 openssl = ["reqwest/native-tls"]
 rustls = ["reqwest/rustls-tls"]
+solc = ["ethers-solc"]
--- a/ethers-etherscan/Cargo.toml
+++ b/ethers-etherscan/Cargo.toml
@@ -37,6 +37,7 @@ tempfile = "3.3.0"
 tokio = { version = "1.18", features = ["macros", "rt-multi-thread", "time"] }
 serial_test = "0.9.0"
 tracing-subscriber = { version = "0.3", default-features = false, features = ["env-filter", "fmt"] }
+ethers-solc = { version = "^0.17.0", path = "../ethers-solc", default-features = false }
 
 [package.metadata.docs.rs]
 all-features = true
--- a/ethers-etherscan/Cargo.toml
+++ b/ethers-etherscan/Cargo.toml
@@ -43,7 +43,6 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-default = ["rustls", "solc"]
+default = ["rustls"]
 openssl = ["reqwest/native-tls"]
 rustls = ["reqwest/rustls-tls"]
-solc = ["ethers-solc"]
--- a/ethers-etherscan/src/contract.rs
+++ b/ethers-etherscan/src/contract.rs
@@ -1,14 +1,16 @@
 use crate::{
     source_tree::{SourceTree, SourceTreeEntry},
+    utils::{deserialize_address_opt, deserialize_string_or_struct, deserialize_version},
     Client, EtherscanError, Response, Result,
 };
 use ethers_core::{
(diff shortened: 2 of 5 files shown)


ethers-etherscan/src/contract.rs
```diff
@@ -347,14 +422,38 @@ impl Client {
...
+            // the SourceCode string value
+            let range = start_idx + 13..end_idx + 3;
+            let source_code = &res[range.clone()];
+            // parse the escaped characters
+            let parsed: String = serde_json::from_str(source_code)?;
+            // skip the first `{` and last `}`
+            let source_code = &parsed[1..parsed.len() - 1];
+            // replace
+            res.replace_range(range, source_code);
+        }
```
mattsse, line 454: can we use untagged enum here instead?

mattsse (changes requested):
some suggestions and qs

ethers-etherscan/src/lib.rs
```diff
@@ -155,55 +155,60 @@ impl Client {
...
+    /// Execute a POST request with a form, without sanity checking the response.
+    async fn post<F: Serialize>(&self, form: &F) -> Result<String> {
+        trace!(target: "etherscan", "POST {}", self.etherscan_api_url);
         let response = self
             .client
-            .get(self.etherscan_api_url.clone())
-            .header(header::ACCEPT, "application/json")
-            .query(query)
+            .post(self.etherscan_api_url.clone())
+            .form(form)
```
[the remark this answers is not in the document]
DaniPopes: yeah .form method adds the header

ethers-etherscan/Cargo.toml
```diff
@@ -19,6 +19,7 @@ keywords = ["ethereum", "web3", "etherscan", "ethers"]
 
 [dependencies]
 ethers-core = { version = "^0.17.0", path = "../ethers-core", default-features = false }
+ethers-solc = { version = "^0.17.0", path = "../ethers-solc", default-features = false }
```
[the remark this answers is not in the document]
DaniPopes: `SourceCodeMetadata.settings` field is exactly a `Settings` from `ethers-solc`

ethers-etherscan/src/contract.rs
DaniPopes: No the field is always a string, and have to check manually if it starts with `{{` / ends with `}}`. It's a bit hacky but couldn't figure out another way

ethers-etherscan/src/contract.rs
```diff
@@ -154,98 +293,34 @@ impl IntoIterator for ContractMetadata {
...
-    #[serde(rename = "LicenseType")]
-    pub license_type: String,
-    #[serde(rename = "Proxy")]
-    pub proxy: String,
-    #[serde(rename = "Implementation")]
-    pub implementation: String,
-    #[serde(rename = "SwarmSource")]
-    pub swarm_source: String,
-}
+/* ------------------------------------------- Client ------------------------------------------- */
```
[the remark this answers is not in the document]
DaniPopes: why, that file's pretty crowded :(

ethers-etherscan/src/contract.rs
mattsse: it's not obvious to me how the two variants differ.

1. the value of `SourceCode ` is a string
2. the value of `SourceCode ` is a json object
?

then let's just create a new type with a source_code field and impl deserialize manually instead of hacking with indices here

ethers-etherscan/src/contract.rs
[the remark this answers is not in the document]
mattsse: feel free to move to new submodule.

shouldn't start randomly adding new header styles

ethers-etherscan/src/contract.rs
DaniPopes: response.result[n].SourceCode is either:
- a string containing all of the source code
- a string containing a JSON object wrapped in an extra set of `{}`
not sure how to implement `Deserialize` manually since in both cases it's a string; we always have to parse it as that first, then check for a JSON, then deserialize that one too

mattsse: I'd just use an untagged enum here for convenience, where the json variant is the first attempt.

ethers-etherscan/Cargo.toml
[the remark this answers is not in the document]
DaniPopes: can make the new parsing and ethers-solc optional with a feature, wdyt

ethers-etherscan/src/contract.rs
```diff
@@ -1,141 +1,240 @@
...
-            self.optimized()
-        } else {
-            self.not_optimized()
+    pub fn sources(&self) -> HashMap<String, SourceCodeEntry> {
+        match self {
+            Self::Metadata { sources, .. } => sources.clone(),
+            Self::SourceCode(s) => {
+                let mut sources = HashMap::with_capacity(1);
+                sources.insert("Contract".into(), s.into());
+                sources
```
mattsse, line 78: can use `HashMap::from([(,,)])` here

ethers-etherscan/src/contract.rs
```diff
@@ -1,141 +1,240 @@
...
 
-    #[must_use]
-    pub fn optimized(mut self) -> Self {
-        self.optimization_used = Some("1".to_string());
-        self
+    pub fn settings(&self) -> &Option<Settings> {
+        match self {
+            Self::Metadata { settings, .. } => settings,
+            Self::SourceCode(_) => &None,
+        }
```
mattsse, line 87:
```suggestion
    pub fn settings(&self) -> Option<&Settings> {
        match self {
            Self::Metadata { settings, .. } => settings.as_ref(),
            Self::SourceCode(_) => None,
        }
```

ethers-etherscan/src/contract.rs
```diff
@@ -1,141 +1,240 @@
...
+    Metadata {
+        /// Programming language of the sources.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        language: Option<SourceCodeLanguage>,
+        /// Source path => source code
+        #[serde(default)]
+        sources: HashMap<String, SourceCodeEntry>,
+        /// Compiler settings, None if the language is not Solidity.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        settings: Option<Settings>,
```
mattsse, line 49: I wonder if we should just keep this as serde_json::Value and feature get the `fn settings -> Settings` function so we can make ethers-sol optional

mattsse (changes requested):
this looks pretty good. thanks for adding all those tests!

a few nits, and suggestion re `Settings` and ethers-solc dependency

ethers-etherscan/Cargo.toml
```diff
@@ -43,6 +43,7 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-default = ["rustls"]
+default = ["rustls", "solc"]
 openssl = ["reqwest/native-tls"]
 rustls = ["reqwest/rustls-tls"]
+solc = ["ethers-solc"]
```
mattsse, line 49: ~~I don't think we need to add a name feature here, can use `ethers-solc` directly, I think.~~

hmm, I just realized `ethers` has `ethers-solc` as non-default, but if we enable it here by default then this won't have any effect.

so I think we should just keep it optional, disabled by default, and instead enable it in `ethers` main cargo toml if solc feature is enabled there

ethers-etherscan/src/contract.rs
```diff
@@ -9,11 +9,13 @@ use ethers_core::{
     abi::{Abi, Address},
     types::{serde_helpers::deserialize_stringified_u64, Bytes},
 };
-use ethers_solc::{artifacts::Settings, EvmVersion, Project, ProjectBuilder, SolcConfig};
 use semver::Version;
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, path::Path};
 
+#[cfg(feature = "solc")]
```
mattsse, line 16:
```suggestion
#[cfg(feature = "ethers-solc")]
```

ethers-etherscan/src/contract.rs
```diff
@@ -72,18 +74,32 @@ impl SourceCodeMetadata {
...
-                let mut sources = HashMap::with_capacity(1);
-                sources.insert("Contract".into(), s.into());
-                sources
-            }
+            Self::SourceCode(s) => HashMap::from([("Contract".into(), s.into())]),
         }
     }
 
-    pub fn settings(&self) -> &Option<Settings> {
+    #[cfg(feature = "solc")]
```
mattsse, line 81:
```suggestion
    #[cfg(feature = "ethers-solc")]
```

ethers-etherscan/src/contract.rs
```diff
@@ -72,18 +74,32 @@ impl SourceCodeMetadata {
...
+                        Ok(Some(serde_json::from_value(value.to_owned())?))
+                    }
+                }
+                None => Ok(None),
+            },
+            Self::SourceCode(_) => Ok(None),
+        }
+    }
+
+    #[cfg(not(feature = "solc"))]
```
mattsse, line 98:
```suggestion
    #[cfg(not(feature = "ethers-solc"))]
```

mattsse (changes requested):

ethers-etherscan/Cargo.toml
DaniPopes: I think `cfg(feature = "ethers-solc")` is enabled whenever the crate is built with the `ethers-solc` dependency so this should work

mattsse (approved):

PR #1757, opened 2022-09-29
fix(core, etherscan): RawAbi and Abi

DaniPopes:
## Motivation

high priority since cast interface stuff is broken @gakonst thx ser

## Solution

- move (only) rawabi as raw.rs to core::abi, testdata not moved (abi jsons and contracts)
- revert etherscan parsing as Abi, add methods to conditionally parse as RawAbi or Abi

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog


mattsse (approved):
tysm for taking this on!

Issue #1676, opened 2022-09-07
Docs for Event type are broken


dbelv:
**Description**
On docs.rs for release 0.17.0 the `Event` type is broken, it doesn't seem to link anywhere:
https://docs.rs/ethers-contract/0.17.0/ethers_contract/struct.Contract.html#method.event

I would have expected that it would have lead to:
https://docs.rs/ethers-contract/0.17.0/src/ethers_contract/event.rs.html#113

PR #1787, opened 2022-10-16
docs: fix Event type not pointing to docs 

mouseless-eth:
## Motivation

Addressing the issue #1676

## Solution

The event builder type is part of the [builders mod in lib.rs](https://github.com/gakonst/ethers-rs/blob/a07581489a12b1007c3a261dc8df2bbdc4e27918/ethers-contract/src/lib.rs#L44) which has the `#[doc(hidden)]` attribute. Moving Event into scope generates docs for the Event type and fixes the [dead Event type link](https://docs.rs/ethers-contract/0.17.0/ethers_contract/struct.Contract.html#method.event) under the Contract documentation. 

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog

### DIFF
--- a/ethers-contract/src/lib.rs
+++ b/ethers-contract/src/lib.rs
@@ -18,7 +18,7 @@ mod factory;
 pub use factory::{ContractDeployer, ContractFactory};
 
 mod event;
-pub use event::EthEvent;
+pub use event::{EthEvent, Event};
 
 mod log;
 pub use log::{decode_logs, EthLogDecode, LogMeta};


mattsse (commented):
while this makes it available for the docs, I think the proper fix for this particular issue would be to adjust the path used in the docs and point that to the right source, should be `ethers_core` probably

gakonst:
I think OK to just re-export this given the docs problems people have been having

PR #1793, opened 2022-10-17
fix(solc): emit empty node vec

mattsse:
closes https://github.com/foundry-rs/foundry/issues/3504

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/types/block.rs
+++ b/ethers-core/src/types/block.rs
@@ -570,9 +570,9 @@ impl BlockNumber {
     pub fn is_finalized(&self) -> bool {
         matches!(self, BlockNumber::Finalized)
     }
-    
+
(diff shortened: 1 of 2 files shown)


antonper:
This adds "nodes": [] to all the AST nodes. Even to those that should not have "nodes" tag.

mattsse:
hmm, we're lacking context when deserializing here, did this change break something?

we could make this work with some additional effort, by checking if the object had a `nodes` key

antonper:
As another way, making `pub nodes: Option<Vec<Node>>,` and adding `skip_serializing_if = "Option::is_none"`

PR #1818, opened 2022-10-31
fix: get_logs_paginated fetches past latest block

rogerso:
## Motivation

`provider.get_logs_paginated` would try to fetch more blocks than are available, because it only considers `page_size` when fetching blocks.

## Solution

Only fetch up to `from_block + page_size` or the last block available, whichever is lower.

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [x] Updated the changelog


mattsse (approved):
sounds right to me, nice catch!

gakonst (approved):
good catch! thanks

PR #1823, opened 2022-11-02
Fixes: #1822

### LINKED ISSUE #1822: `utils::parse_units` only supports numbers up to 2^96
### Version
1.0.0

### Description

The function `ethers::utils::parse_units` supports only parsing of numbers that are less than  2^96 as it is bounded by requirements of `rust_dec::Decimal`.

#### I tried this code:

```rust
use ethers::utils::parse_units;

let e15 = parse_units("1", 15).unwrap();
let e30 = e15 * e15;
assert_eq!(parse_units("1", 30).unwrap(), e30); // fails
```

#### I expected to see this happen: 
Does not revert

#### Instead, this happened: 
Reverts with `DecimalError(ExceedsMaximumPossibleValue)`

halo3mic:
## Motivation

Fixes #1822.

## Solution

For function `parse_units` replace usage of `rust_dec::Decimal` with alternative.

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog

### DIFF
--- a/ethers-core/src/utils/mod.rs
+++ b/ethers-core/src/utils/mod.rs
@@ -56,8 +56,8 @@ pub enum ConversionError {
     InvalidFloat(#[from] std::num::ParseFloatError),
     #[error(transparent)]
     FromDecStrError(#[from] FromDecStrErr),
-    #[error(transparent)]
-    DecimalError(#[from] rust_decimal::Error),
+    #[error("Overflow parsing string")]
+    ParseOverflow,
 }
 
 /// 1 Ether = 1e18 Wei == 0x0de0b6b3a7640000 Wei
@@ -173,7 +173,7 @@ where
         let mut a_uint = U256::from_dec_str(&amount_str)?;
         a_uint *= U256::from(10)
             .checked_pow(U256::from(exponent - dec_len))
-            .ok_or(rust_decimal::Error::ExceedsMaximumPossibleValue)?;
+            .ok_or(ConversionError::ParseOverflow)?;
         Ok(a_uint)
     }
 }
--- a/ethers-core/src/utils/mod.rs
+++ b/ethers-core/src/utils/mod.rs
@@ -155,19 +155,28 @@ where
     S: ToString,
     K: TryInto<Units, Error = ConversionError> + Copy,
 {
-    use rust_decimal::{Decimal, MathematicalOps};
+    let exponent: u32 = units.try_into()?.as_num();
+    let mut a_str = amount.to_string().replace("_", "");
+    let dot_i = if let Some(di) = a_str.find('.') {
+        a_str.remove(di);
+        di
+    } else {
+        a_str.len()
+    };
+    let a_len = a_str.len();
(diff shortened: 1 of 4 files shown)


ethers-core/src/utils/mod.rs
```diff
@@ -155,19 +155,28 @@ where
...
-    use rust_decimal::{Decimal, MathematicalOps};
-
-    let num: Decimal = amount.to_string().parse()?;
-    let exponent = units.try_into()?.as_num();
-
-    let multiplier = Decimal::TEN
-        .checked_powu(exponent.into())
-        .ok_or(rust_decimal::Error::ExceedsMaximumPossibleValue)?;
+    let exponent: u32 = units.try_into()?.as_num();
+    let mut a_str = amount.to_string().replace("_", "");
```
mattsse, line 159: I'm sensing, this is for `amount`?

let's rename the variable accordingly

ethers-core/src/utils/mod.rs
```diff
@@ -155,19 +155,28 @@ where
...
+    let dot_i = if let Some(di) = a_str.find('.') {
+        a_str.remove(di);
+        di
+    } else {
+        a_str.len()
+    };
+    let a_len = a_str.len();
+    let dec_len = (a_len - dot_i) as u32;
```
mattsse, line 167: can be flipped `a_str[di..].len()` else `0` ?

ethers-core/src/utils/mod.rs
```diff
@@ -155,19 +155,28 @@ where
...
 
-    let val =
-        num.checked_mul(multiplier).ok_or(rust_decimal::Error::ExceedsMaximumPossibleValue)?;
-    let u256_n: U256 = U256::from_dec_str(&val.round().to_string())?;
-    Ok(u256_n)
+    if dec_len > exponent {
+        let a_str = &a_str[..(a_len - (dec_len - exponent) as usize)];
```
mattsse, line 170: what's the rational behind this?

let's document this

mattsse (changes requested):
thanks for picking this up.

looks like we can get rid of rust_decimal then as well?

ethers-core/src/utils/mod.rs
halo3mic: Does "Truncate the decimal part if it is longer than the exponent" describe it well enough?

halo3mic:
> thanks for picking this up.
> 
> looks like we can get rid of rust_decimal then as well?

I left it for two reasons. The first is that is backwards compatible with the old error when the number was too large. Second is that the function must return enum`ConversionError`, and this error seems the most appropriate out of the options. 

Please let me know if I am mistaken, and I will change it.

mattsse:
that be great! yes let's do that

halo3mic:
@mattsse Implemented your suggestion with the latest two commits :)

mattsse (approved):
lgtm

gakonst (approved):
love it

PR #1825, opened 2022-11-03
Convert Node to JSON using serde.

Ethnical:
## Motivation

I am currently trying to create a static analyzer and would like to work directly on tree/ast to avoid compiling using solc compiler. 
To help me do that, I created a PR (https://github.com/hyperledger/solang/pull/1055) on `solang-parser` repo to implement serde `Serialization/Deserialization` using the feature  `features = ["pt-serde"]`. 
Thanks to this PR, we can easily Serialize the `Node` type : 
```rust
let x = json!(ethers_solc::resolver::Node::read(&args[1]).unwrap()); // x is now of type serde::Value.
println!("path (PathBuff) => {}", x["path"]);
    println!("Parsing tree without compiling (JSON) => {} ",serde_json::to_string_pretty(&x["data"]).unwrap());
```

So this can help us to work faster on JSON stuff. 
I am new in the Rust Ecosystem, so don't hesitate to tell me if there is something wrong  :) 

## Solution

So the solution to do that was to only add `Serialize, Deserialize` inside the derive over the type needed by `Node`

```rust
pub struct Node {
    /// path of the solidity  file
    path: PathBuf,
    /// content of the solidity file
    source: Source,
    /// parsed data
    data: SolData,
}
```
For example, SolData is now : 

```rust
#[derive(Debug, Deserialize, Serialize)]
#[allow(unused)]
pub struct SolData {
    pub license: Option<SolDataUnit<String>>,
    pub version: Option<SolDataUnit<String>>,
    pub experimental: Option<SolDataUnit<String>>,
    pub imports: Vec<SolDataUnit<SolImport>>,
    pub version_req: Option<VersionReq>,
    pub libraries: Vec<SolLibrary>,
    pub contracts: Vec<SolContract>,
}
```
Also needed to activate the feature inside  the `cargo.toml` of `ethers-solc`. 

```toml
solang-parser = { default-features = false, git = "https://github.com/hyperledger/solang.git", features = ["pt-serde"] }
```
Maybe adding them as a `feature` can save time for the compiling process.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog

### DIFF
--- a/ethers-contract/src/multicall/mod.rs
+++ b/ethers-contract/src/multicall/mod.rs
@@ -29,7 +29,7 @@ pub const MULTICALL_ADDRESS: Address = H160([
 
 /// The chain IDs that [`MULTICALL_ADDRESS`] has been deployed to.
 /// Taken from: https://github.com/mds1/multicall#multicall3-contract-addresses
-pub static MULTICALL_SUPPORTED_CHAIN_IDS: Lazy<[U256; 47]> = Lazy::new(|| {
+pub static MULTICALL_SUPPORTED_CHAIN_IDS: Lazy<[U256; 48]> = Lazy::new(|| {
     use Chain::*;
     [
         U256::from(Mainnet),                  // Mainnet


mattsse (changes requested):
no strong feelings about adding serde derives to the types.

But I think we should hold off on patching the solang-parser dependency, since unrelated.

PR #1826, opened 2022-11-03
Impl `IntoFuture` for `ContractCall`

TmLev:
## Motivation

Continuation of #1797.

## Solution

Implement `IntoFuture` trait for `ContractCall`.
Defaults to calling `ContractCall::call()`.

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog

### DIFF
--- a/ethers-contract/src/call.rs
+++ b/ethers-contract/src/call.rs
@@ -223,8 +223,9 @@ where
 /// Defaults to calling [`ContractCall::call`].
 impl<M, D> IntoFuture for ContractCall<M, D>
 where
-    M: Middleware + 'static,
-    D: Detokenize + 'static,
+    Self: 'static,
+    M: Middleware,
+    D: Detokenize,
 {
     type Output = Result<D, ContractError<M>>;
     type IntoFuture = Pin<Box<dyn Future<Output = Self::Output>>>;
--- a/ethers-contract/src/call.rs
+++ b/ethers-contract/src/call.rs
@@ -14,7 +14,14 @@ use ethers_providers::{
     Middleware, PendingTransaction, ProviderError,
 };
 
-use std::{borrow::Cow, fmt::Debug, future::Future, marker::PhantomData, sync::Arc};
+use std::{
+    borrow::Cow,
+    fmt::Debug,
+    future::{Future, IntoFuture},
+    marker::PhantomData,
+    pin::Pin,
+    sync::Arc,
+};
 
 use thiserror::Error as ThisError;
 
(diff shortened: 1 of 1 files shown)


TmLev:
I'm not fond of a needless allocation, but it helps to hide opaque future type.

ethers-contract/src/call.rs
```diff
@@ -211,3 +218,18 @@ where
...
+
+/// [`ContractCall`] can be turned into [`Future`] automatically with `.await`.
+/// Defaults to calling [`ContractCall::call`].
+impl<M, D> IntoFuture for ContractCall<M, D>
+where
+    M: Middleware + 'static,
+    D: Detokenize + 'static,
+{
+    type Output = Result<D, ContractError<M>>;
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output>>>;
```
TmLev, line 230: I'm not fond of a needless allocation, but it helps to hide opaque future type.

ethers-contract/src/call.rs
```diff
@@ -211,3 +218,18 @@ where
             .map_err(ContractError::MiddlewareError)
     }
 }
+
+/// [`ContractCall`] can be turned into [`Future`] automatically with `.await`.
+/// Defaults to calling [`ContractCall::call`].
+impl<M, D> IntoFuture for ContractCall<M, D>
+where
+    M: Middleware + 'static,
+    D: Detokenize + 'static,
```
TmLev, line 227: For some reason, compiler refuses to accept code without `'static` lifetimes.

ethers-contract/src/call.rs
```diff
@@ -211,3 +218,18 @@ where
...
+{
+    type Output = Result<D, ContractError<M>>;
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output>>>;
+
+    fn into_future(self) -> Self::IntoFuture {
+        Box::pin(async move { self.call().await })
+    }
```
TmLev, line 234: I'm not fond of excessive allocations, but it helps to hide an opaque future type.

TmLev:
сс @prestwich @gakonst

ethers-contract/src/call.rs
prestwich: The reason is that the future created is `'static` so it needs to know that neither the `Middleware` nor the `Detokenize` will have a shorter lifetime than `'static`. The best way to express this is probably as follows:

```
impl<M, D> IntoFuture for ContractCall<M, D>
where
    Self: 'static,
    M: Middleware,
    D: Detokenize,
```

TmLev: Moved to `Self`.

ethers-contract/src/call.rs
prestwich: GATs may help avoid this one in the future, but because this is a static lifetime future, it may be impossible to avoid

prestwich (approved):

PR #1832, opened 2022-11-04
enhance signer middleware to automatically switch legacy

kristian1108:
## Motivation

If you try to use SignerMiddleware on chains that don't support EIP-1559 (Optimism, BSC, etc), you'll get an error. Client code has to check the chain id everywhere, which is fragile and annoying. The middleware should automatically handle this, especially since the chain information is already stored in ethers-rs.

## Solution

Parse Chain from chain_id. Check if it's a "legacy only" chain. If so, convert the transaction to a legacy transaction as part of the fill logic.

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [] Updated the changelog  --> not sure what to do here?

### DIFF
--- a/ethers-middleware/src/signer.rs
+++ b/ethers-middleware/src/signer.rs
@@ -1,9 +1,10 @@
 use ethers_core::types::{
     transaction::{eip2718::TypedTransaction, eip2930::AccessListWithGasUsed},
-    Address, BlockId, Bytes, Signature, U256,
+    Address, BlockId, Bytes, Chain, Signature, TransactionRequest, U256,
 };
 use ethers_providers::{maybe, FromErr, Middleware, PendingTransaction};
 use ethers_signers::Signer;
+use std::convert::TryFrom;
 
 use async_trait::async_trait;
 use thiserror::Error;
@@ -251,6 +252,18 @@ where
             tx.set_chain_id(chain_id);
         }
 
+        // If a chain_id is matched to a known chain that doesn't support EIP-1559, automatically
+        // change transaction to be Legacy type.
+        if let Some(chain_id) = tx.chain_id() {
+            let chain = Chain::try_from(chain_id.as_u64());
+            if chain.unwrap_or_default().is_legacy() {
+                if let TypedTransaction::Eip1559(inner) = tx {
+                    let tx_req: TransactionRequest = inner.clone().into();
+                    *tx = TypedTransaction::Legacy(tx_req);
+                }
+            }
+        }
+
         let nonce = maybe(tx.nonce().cloned(), self.get_transaction_count(from, block)).await?;
         tx.set_nonce(nonce);
         self.inner()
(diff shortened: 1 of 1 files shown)


ethers-middleware/src/signer.rs
```diff
@@ -251,6 +252,24 @@ where
             tx.set_chain_id(chain_id);
         }
 
+        // If a chain_id is matched to a known chain that doesn't support EIP-1559, automatically
+        // change transaction to be Legacy type.
+        if tx.chain_id().is_some() {
+            let tx_chain_id = tx.chain_id().unwrap(); // safe unwrap after is_some check above
```
mattsse, line 258: use if let

ethers-middleware/src/signer.rs
```diff
@@ -251,6 +252,24 @@ where
...
+            let chain = Chain::try_from(tx_chain_id.as_u64());
+            if chain.is_ok() && chain.unwrap().is_legacy() {
```
mattsse, line 260: can simplify with unwrap_or_default

ethers-middleware/src/signer.rs
```diff
@@ -251,6 +252,24 @@ where
...
+                // safe unwrap after is_ok check
+                let legacy_tx = match tx.clone() {
+                    TypedTransaction::Eip1559(inner) => {
+                        let i_tx: TransactionRequest = inner.into();
+                        TypedTransaction::Legacy(i_tx)
+                    }
+                    other => other,
+                };
+                *tx = legacy_tx;
```
mattsse, line 269: this clone is unnecessary if already legacy.
match via `ref` should work here and clone if eip1559

mattsse (changes requested):

ethers-middleware/src/signer.rs
```diff
@@ -251,6 +252,18 @@ where
             tx.set_chain_id(chain_id);
         }
 
+        // If a chain_id is matched to a known chain that doesn't support EIP-1559, automatically
+        // change transaction to be Legacy type.
+        if let Some(chain_id) = tx.chain_id() {
+            let chain = Chain::try_from(chain_id.as_u64());
+            if chain.unwrap_or_default().is_legacy() {
```
mattsse, line 259:
```suggestion
            if chain.map(Chain::is_legacy).unwrap_or_default() {
```

ethers-middleware/src/signer.rs
```diff
@@ -251,6 +252,18 @@ where
...
+                if let TypedTransaction::Eip1559(inner) = tx {
+                    let i_tx: TransactionRequest = inner.clone().into();
+                    *tx = TypedTransaction::Legacy(i_tx);
```
mattsse, line 262:
```suggestion
                    let tx_req: TransactionRequest = inner.clone().into();
                    *tx = TypedTransaction::Legacy(tx_req);
```

mattsse (approved):
ty, lgtm.

this is a reasonable additional check imo.

smol pedantic nits

gakonst (approved):
nice! thank you

PR #1834, opened 2022-11-04
Match msg.sender to tx.origin during multicall

34x4p08:
Allows reading contract data with restricted access


gakonst (commented):
I'm not sure about this. Do we always want to do this? I have no strong opiion but dont' want to break expectations about what the code is supposed to do. Mabye @prestwich has thoughts?

34x4p08:
Proposed solution may break the functionality on some chains [[source]](https://github.com/Destiner/ethcall/issues/146) 
It's better to add possibility to configure 'from'. 
I'm closing this PR.

PR #1838, opened 2022-11-06
feat: warnings as errors

jatokz:
I would like to submit a new feature that would allow to create a setting for compile warnings to be included as errors.

## Motivation
I wanted to work on this feature request here:
https://github.com/foundry-rs/foundry/issues/3476

Following the guidance from that thread I am starting with making the additions here and I would be interested to also make the feature in Foundry after this is resolved.

With this setting turned on, the compiler when checking for errors, will also look for warnings

## Solution

I have made the code changes that added a property to Project/ProjectBuilder and it goes all the way down to the compiler output, quite similar to how "ignored_error_codes" is done. I also added one test. I feel that I should mention I am a beginner to rust and I would greatly appreciate any feedback or guidance :smile: 

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog


ethers-solc/src/compile/project.rs
```diff
@@ -358,14 +358,16 @@ impl<'a, T: ArtifactOutput> ArtifactsState<'a, T> {
     fn write_cache(self) -> Result<ProjectCompileOutput<T>> {
         trace!("write cache");
         let ArtifactsState { output, cache, compiled_artifacts } = self;
-        let ignored_error_codes = cache.project().ignored_error_codes.clone();
-        let skip_write_to_disk = cache.project().no_artifacts || output.has_error();
+        let project = cache.project();
+        let ignored_error_codes = project.ignored_error_codes.clone();
```
davidiola, line 362: perhaps follow convention here and change to:

```
let warnings_as_errors = cache.project().warnings_as_errors.clone();
```

davidiola (commented):
i had also been flirting with implementing this feature and happened to stumble across your PR so i put some of my thoughts in a review here 😄  

cheers to learning together!

ethers-solc/src/compile/output/mod.rs
```diff
@@ -426,8 +428,14 @@ impl AggregatedCompilerOutput {
     }
 
     /// Whether the output contains a compiler error
-    pub fn has_error(&self) -> bool {
-        self.errors.iter().any(|err| err.severity.is_error())
+    pub fn has_error(&self, warnings_as_errors: bool) -> bool {
+        self.errors.iter().any(|err| {
+            if warnings_as_errors {
+                true
```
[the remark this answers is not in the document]
gakonst: implementation pattern LGTM - defer to @mattsse on what the right choice to do here is

gakonst (approved):

ethers-solc/src/compile/output/mod.rs
```diff
@@ -426,8 +428,14 @@ impl AggregatedCompilerOutput {
     }
 
     /// Whether the output contains a compiler error
-    pub fn has_error(&self) -> bool {
-        self.errors.iter().any(|err| err.severity.is_error())
+    pub fn has_error(&self, warnings_as_errors: bool) -> bool {
```
mattsse, line 431: I think this is fine.

Instead of the bool argument this could just take the `Severity` enum as argument and find all errors that are `>=` (needs Ord impl)

mattsse (changes requested):
let's replace the bool arg with `Severity`

ethers-solc/src/compile/output/mod.rs
jatokz: I pushed a second commit that changed to using Severity as argument. I'm assuming it was following the correct pattern. If you think there are any specifics that could be adjusted, feel free to let me know.

During implement I had to choose between adding a Copy trait to Severity or passing a pointer down. I chose pointer since it was similar to "ignored_error_codes" and wasn't sure if Copy is something that is generally not preferred?

jatokz:
I added a second commit to go with Severity as an argument instead.

jatokz:
I took a look at the logs during the workflows lint step and noticed it exited when it got to my test.

I made another commit to see if formatting that code will fix it, but I am not really sure what went wrong.

ethers-solc/src/compile/output/mod.rs
```diff
@@ -449,7 +443,7 @@ impl AggregatedCompilerOutput {
         })
     }
 
-    pub fn diagnostics<'a>(&'a self, ignored_error_codes: &'a [u64], warnings_as_errors: bool) -> OutputDiagnostics {
+    pub fn diagnostics<'a>(&'a self, ignored_error_codes: &'a [u64], warnings_as_errors: &'a Severity) -> OutputDiagnostics {
```
mattsse, line 446: no need to take this as ref
```suggestion
    pub fn diagnostics<'a>(&'a self, ignored_error_codes: &'a [u64], warnings_as_errors: Severity) -> OutputDiagnostics {
```

ethers-solc/src/compile/output/mod.rs
```diff
@@ -711,7 +705,7 @@ pub struct OutputDiagnostics<'a> {
     /// the error codes to ignore
     ignored_error_codes: &'a [u64],
     /// treat warnings as errors
-    warnings_as_errors: bool,
+    warnings_as_errors: &'a Severity,
```
mattsse, line 708:
```suggestion
    warnings_as_errors: Severity,
```

mattsse, line 708: please rename to something akin to severity

mattsse (changes requested):

jatokz:
It was my pleasure :100:

mattsse (approved):
lgtm, ty!

ethers-solc/tests/project.rs
```diff
@@ -1618,6 +1618,25 @@ fn can_compile_model_checker_sample() {
     assert!(compiled.has_compiler_warnings());
 }
 
+#[test]
```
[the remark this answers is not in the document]
gakonst: Yes please!

gakonst (changes requested):
This is looking great, but let's add that one final test just to make sure we don't break anything on Foundry.

ethers-solc/tests/project.rs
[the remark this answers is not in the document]
jatokz: Great catch, I wrote a new test which failed as predicted by David. I adjusted the has_error() fn to be a conjunction of the two solutions. Ty for the guidance everyone :pray:

jatokz:
One thing to note, the first test checks all smart contracts in that test-data dir I created.

The second test I was much more specific and only tested 1 smart contract and the error code that goes with it.

mattsse (approved):

PR #1839, opened 2022-11-07
Examples event streams

0xMelkor:
## Motivation
As a library user I spent sometime to figure out how to manage event streams from `abigen` compiled contracts.
For such reason I added a couple of examples that can help new coming users.

## Solution
Two new examples were added:
1. Stream events from a compiled contract
2. Stream events and log meta from a compiled contract

## PR Checklist

-   [ ] Added Tests
- [x]  Added Documentation
-   [ ] Updated the changelog


mattsse (approved):
tysm!

PR #1844, opened 2022-11-08
fix: transaction type in TxpoolContent

### LINKED ISSUE #1843: TxpoolTransaction is missing some fields
**Version**

Current master.

**Description**

The RPC endpoint `txpool_content` returns transactions that look like this:

```json
{
    "accessList": [],
    "blockHash": null,
    "blockNumber": null,
    "chainId": "0x1",
    "from": "0xffd25e260c329253b8beb1ebcf6fe6eed215923d",
    "gas": "0x10bca",
    "gasPrice": "0x2363e7f00",
    "hash": "0x4be8632ab706e2a98a17e1cdc9cb200d3f6d6d80de0496b7241e6a04ee9d26e9",
    "input": "0x095ea7b3000000000000000000000000c36442b4a4522e871399cd717abdd847ab11fe88ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
    "maxFeePerGas": "0x2363e7f00",
    "maxPriorityFeePerGas": "0x5c7261c0",
    "nonce": "0x1",
    "r": "0xca87401a4cc0a354812a7150a2d4364b10d655108e62c9691b62f87f548e16c0",
    "s": "0x4368666a474a919e5454283b9a71e3e90cc37c816a52c8348c2e659c47c546e4",
    "to": "0x502a6dc08bf27a114deca570d561a12a0d8bfc30",
    "transactionIndex": null,
    "type": "0x2",
    "v": "0x1",
    "value": "0x0"
}
```

The type [TxpoolTransaction](https://github.com/gakonst/ethers-rs/blob/2418098d8c8a35f4674c57ea20d170ba3b50ae28/ethers-core/src/types/txpool.rs#L121-L133) only has the following fields:

```rust
pub struct TxpoolTransaction {
    pub block_hash: Option<H256>,
    pub block_number: Option<U64>,
    pub from: Option<Address>,
    pub gas: Option<U256>,
    pub gas_price: Option<U256>,
    pub hash: H256,
    pub input: Bytes,
    pub nonce: U256,
    pub to: Option<Address>,
    pub transaction_index: Option<U64>,
    pub value: U256,
}
```

That is, the following fields are missing:

- access list
- chain id
- max fee per gas
- max priority fee per gas
- type
- v, r, s

jannikluhn:
## Motivation

Fixes #1843 

## Solution

As suggested by @mattsee, this just replaces the `TxpoolTransaction` type with the normal `Transaction` type.

I updated the test with new data I gathered today. The test would have failed with the old data because it's missing some transaction fields that are now required.

I cheated a tiny bit: Geth encodes the addresses used as keys in the mappings (but not those used in the transaction as `from` and `to`) using the checksum format. We can deserialize this just fine, but when we serialize it again in the test to compare the result we encode them as lowercase. This would make our test fail. Therefore I opted to just manually make the addresses in the test data lowercase.

## PR Checklist

-   [x] ~Added~ Updated Test~s~
-   [ ] Added Documentation
-   [ ] Updated the changelog


mattsse (approved):
I'm a bit confused by this because that's not how it's documented here 
https://geth.ethereum.org/docs/rpc/ns-txpool

however quicknode docs for example, states that these are the regular transaction objects: https://www.quicknode.com/docs/ethereum/txpool_content

which makes more sense

mattsse:
> I'm guessing Anvil is using the other transaction type

that's it. do you want to migrate this as well @jannikluhn ?

mattsse:
awesome lmk if you're facing issues. it should simplify things a bit I think

PR #1847, opened 2022-11-09
fix: failing CI

DaniPopes:
## Motivation

should fix ~all~ most broken CI from last commits;
see https://github.com/gakonst/ethers-rs/pull/1844#issuecomment-1308126811 for failing txpool test;
~looks like [here](https://github.com/gakonst/ethers-rs/actions/runs/3425007628/jobs/5705345370) different tests sent transactions too close to eachother, but it doesn't happen always~ https://github.com/gakonst/ethers-rs/pull/1847#issuecomment-1308703933

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -174,3 +174,13 @@ required-features = ["yubi"]
 name = "paginated_logs"
 path = "examples/paginated_logs.rs"
 required-features = ["rustls"]
+
+[[example]]
+name = "subscribe_contract_events"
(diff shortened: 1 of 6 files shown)


mattsse (approved):
ty

DaniPopes:
> perhaps we can just add a sleep there?

Actually it might've just been because that action was running almost in parallel to a dependabot one, it should be fine

PR #1812, opened 2022-10-28
chore: clippy

DaniPopes:
## Motivation

new clippy nightly lints, same as foundry-rs/foundry#3546

## Solution

`cargo +nightly clippy --all --all-features --fix`

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog

PR #1851, opened 2022-11-10
fix: broken regex

DaniPopes:
## Motivation

https://github.com/gakonst/ethers-rs/actions/runs/3433299836/jobs/5723485668

#1812: 
image

Cannot replicate locally, I assume it was just a nightly clippy bug

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog


mattsse (approved):
I think the name value is actually wrong here, making this a noncapturing group makes sense.

PR #1852, opened 2022-11-10
fix: generated crate not using generated version

Will-Smith11:
## Motivation
currently when abigen generates a new crate. it uses the git version of the repo. However this is wrong as if someone set there ethers-rs dep to version 1.0.0, if they generated a crate for bindings. it would be the git version causing a dependency conflict 

## Solution
to fix this problem we simply read from the cargo manifest directory and then parse what version the bindings are being generated from and use that over the default hardcoded version

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/multi.rs
+++ b/ethers-contract/ethers-contract-abigen/src/multi.rs
@@ -596,9 +596,9 @@ impl MultiBindingsInner {
                         return Ok("ethers = {{ git = \"https://github.com/gakonst/ethers-rs\", default-features = false, features = [\"abigen\"] }}".to_string());
                     }
                 } else {
-                    let regex = Regex::new("ethers=\"[^\"]*\"")?;
+                    let regex = Regex::new("\"[^\"]*\"")?;
                     let Some(version) = regex.captures(&parsed) else { eyre::bail!("couldn't parse version regex")};
-                    let res = &version.get(0).unwrap().as_str()[7..];
+                    let res = &version.get(0).unwrap().as_str();
                     return Ok(format!("ethers = {{ version={}, default-features = false, features = [\"abigen\"] }}", res));
                 }
             }
--- a/ethers-contract/ethers-contract-abigen/src/multi.rs
+++ b/ethers-contract/ethers-contract-abigen/src/multi.rs
@@ -576,17 +576,17 @@ impl MultiBindingsInner {
         let file = File::open(cargo_dir)?;
         let reader = BufReader::new(file).lines();
         for line in reader.flatten() {
-            let parsed = line.trim();
+            let parsed: String = line.split_whitespace().collect();
             if parsed.starts_with("ethers") {
-                if parsed.contains("{{") {
+                if parsed.contains('{') {
                     if parsed.contains("git") && parsed.contains("rev") {
                         let regex = Regex::new("rev=\"[^\"]*\"")?;
-                        let Some(rev) = regex.captures(parsed) else { eyre::bail!("couldn't capture revision regex")};
+                        let Some(rev) = regex.captures(&parsed) else { eyre::bail!("couldn't capture revision regex")};
                         let res = rev.get(0).unwrap().as_str();
                         return Ok(format!("ethers = {{ git = \"https://github.com/gakonst/ethers-rs\", {}, default-features = false, features = [\"abigen\"] }}", res));
                     } else if parsed.contains("version") {
                         let regex = Regex::new("version=\"[^\"]*\"")?;
-                        let Some(version) = regex.captures(parsed) else { eyre::bail!("couldn't parse extra args version regex")};
+                        let Some(version) = regex.captures(&parsed) else { eyre::bail!("couldn't parse extra args version regex")};
                         let res = version.get(0).unwrap().as_str();
                         return Ok(format!(
                             "ethers = {{ {}, default-features = false, features = [\"abigen\"] }}",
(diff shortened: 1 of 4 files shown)


mattsse (changes requested):
> However this is wrong as if someone set there ethers-rs dep to version 1.0.0

right. good idea to improve that.

but instead of parsing this via regex, lets parse this as via toml instead, there's a toml parsing crate

Will-Smith11:
@mattsse is this more what you had in mind?

ethers-contract/ethers-contract-abigen/src/multi.rs
```diff
@@ -573,38 +572,24 @@ impl MultiBindingsInner {
...
-                } else {
-                    let regex = Regex::new("\"[^\"]*\"")?;
-                    let Some(version) = regex.captures(&parsed) else { eyre::bail!("couldn't parse version regex")};
-                    let res = version.get(0).unwrap().as_str();
-                    return Ok(format!("ethers = {{ version={}, default-features = false, features = [\"abigen\"] }}", res));
-                }
+        let data = std::fs::read_to_string(cargo_dir)?;
+        let toml = data.parse::<Value>()?;
+
+        let Some(Some(ethers)) = toml.get("dependencies").map(|v| v.get("ethers")) else { eyre::bail!("couldn't find ethers dep")};
```
mattsse, line 578: the can be simplified with `and_then` instead of map.

ethers-contract/ethers-contract-abigen/src/multi.rs
```diff
@@ -573,38 +572,24 @@ impl MultiBindingsInner {
...
+        if ethers.is_table() {
+            if let Some(rev) = ethers.get("rev") {
+                return Ok(format!("ethers = {{ git = \"https://github.com/gakonst/ethers-rs\", rev = {}, default-features = false, features = [\"abigen\"] }}", rev));
+            } else if let Some(version) = ethers.get("version") {
+                return Ok(format!("ethers = {{ version = {}, default-features = false, features = [\"abigen\"] }}", version ));
+            } else {
+                return Ok("ethers = {{ git = \"https://github.com/gakonst/ethers-rs\", default-features = false, features = [\"abigen\"] }}".to_string());
```
mattsse, line 585: I'm not quite sure this is correct, since version should also be allowed with `path`.

I think we can simplify all of this by checking for `path`?

mattsse (changes requested):

ethers-contract/ethers-contract-abigen/src/multi.rs
Will-Smith11: when you say path I'm assuming we are talking about something like this `ethers-core = { version = "^1.0.0", path = "../ethers-core", default-features = false }` ? I also see another case as if someone was just importing `ethers-contracts` this would also miss out on it.

mattsse: yes the path argument. 

> importing ethers-contracts this would also miss out on it.

right, let's check for both `ethers` and `ethers_contract` then?

Will-Smith11:
not super happy with how this currently looks, feels pretty clunky with managing path depth. Any ideas on cleanup?

ethers-contract/ethers-contract-abigen/src/multi.rs
```diff
@@ -570,22 +570,36 @@ impl MultiBindingsInner {
...
+                    res.extend(path);
+
+                    Ok(format!("ethers = {{ version = {}, path = {} default-features = false, features = [\"abigen\"] }}", version, res.to_string_lossy()))
+                } else {
+                    Ok(format!("ethers = {{ version = {}, default-features = false, features = [\"abigen\"] }}", version ))
+                }
             } else {
-                return Ok("ethers = {{ git = \"https://github.com/gakonst/ethers-rs\", default-features = false, features = [\"abigen\"] }}".to_string());
+                Ok("ethers = {{ git = \"https://github.com/gakonst/ethers-rs\", default-features = false, features = [\"abigen\"] }}".to_string())
             }
```
mattsse, line 600: I'm sorry I made a mistake in the last review, it should of course be `git` not `path`.

imo this can all be replaced by checking only for the existence of `git`. I wouldn't bother with the `path` at all. 

so the condition should just be `if ethers.get("git").is_some()`

mattsse (changes requested):

ethers-contract/ethers-contract-abigen/src/multi.rs
Will-Smith11: Right because abigen is a feature so will always be a table.

Will-Smith11: > so the condition should just be `if ethers.get("git").is_some()`
I don't believe we can just check for `git` though as another possible flag is `rev` which specifies a specific git hash. I think the best way to do this would be the way that the code I pushed currently handles it.

Will-Smith11: @mattsse thoughts on above?

mattsse: rev requires git tough: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

Will-Smith11: > rev requires git tough

yes I'm aware. I think its a good idea to allow specification of a specific hash due to dependency conflicts with other crates and I've actually had to use this a few times where the most recent version had a bug i wanted to avoid but the most recent git verision gave me a dep conflict. It's a pretty niche case but I think its more correct overall. Thinking o

If you think its a bit overkill I can remove it, just lmk.

gakonst: i think OK as-is

Issue #1820, opened 2022-11-02
retry policy backs off too aggressively when doing many parallel requests


ncitron:
**Version**
1.0.0

**Description**
When making many requests in parallel, the retry provider seems to ramp up the backoff too aggressively. This is likely because if you are performing many parallel requests, some of them are likely to fail many times in a row, and have the wait period continue to grow exponentially.

I was able to quickly hack together a fix by changing https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/transports/retry.rs#L297 to just return the initial backoff time rather than applying the exponential. This change made a parallel 250 requests to Alchemy go from taking 3 minutes to around 10 seconds.

At the very least, we probably don't want to be raising the retry length to the power of the number of failed attempts for that request. Maybe doing something like increasing by x% each successive failure would be more reasonable.

Happy to open a PR here, but want to agree on the best plan first.

mattsse:
this makes sense, considering that we already account for many concurrent requests here 

https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/transports/retry.rs#L316

so the exponential backoff is not actually practical here.

ncitron:
What do you think the best thing to do here is? I just removed the exponential and replaced it with just the initial backoff, but maybe its better that we completely cut this part and set it to 0, since it already adds the delay based on compute units later?

mattsse:
Right, I guess we can just remove the timeout since we already account for the current budget. but this only works for providers that measure via compute units per second.

I guess we should keep the timeout. We could set the timeout to something below 1s so the call is _likely_ rescheduled in the next 1s window. Maybe 0.75s is a reasonable value here that gives us a high probability that we queue in the next 1s window?

But I don't know what's alchemy's/infura's fidelity for the compute budget is.

budget estimation is also based on averages so with more expensive calls it would be more likely that 0.7s timeout is not quite enough.

But let's try that and I'll also monitor all of foundry's test keys once upstreamed and check if there are big changes, rn we currently Hit rate limits in 1% of all cases, which indicates that we can decrease the timeout aggressively.

So let's try

* rm exponential timeout
* decrease default timeout to ~<1s

wdyt?

PR #1855, opened 2022-11-12
fix: better retry timing

ncitron:
addresses #1820 

cc @mattsse


mattsse (approved):
lgtm, as discussed. will monitor once upstreamed in foundry

PR #1861, opened 2022-11-15
docs: updated logs event filtering for examples with new syntax

VanjaRo:
## Motivation

There is a separate function that does exactly same thing.

## Solution

Replaced:  
```rs
topic0(ValueOrArray::Value(H256::from(keccak256(...))))
```
with
```rs
event()
```


mattsse (approved):
lgtm

PR #1869, opened 2022-11-17
Fix: use cwd manifest

Will-Smith11:
## Motivation
the previous fix used CARGO_MANIFEST_DIR however this turns out to return the crate directory rather than the dependant crate directory making this impl wrong 

## Solution
use the cwd instead. This doesn't cover all cases but is better than the current which misses all the cases

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/multi.rs
+++ b/ethers-contract/ethers-contract-abigen/src/multi.rs
@@ -571,7 +571,7 @@ impl MultiBindingsInner {
 
     /// parses the active Cargo.toml to get what version of ethers we are using
     fn find_crate_version(&self) -> Result<String> {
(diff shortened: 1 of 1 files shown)


gakonst (changes requested):
can we please add some tests for this on the edge cases?

mattsse (approved):
this should be enough, can revise once edge cases come up

gakonst:
Let's just merge so we get this out, but let's follow up with a test PR

gakonst:
I think this. MAY have broken docs.rs builds https://docs.rs/crate/ethers/1.0.1/builds/679442?

mattsse:
hmm don't think so since this is shouldn't be used during macro expansion.

PR #1876, opened 2022-11-21
fix: find_contract and find not working

oneforalone:
## Motivation

`find` and `find_contract` method return None in `ethers_solc::output::ProjectCompileOutput`.

## Solution

`find_contract` and `find` method do not working with relative path, so i change `contract_path` to absolute path, and it works as the example showed.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog

### DIFF
--- a/ethers-solc/src/compile/output/mod.rs
+++ b/ethers-solc/src/compile/output/mod.rs
@@ -12,7 +12,7 @@ use crate::{
 };
 use contracts::{VersionedContract, VersionedContracts};
 use semver::Version;
-use std::{collections::BTreeMap, fmt, path::Path};
+use std::{collections::BTreeMap, env, fmt, path::Path};
 use tracing::trace;
 
 pub mod contracts;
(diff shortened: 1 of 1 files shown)


ethers-solc/src/compile/output/mod.rs
```diff
@@ -278,7 +278,9 @@ impl<T: ArtifactOutput> ProjectCompileOutput<T> {
     /// # }
     /// ```
     pub fn find(&self, path: impl AsRef<str>, contract: impl AsRef<str>) -> Option<&T::Artifact> {
-        let contract_path = path.as_ref();
+        let mut current_dir = env::current_dir().unwrap();
+        current_dir.push(path.as_ref());
+        let contract_path = current_dir.to_str().unwrap();
```
mattsse, line 283: I'm strongly against this here.
There's no guarantee that the cwd is the root dir.

what's an example where you'd need this?

mattsse (changes requested):
> find_contract and find method do not working with relative path

can you elaborate on this please?

ethers-solc/src/compile/output/mod.rs
oneforalone: I'm just following a tutorial to deploy a demo contract, but failed with getting the contract from the output after compiling. Is just annoying to setting the absolute path of the contract.

Okay, i got you point. So, it's proper to use `find_first` method?

mattsse: > Is just annoying to setting the absolute path of the contract.

right, there are functions to strip/join the root path for example.

`find_first` returns the first matching contract by name, if this is unique this is easier to use.

PR #1877, opened 2022-11-22
feat: retry client wasm support

ncitron:
Retry client builds under wasm32 now.

Curious what you think about the change to `maybe_connectivity`. `ReqwestErr` does not have `is_connect` in the wasm build, so we just don't check it. I don't think this should cause any major issues.

Also I added a new `TimerError` to `RetryClientError` which seemed a bit frivolous, but delays using `wasm_timer` can fail, so needed to cast it to some sort of error.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1160,9 +1160,9 @@ dependencies = [
 
 [[package]]
 name = "ethabi"
-version = "17.2.0"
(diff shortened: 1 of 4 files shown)


mattsse (approved):
ty, lgtm

gakonst (approved):
gg love it

PR #1880, opened 2022-11-22
feat(providers): add a subset of admin namespace

Rjected:
## Motivation

Geth includes an `admin` namespace which is especially useful for querying its p2p details and prompting the node to add or remove peers.

## Solution

 - Add `NodeInfo` for the `admin_nodeInfo` return type
 - Add `PeerInfo` for the `admin_peers` return type
 - Add the following `admin` namespace RPCs to `Middleware`, implementing them on `Provider<P>`:
   - `admin_nodeInfo`
   - `admin_peers`
   - `admin_addPeer`
   - `admin_addTrustedPeer`
   - `admin_removePeer`
   - `admin_removeTrustedPeer`

This also modifies the `Geth` and `GethInstance` structs to support custom chain ids and running without the `--dev` flag. `--dev`-specific configuration options, and options that are incompatible with `--dev` are kept mutually exclusive with the `PrivateNetOptions` enum. Unless a dev-incompatible option is set, `--dev` mode will remain the default. An option for disabling discovery is also added.

This also adds a simple test using the new `Geth` options to set a few custom fields that are visible in the `admin_nodeInfo` output.

Misc change:
 - expose `from_int_or_hex` as a general utility so it can be used for `admin_nodeInfo` response deserialization.

TODO:
 - [x] cover other RPCs - test for adding and removing peers, as well as listing peers.
   - the test added does the following:
     - connect two geth peers
     - ensure that the `admin_peers` endpoint deserializes correctly
     - the `id` obtained from the `admin_peers` RPC call matches the `id` from the `admin_nodeInfo` response for that peer
   - more tests should be added for admin endpoints, maybe good for a future PR or issue
   - another thing out of scope for this PR: `genesis.json` deserialization tests
 - [x] simple (de)serialization tests for `NodeInfo` and `PeerInfo`
 - [x] consider other `admin_` endpoints to add - `admin_peerEvents` maybe?
   - `admin_peerEvents` is of scope for this PR, will create an issue

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog

### DIFF
--- a/ethers-core/src/utils/geth.rs
+++ b/ethers-core/src/utils/geth.rs
@@ -5,7 +5,7 @@ use std::{
     fs::File,
     io::{BufRead, BufReader},
     path::PathBuf,
-    process::{Child, Command},
+    process::{Child, Command, Stdio},
     time::{Duration, Instant},
 };
 
@@ -101,6 +101,9 @@ impl GethInstance {
 impl Drop for GethInstance {
     fn drop(&mut self) {
         self.pid.kill().expect("could not kill geth");
+        // wait for the process to exit - we don't need to use the Result because the process could
+        // have already exited
+        _ = self.pid.wait();
     }
 }
 
@@ -312,12 +315,15 @@ impl Geth {
                 init_cmd.arg("--datadir").arg(data_dir);
             }
 
+            // set the stderr to null so we don't pollute the test output
+            init_cmd.stderr(Stdio::null());
+
             init_cmd.arg("init").arg(temp_genesis_path);
             init_cmd
                 .spawn()
                 .expect("failed to spawn geth init")
                 .wait()
-                .expect("failed to wait for geth init");
+                .expect("failed to wait for geth init to exit");
         }
 
         if let Some(ref data_dir) = self.data_dir {
@@ -359,10 +365,10 @@ impl Geth {
 
         let mut child = cmd.spawn().expect("couldnt start geth");
 
-        let stdout = child.stderr.expect("Unable to get stderr for geth child process");
+        let stderr = child.stderr.expect("Unable to get stderr for geth child process");
 
         let start = Instant::now();
-        let mut reader = BufReader::new(stdout);
+        let mut reader = BufReader::new(stderr);
 
         // we shouldn't need to wait for p2p to start if geth is in dev mode - p2p is disabled in
         // dev mode
--- a/ethers-core/src/utils/geth.rs
+++ b/ethers-core/src/utils/geth.rs
@@ -2,7 +2,7 @@ use super::{unused_port, Genesis};
 use std::{
     env::temp_dir,
     fs::File,
-    io::{BufRead, BufReader, Read},
+    io::{BufRead, BufReader},
     path::PathBuf,
     process::{Child, ChildStderr, Command},
     time::{Duration, Instant},
@@ -304,6 +304,9 @@ impl Geth {
             }
         }
 
+        // verbosity 5
+        cmd.arg("--verbosity").arg("5");
+
         if let Some(ref ipc) = self.ipc_path {
             cmd.arg("--ipcpath").arg(ipc);
         }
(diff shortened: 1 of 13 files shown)


Rjected:
Currently fails the `deserialize_node_info_post_merge` test, ready for review after fixing that bug

Rjected:
Tests pass and the bug is fixed, the bug was that `from_int_or_hex` wouldn't deserialize big numbers, luckily `serde_json::Number` can be used with the `arbitrary-precision` feature from `serde_json`

ethers-providers/src/provider.rs
```diff
@@ -2124,4 +2169,118 @@ mod tests {
...
+
+        // create enodes for each geth instance using each id and port
+        let first_enode = format!("{}@localhost:{}", first_prefix.first().unwrap(), first_port);
+
+        // add the second geth as a peer for the first
+        let res = second_peer.add_peer(first_enode).await.unwrap();
+        assert!(res);
+
+        // wait for the geth dial loop to start and connect the static nodes
+        sleep(Duration::from_secs(20)).await;
```
gakonst, line 2268: this takes that long? can we instead poll the node like we poll ganache-cli instaed of hardcoding a wait time? it's gonna screw up our reth integration tests otherwise imo

ethers-providers/src/admin.rs
```diff
@@ -0,0 +1,330 @@
...
+    #[serde(rename = "static")]
+    pub static_node: bool,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn deserialize_peer_info() {
```
gakonst, line 188: sweet

ethers-providers/Cargo.toml
```diff
@@ -35,6 +35,9 @@ futures-timer = { version = "3.0.2", default-features = false }
 futures-channel = { version = "0.3.16", default-features = false, optional = true }
 pin-project = { version = "1.0.11", default-features = false }
 
+# peer-related amin namespace
```
gakonst, line 38:
```suggestion
# peer-related admin namespace
```

ethers-core/src/utils/mod.rs
```diff
@@ -441,6 +447,34 @@ pub fn eip1559_default_estimator(base_fee_per_gas: U256, rewards: Vec<Vec<U256>>
     (max_fee_per_gas, max_priority_fee_per_gas)
 }
 
+/// Deserializes the input into a U256, accepting both 0x-prefixed hex and decimal strings with
+/// arbitrary precision, defined by serde_json's [`Number`](serde_json::Number).
+pub fn from_int_or_hex<'de, D>(deserializer: D) -> Result<U256, D::Error>
```
gakonst, line 452: should we have this in other desers too? no need to do now

ethers-core/src/utils/geth.rs
```diff
@@ -93,10 +154,54 @@ impl Geth {
...
+        self.chain_id = Some(chain_id.into());
+        self
+    }
+
+    /// Disable discovery for the geth instance.
+    ///
+    /// This will put the geth instance into non-dev mode, discarding any previously set dev-mode
+    /// options.
+    #[must_use]
+    pub fn disable_discovery(mut self) -> Self {
```
gakonst, line 197: nice. we should consider thinking of how can we use these options on geth to cause partitions intentionally and see how our failure recovery works

gakonst (changes requested):
nice work! nits

gakonst:
@Rjected lets try to merge by eod?

Rjected:
@gakonst sounds good, will prioritize

ethers-core/src/utils/mod.rs
Rjected: some desers already use this, but it would be worth checking other types of rpc responses that might have decimal values beyond `u64::MAX`

ethers-providers/src/provider.rs
Rjected: yeah, added a method `wait_to_add_peer` which polls geth's stderr

Rjected: another method would be polling the `admin_peers` endpoint, we might be able to accomplish this using `admin_peerEvents` in the future

ethers-providers/src/admin.rs
Rjected: :)

Rjected:
Mysterious connection refused error in an unrelated test hmmm

gakonst (approved):
amazing work

PR #1882, opened 2022-11-22
Instantiate an event builder without a contract instance

0xMelkor:
## Motivation
The `Event` builder system requires to instantiate a new `Contract` to create a typed event builder ([as implemented here](https://github.com/gakonst/ethers-rs/blob/master/ethers-contract/src/contract.rs#L205)).
```rust
  /// Returns an [`Event`](crate::builders::Event) builder for the provided event.
  pub fn event<D: EthEvent>(&self) -> Event<M, D> {
        self.event_with_filter(Filter::new().event(&D::abi_signature()))
  }
```
To stream the same event from many different contracts, developers tend to subscribe to a `Log` stream, needing to map the `Log` struct into a typed event struct.

[](https://github.com/gakonst/ethers-rs/blob/master/ethers-contract/src/contract.rs#L205)

```rust
    let filter = Filter::new()
                     .event("Transfer(address,address,uint256)")
                     .address(ValueOrArray::Array(vec![
                        "0x..".parse()?,
                        "0x..".parse()?,
                        "0x..".parse()?,
                      ]));

    let mut stream = client.subscribe_logs(&filter).await?;

    while let Some(log) = stream.next().await {
         // Developers need to map 'Log' into the 'Transfer' event struct 😭😭😭😭😭
    }
```
## Solution
The proposed solution allows to create an event builder from an event type. The returned builder can be combined with all the existing builder functions. 

I did the following:
1. Added the static `event_of_type` function to the `Contract` struct.
2. Added the `address` function to the `Event` struct in order to extend the filter with required addresses
3. Added unit tests
4. Added a new example

Usage is like this:
```rust
    abigen!(
        ERC20,
        r#"[
            event Transfer(address indexed from, address indexed to, uint256 value)
        ]"#,
    );

    let event = Contract::event_of_type::<Transfer>(&client)
                       .address(ValueOrArray::Array(vec![
                           "0x..".parse()?,
                           "0x..".parse()?,
                           "0x..".parse()?,
                       ]));

    let mut stream = event.subscribe().await?;

    while let Some(Ok(log)) = stream.next().await {
         // 'Log' has the 'Transfer' type. No mapping is required 🥳🥳🥳🥳🥳
    }
```

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-contract/src/event.rs
+++ b/ethers-contract/src/event.rs
@@ -3,7 +3,7 @@
 use crate::{log::LogMeta, stream::EventStream, ContractError, EthLogDecode};
 use ethers_core::{
     abi::{Detokenize, RawLog},
-    types::{BlockNumber, Filter, Log, Topic, H256},
+    types::{BlockNumber, Filter, Log, Topic, H256, ValueOrArray, H160},
 };
 use ethers_providers::{FilterWatcher, Middleware, PubsubClient, SubscriptionStream};
 use std::{borrow::Cow, marker::PhantomData};
@@ -106,6 +106,12 @@ impl<M, D: EthLogDecode> Event<'_, M, D> {
     /// Sets the filter's 3rd topic
     pub fn topic3<T: Into<Topic>>(mut self, topic: T) -> Self {
         self.filter.topics[3] = Some(topic.into());
+        self
+    }
+
+    /// Sets the filter's address.
+    pub fn address(mut self, address: ValueOrArray<H160>) -> Self {
+        self.filter.address = Some(address);
         self
     }
 }
--- a/ethers-contract/src/event.rs
+++ b/ethers-contract/src/event.rs
@@ -3,7 +3,7 @@
 use crate::{log::LogMeta, stream::EventStream, ContractError, EthLogDecode};
 use ethers_core::{
     abi::{Detokenize, RawLog},
-    types::{BlockNumber, Filter, Log, Topic, H256, ValueOrArray, H160},
+    types::{BlockNumber, Filter, Log, Topic, ValueOrArray, H160, H256},
 };
 use ethers_providers::{FilterWatcher, Middleware, PubsubClient, SubscriptionStream};
 use std::{borrow::Cow, marker::PhantomData};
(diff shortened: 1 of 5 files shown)


ethers-contract/src/event.rs
```diff
@@ -108,6 +108,12 @@ impl<M, D: EthLogDecode> Event<'_, M, D> {
         self.filter.topics[3] = Some(topic.into());
         self
     }
+
+    /// Sets the filter's address.
+    pub fn address(mut self, address: ValueOrArray<H160>) -> Self {
```
mattsse, line 113:
```suggestion
    pub fn address(mut self, address: ValueOrArray<Address>) -> Self {
```

mattsse (approved):
smol nit

ethers-contract/src/event.rs
0xMelkor: Change applied ✌️

ethers-contract/tests/it/contract.rs
```diff
@@ -324,6 +324,21 @@ mod eth_tests {
...
+            AggregatorInterface,
+            r#"[
+                event AnswerUpdated(int256 indexed current, uint256 indexed roundId, uint256 updatedAt)
+            ]"#,
+        );
+
+        let anvil = Anvil::new().spawn();
+        let client = connect(&anvil, 0);
+        let event = ethers_contract::Contract::event_of_type::<AnswerUpdatedFilter>(&client);
+        assert_eq!(event.filter, Filter::new().event(&AnswerUpdatedFilter::abi_signature()));
```
gakonst, line 339: brilliant

examples/subscribe_events_by_type.rs
```diff
@@ -0,0 +1,52 @@
...
+            PRICE_FEED_3.parse()?,
+        ]));
+
+    let mut stream = event.subscribe_with_meta().await?;
+
+    // Note that `log` has type AnswerUpdatedFilter
+    while let Some(Ok((log, meta))) = stream.next().await {
+        println!("{:?}", log);
+        println!("{:?}", meta)
+    }
```
gakonst, line 43: just beautiful

gakonst (approved):

PR #1890, opened 2022-11-24
Feat middleware stack builder

0xMelkor:
## Motivation
The ability to stack different `Middleware`s  is a powerful feature of this library, since it promotes "Single Responsibility", "Open Closed Principle" and other good stuff. 
I think that readability and composability can be further improved. For such reason I propose another approach to build stacked `Middleware`s using a builder struct.

## Solution
I provided
- A `ProviderBuilder` module under  the `ethers-middleware` lib
- Extensive documentation
- Unit tests

It works as follows 
```rust
fn example() {
    let key = "fdb33e2105f08abe41a8ee3b758726a31abdd57b7a443f470f23efce853af169";
    let signer = key.parse::<LocalWallet>().unwrap();
    let address = signer.address();
    let escalator = GeometricGasPrice::new(1.125, 60_u64, None::<u64>);
 
    let provider = Provider::<Http>::try_from("http://localhost:8545").unwrap();
 
    ProviderBuilder::from(provider)
        .wrap_into(|p| GasEscalatorMiddleware::new(p, escalator, Frequency::PerBlock))
        .wrap_into(|p| SignerMiddleware::new(p, signer))
        .wrap_into(|p| GasOracleMiddleware::new(p, EthGasStation::new(None)))
        .wrap_into(|p| NonceManagerMiddleware::new(p, address)) // Outermost layer
        .build();
}
```

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ethers-middleware/src/builder.rs
+++ b/ethers-middleware/src/builder.rs
@@ -1,8 +1,9 @@
 use ethers_providers::Middleware;
 
-/// A builder struct useful to compose different [`Middleware`](crate::Middleware) layers and then
-/// build a composed [`Provider`](crate::Provider) architecture. [`Middleware`](crate::Middleware)
-/// composition acts in a wrapping fashion. Adding a new layer results in wrapping its predecessor.
+/// A builder struct useful to compose different [`Middleware`](ethers_providers::Middleware) layers
+/// and then build a composed [`Provider`](ethers_providers::Provider) architecture.
+/// [`Middleware`](ethers_providers::Middleware) composition acts in a wrapping fashion. Adding a
+/// new layer results in wrapping its predecessor.
 ///
 /// Builder can be used as follows:
 ///
--- a/ethers-middleware/src/builder.rs
+++ b/ethers-middleware/src/builder.rs
@@ -1,7 +1,8 @@
+use ethers_core::types::Address;
 use ethers_providers::Middleware;
 use ethers_signers::Signer;
 
-use crate::SignerMiddleware;
+use crate::{SignerMiddleware, NonceManagerMiddleware};
 
 /// A builder struct useful to compose different [`Middleware`](ethers_providers::Middleware) layers
 /// and then build a composed [`Provider`](ethers_providers::Provider) architecture.
@@ -41,11 +42,23 @@ impl<M> ProviderBuilder<M>
 where
     M: Middleware,
 {
+
+    /// Wraps the current [`Middleware`](ethers_providers::Middleware) inside a [`SignerMiddleware`](crate::SignerMiddleware).
+    ///
+    /// [`Signer`] ethers_signers::Signer
     pub fn with_signer<S>(self, signer: S) -> ProviderBuilder<SignerMiddleware<M, S>> 
     where
         S: Signer
     {
         let provider = SignerMiddleware::new(self.inner, signer);
+        ProviderBuilder::from(provider)
+    }
+
+    /// Wraps the current [`Middleware`](ethers_providers::Middleware) inside a [`NonceManagerMiddleware`](crate::NonceManagerMiddleware).
+    ///
+    /// [`Address`] ethers_core::types::Address
+    pub fn nonce_manager(self, address: Address) -> ProviderBuilder<NonceManagerMiddleware<M>> {
+        let provider = NonceManagerMiddleware::new(self.inner, address);
         ProviderBuilder::from(provider)
     }
 
@@ -63,7 +76,7 @@ where
         ProviderBuilder::from(provider)
     }
 
-    /// Returns the overall[`Middleware`](ethers_providers::Middleware) as a reference to the
+    /// Returns the overall[`Middleware`](ethers_providers::Middleware) as an owned reference to the
     /// outermost layer
     pub fn build(self) -> M {
         self.inner
--- a/ethers-middleware/src/builder.rs
+++ b/ethers-middleware/src/builder.rs
@@ -27,7 +27,7 @@ use crate::{
 ///     let address = signer.address();
 ///     let escalator = GeometricGasPrice::new(1.125, 60_u64, None::<u64>);
 ///     let gas_oracle = EthGasStation::new(None);
-/// 
+///
 ///     let provider = Provider::<Http>::try_from("http://localhost:8545").unwrap();
 ///
 ///     let provider = ProviderBuilder::from(provider)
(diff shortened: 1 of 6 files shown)


gakonst (commented):
this is really neat, i like it. curious what @mattsse @prestwich think.

prestwich:
would this be better hardcoding common middleware?

````
ProviderBuilder::from(provider)
    .with_signer(signer)
    .nonce_manager(addr)
    .wrap_into(|p| SomeCustomMiddleware(p, other_arg))
    .build();

File: ethers-middleware/src/builder.rs [hunk 1]
```diff
@@ -0,0 +1,67 @@
...
+///     ProviderBuilder::from(provider)
+///         .wrap_into(|p| GasEscalatorMiddleware::new(p, escalator, Frequency::PerBlock))
+///         .wrap_into(|p| SignerMiddleware::new(p, signer))
+///         .wrap_into(|p| GasOracleMiddleware::new(p, EthGasStation::new(None)))
+///         .wrap_into(|p| NonceManagerMiddleware::new(p, address)) // Outermost layer
+///         .build();
+/// }
+/// ```
+pub struct ProviderBuilder<M> {
+    inner: Option<M>,
```
prestwich, line 34: why is this an option if all uses of it immediately `take()` and `unwrap()`?

prestwich:
should this be implemented as provided functions on the Middleware trait?

prestwich:
direction is very good. ergonomics I'm not so sure about? 🤔

ethers-middleware/src/builder.rs
0xMelkor: I found it handy to take the ownership on the inner Middleware and pass it through. Maybe restricting the M bound to Clone could work as well. Other suggestions?

0xMelkor:
Probably this could be part of the Middleware trait.
For what concerns ergonomics I can improve it. Do you have ideas?

ethers-middleware/src/builder.rs
prestwich: you can take ownership of the `inner` by changing `wrap_into` and build to consume `self` as follows. 

```
pub fn wrap_into<F, R>(self, builder_fn: F) -> ProviderBuilder<R>
    where
        F: FnOnce(M) -> R,
        R: Middleware,
    {
        let provider = self.inner;
        let provider = builder_fn(provider.unwrap());
        ProviderBuilder::from(provider)
    }

    /// Returns the overall[`Middleware`](ethers_providers::Middleware) as a reference to the
    /// outermost layer
    pub fn build(self) -> M {
        self.inner
    }
}```


Consuming is best-practice for a builder anyway, because it prevents misuse. Current design permits bad usage in sensible-looking code like this:

```
// normal looking code
let mut builder = ProviderBuilder::from(provider);

// A user who hasn't read the docs may not know that chained access is required
// this `take()` s from builder and returns a provider builder that the user doesn't bind
builder.wrap_into(|p| SomeMiddleware::new(p);

// this panics because the `wrap_into` took the `inner` and builder is now empty
builder.build()
```

**Comment by prestwich** [comment]
> Probably this could be part of the Middleware trait.

I think this would be really cool :)

**Comment by 0xMelkor** [comment]
> > Probably this could be part of the Middleware trait.
> 
> I think this would be really cool :)

The idea is interesting, but the `Middleware` trait (defined in `ethers-providers`) would need to reference the `ethers-middleware` module. The project structure doesn't allow this at the moment. 

We have at least 2 chances:
1. Update the workspace structure to link `ethers-providers` -> `ethers-middleware`. Probably not the best choice.
2. Keep the builder inside the `ethers-middleware` package

What do you think?

**Comment by prestwich** [comment]
> The idea is interesting, but the Middleware trait (defined in ethers-providers) would need to reference the ethers-middleware module. The project structure doesn't allow this at the moment.

well what I mean is that there's now no reason to have this be an independent struct at all. it could be provided methods on the `Middleware` trait itself 

@gakonst side note, if we're going to keep middleware pattern in 2.0, we should design a builder that prevents misuse like multiple signers, or combining known-incompatible middleware🤔

File: ethers-middleware/src/builder.rs [hunk 2]
```diff
@@ -0,0 +1,123 @@
+use ethers_core::types::Address;
+use ethers_providers::Middleware;
+use ethers_signers::Signer;
+
```
mattsse, line 4:
```suggestion
```

mattsse (commented):
this makes a ton of sense imo, supportive.

0xMelkor:
Ok guys,

I'll figure out a solution to add builder methods to the `Middleware`trait.
 
## About 2.0.0 - prevent builder misuse
I'd love it. We have 2 choices:
* Compile time check: 
    * Each middleware defines trait bounds on the downstream middlewares. This can avoid duplicates and incompatible compositions.
    * Concerns: Recursive trait bounds can be difficult to maintain. I need to check.
* Runtime check:
    * Builder checks constraints at runtime, raising panic if necessary.
    * Each middleware defines own composition rules, that are centrally validated by the builder on `build()`
    * Concerns: Allows wrong compositions to be shipped

0xMelkor:
Hi @prestwich

I investigated the extension of the `Middleware` trait. I'm afraid its current definition poses some limitations:
* The macro `#[auto_impl(&, Box, Arc)]` at [L149](https://github.com/0xMelkor/ethers-rs/blob/master/ethers-providers/src/lib.rs#L149) doesn't allow to consume  `self` in trait functions as per builder design.
* The same macro makes it hard to reason about the `FnOnce` arguments, which would instead work in a simpler trait case.
* Builder functions like `with_signer()` and `nonce_manager()` need to reference `ethers-middleware` definitions in their return types. But we are not allowed to reference them from `ethers-providers` since cargo complains about circular dependency errors.

What's your opinion?

prestwich:
given constraints, probably the best thing is an Ext trait in ethers_middleware? something like this

```
pub trait MiddlewareBuilder: ethers_providers::Middleware + Sized + 'static {
    fn wrap_into<M, F, T>(self, f: F) -> T
    where
        F: FnOnce(Self) -> T,
        T: ethers_providers::Middleware,
    {
        f(self)
    }

    fn with_signer<S>(self, s: S) -> SignerMiddleware<Self, S>
    where
        S: ethers_signers::Signer,
    {
        SignerMiddleware::new(self, s)
    }
}

impl<M> MiddlewareBuilder for M where M: ethers_providers::Middleware + Sized + 'static {}
```

gakonst (approved):
absolutely love this. killer work.

> The ability to stack different Middlewares is a powerful feature of this library, since it promotes "Single Responsibility", "Open Closed Principle" and other good stuff.

Would be keen to hear more of your thoughts here, as we can actually talk about this in our docs and explain the rationale.

0xMelkor:
> Would be keen to hear more of your thoughts here, as we can actually talk about this in our docs and explain the rationale.
Proud to take part in this.  I'll extend the docs to better convey the base concepts.

PR #1894, opened 2022-11-27
fix(core): format_units overflow

DaniPopes:
## Motivation

U256::from(u128...);
U256::exp10(x) with x > 77;
also lots of unnecessary calculations and casts

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/utils/mod.rs
+++ b/ethers-core/src/utils/mod.rs
@@ -159,15 +159,20 @@ where
     K: TryInto<Units, Error = ConversionError>,
 {
     let units: usize = units.try_into()?.into();
+    let amount = amount.into();
 
-    // 2**256 ~= 10**77
-    if units > 77 {
-        return Err(ConversionError::ParseOverflow)
-    }
+    // 78 overflows U256, 77 overflows I256
+    match amount {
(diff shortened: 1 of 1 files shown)


mattsse (commented):
nice catch, can we get a smol test for these edge cases?

ethers-core/src/utils/mod.rs
```diff
@@ -158,29 +158,32 @@ where
...
 {
-    let units = units.try_into()?;
-    match amount.into() {
+    let units: usize = units.try_into()?.into();
+    let amount = amount.into();
+
+    // 78 overflows U256, 77 overflows I256
+    match amount {
+        // 2**256 ~= 1.16e77
+        ParseUnits::U256(_) if units >= 78 => return Err(ConversionError::ParseOverflow),
```
0xMelkor, line 167: Despite the comments `77` and `78` sound like magic numbers. Should we assign constants with meaningful names?

gakonst: +1

gakonst (approved):
Love it - let's move the magic nums in consts and merge

Issue #1902, opened 2022-11-28
Fix: Change `VMOperation`'s `op` field type to ensure future proof


mouseless-eth:
**Version**
ethers 1.0.0

**Description**
The [pull request #1858](https://github.com/gakonst/ethers-rs/pull/1858) addressing [issue #1857](https://github.com/gakonst/ethers-rs/issues/1857) was implemented using a **unstable** design pattern that will cause issues in the future. A better approach would be to make the `op` field in [`VMOperation`](https://docs.rs/ethers/latest/ethers/core/types/struct.VMOperation.html) of type String instead of the newly introduced enum [`OpCode`](https://docs.rs/ethers/latest/ethers/types/enum.OpCode.html). 

**Making this change solves the following potential issues:** 
- Different clients labelling the same opcodes differently. (E.g. [Opcode 0x20 on Geth](https://github.com/ethereum/go-ethereum/blob/53d1ae096ac0515173e17f0f81a553e5f39027f7/core/vm/opcodes.go#L67) is called `KECCAK256` whilst opcode 0x20 on erigon use to be called `SHA3` until [very recently where 0x20 was renamed to KECCAK256](https://github.com/ledgerwatch/erigon/pull/5890)) 
> **Warning**
> Currently, [opcode 0x20 on enum `ethers::core::types::OpCode`](https://github.com/gakonst/ethers-rs/blob/ed47eaadad7d751cff9c0f59517b0fc3fb284cd3/ethers-core/src/types/trace/opcodes.rs#L69) is named SHA3 which causes any rpc call with `VmTrace` enabled on a geth node or updated erigon node to panic as it can't deserialize and map KECCAK256 to SHA3
    
- If opcodes are added or if existing opcodes are renamed (e.g. [eip-4758](https://eips.ethereum.org/EIPS/eip-4758)), [`ethers::core::types::OpCode`](https://github.com/gakonst/ethers-rs/blob/master/ethers-core/src/types/trace/opcodes.rs) will need to be updated to reflect this. Otherwise existing applications will panic whenever a `VmTrace` is requested as it won't be able to deserialize and parse the new opcodes from the trace.

PR #1904, opened 2022-11-28
fix: remove `OpCode` enum and update `VMOperation`'s `op` field type

mouseless-eth:
## Motivation

Addresses #1902

Ensures that tracing with VmTrace enabled will continue to be future proof in the presence of any opcodes added/renamed when ethereum undergoes a new hardfork.

## Solution

~~Removed the ethers::core::types::OpCode enum by deleting [it's source file](https://github.com/gakonst/ethers-rs/blob/master/ethers-core/src/types/trace/opcodes.rs#L70)~~

~~Changed VMOperation's op field to be of type string.~~

Create new enum `ExecutedInstruction` that deserializes opcodes from VmTrace into either `ExectutedInstruction::Known(Opcode)` or `ExecutedInstruction::Unknown(String)`

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog


mattsse (approved):
imo this is probably easier to maintain than an enum, so I'm supportive of this,

gakonst:
Should we instead do `enum Opcode { Known(Opcode), Unknown(String)`?

mouseless-eth:
> Should we instead do `enum Opcode { Known(Opcode), Unknown(String)`? I like the ability to match.

This route looks like a valid solution to unwanted serialization errors - I will work on implementing it, and we can test how it works.

ethers-core/src/types/trace/mod.rs
```diff
@@ -208,4 +225,19 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_deserialize_unknown_opcode() {
+        let example_opcodes = r#"["GAS", "CREATE2", "CUSTOMOP"]"#;
+        let parsed_opcodes: Vec<ExectuedInstruction> =
+            serde_json::from_str(example_opcodes).unwrap();
+        assert_eq!(
+            vec![
+                ExectuedInstruction::Known(Opcode::GAS),
```
gakonst, line 236:
```suggestion
                ExecutedInstruction::Known(Opcode::GAS),
```

gakonst (changes requested):
typo on the type, lgtm otherwise

mouseless-eth (commented):
Good catch, typo has been fixed

mouseless-eth (commented):
forgot to add files to prev comment

PR #1905, opened 2022-11-28
fix: txpool_inspect unable to parse contract creations

### LINKED ISSUE #1899: Error while calling txpool_inspect
**Version**
List the versions of all `ethers-rs` crates you are using. The easiest way to get
this information is using `cargo-tree`.
```
ethers = { version = "1.0", features = ["rustls", "abigen"]  }
```
**Platform**
The output of `uname -a` (UNIX), or version and 32 or 64-bit (Windows)
```
Linux <my_computer> #148-Ubuntu SMP Mon Oct 17 16:02:06 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```
**Description**
Enter your issue details here.
One way to structure the description:

I got error with `txpool_inspect` method. The RPC endpoint I am using is Quicknode.

This is code I was running:
```rust
let content = client.txpool_inspect().await?;
```
and I got the following error message.

```
Error: MiddlewareError(JsonRpcClientError(SerdeJson { err: Error("Invalid character 'n' at position 0", line: 1, column: 256), text: "{\"pending\":{\"0x00000000009A41862f3b2b0C688b7C0d1940511e\":{\"2964\":\"0x9C97e88452F9159BC7b5df9bcd546042940e4655: 0 wei + 30826 gas × 7000000000 wei\"},\"0x00000000863B56a3C1f0F1be8BC4F8b7BD78F57a\":{\"40\":\"contract creation: 0 wei + 612412 gas × 6000000000 wei\"},\"0x000000920f17A49d80550F044C9Bba7923A9fa36\":{\"3\":\"0x5403871b3ebA8B968381476bca2F10C34a5701A4: 50944227810000 wei + 21000 gas × 5358354608 wei\"},\"0x0000009a1363A910E457a2917c915dD4BA1297CB\":{\"34\":\"0x0000000000b3F879cb30FE243b4Dfee438691c04: 0 wei + 68329 gas × 1744367930 wei\"},\"0x000007cDf16e8173C035eC44E4D8a407d214F0f3\":{\"3\":\"0x5BA50c30A18848e33bBF5B71f26DdeD6Baf9F386: 2034918954089932 wei + 21000 gas × 5000000000 wei\"},\"0x00000e2B6fe65400B3a96c5b13E833B99622875d\":{\"5\":\"0xF74B1 ...
```

The error message sounds like there is a parsing error.

mouseless-eth:
## Motivation

Fixes #1899 

## Description of problem

The error occurs due to `TxpoolInspectSummary`'s deserialize method not having a way to parse the following response

```
"0x00000000863B56a3C1f0F1be8BC4F8b7BD78F57a": {
  "40": "contract creation: 0 wei + 612412 gas × 6000000000 wei"
}

```
## Solution
- Add a new match arm to map "contract creation" into None (as specified [here](https://github.com/gakonst/ethers-rs/blob/79e64a8378891df5afc6b83f75059c5afc587ae1/ethers-core/src/types/txpool.rs#L11)) when deserializing  `TxpoolInspectSummary`'s `to` field

- Rewrite tests to account for deserializing and serializing logs that include a contract creation

- `TxpoolInspectSummary`'s serialize method should serialize a `to` field equal to **None** into "contract creation" or "0x" instead of the default value for address type ("0x0000000000000000000000000000000000000000" - treated as a valid address)

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog

### DIFF
--- a/ethers-core/src/types/txpool.rs
+++ b/ethers-core/src/types/txpool.rs
@@ -65,9 +65,9 @@ impl<'de> Visitor<'de> for TxpoolInspectSummaryVisitor {
                 Some(Address::from_str(&addr.to_string().split_off(2)).map_err(de::Error::custom)?)
             }
         };
-        let value = U256::from(u64::from_str(value_split[0]).map_err(de::Error::custom)?);
-        let gas = U256::from(u64::from_str(gas_split[0]).map_err(de::Error::custom)?);
-        let gas_price = U256::from(u64::from_str(gas_price_split[0]).map_err(de::Error::custom)?);
+        let value = U256::from(u128::from_str(value_split[0]).map_err(de::Error::custom)?);
+        let gas = U256::from(u128::from_str(gas_split[0]).map_err(de::Error::custom)?);
+        let gas_price = U256::from(u128::from_str(gas_price_split[0]).map_err(de::Error::custom)?);
 
         Ok(TxpoolInspectSummary { to: addr, value, gas, gas_price })
     }
@@ -386,9 +386,9 @@ mod tests {
             "124930".to_string(),
             TxpoolInspectSummary {
                 to: Some(Address::from_str("000000000000000000000000000000000000007E").unwrap()),
-                value: U256::from(0u64),
-                gas: U256::from(100187u64),
-                gas_price: U256::from(20000000000u64),
+                value: U256::from(0u128),
+                gas: U256::from(100187u128),
+                gas_price: U256::from(20000000000u128),
             },
         );
         pending_map.insert(
@@ -400,36 +400,36 @@ mod tests {
             "252350".to_string(),
             TxpoolInspectSummary {
                 to: Some(Address::from_str("d10e3Be2bc8f959Bc8C41CF65F60dE721cF89ADF").unwrap()),
-                value: U256::from(0u64),
-                gas: U256::from(65792u64),
-                gas_price: U256::from(2000000000u64),
+                value: U256::from(0u128),
+                gas: U256::from(65792u128),
+                gas_price: U256::from(2000000000u128),
             },
         );
         pending_map_inner.insert(
             "252351".to_string(),
             TxpoolInspectSummary {
                 to: Some(Address::from_str("d10e3Be2bc8f959Bc8C41CF65F60dE721cF89ADF").unwrap()),
-                value: U256::from(0u64),
-                gas: U256::from(65792u64),
-                gas_price: U256::from(2000000000u64),
+                value: U256::from(0u128),
+                gas: U256::from(65792u128),
+                gas_price: U256::from(2000000000u128),
             },
         );
         pending_map_inner.insert(
             "252352".to_string(),
             TxpoolInspectSummary {
                 to: Some(Address::from_str("d10e3Be2bc8f959Bc8C41CF65F60dE721cF89ADF").unwrap()),
-                value: U256::from(0u64),
-                gas: U256::from(65780u64),
-                gas_price: U256::from(2000000000u64),
+                value: U256::from(0u128),
+                gas: U256::from(65780u128),
+                gas_price: U256::from(2000000000u128),
             },
         );
         pending_map_inner.insert(
             "252353".to_string(),
             TxpoolInspectSummary {
                 to: Some(Address::from_str("d10e3Be2bc8f959Bc8C41CF65F60dE721cF89ADF").unwrap()),
-                value: U256::from(0u64),
-                gas: U256::from(65780u64),
-                gas_price: U256::from(2000000000u64),
+                value: U256::from(0u128),
+                gas: U256::from(65780u128),
+                gas_price: U256::from(2000000000u128),
             },
         );
         pending_map.insert(
@@ -441,9 +441,9 @@ mod tests {
             "40".to_string(),
             TxpoolInspectSummary {
                 to: None,
-                value: U256::from(0u64),
-                gas: U256::from(612412),
-                gas_price: U256::from(6000000000u64),
+                value: U256::from(0u128),
+                gas: U256::from(612412u128),
+                gas_price: U256::from(6000000000u128),
             },
         );
         pending_map.insert(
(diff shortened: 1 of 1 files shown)


ethers-core/src/types/txpool.rs
```diff
@@ -89,12 +90,13 @@ impl Serialize for TxpoolInspectSummary {
     where
         S: serde::Serializer,
     {
+        let formatted_to = match self.to {
+            Some(to) => format!("{:?}", to),
+            None => "contract creation".to_string(),
+        };
```
mattsse, line 96: nit: use if let Some() else here

mattsse (commented):
interesting, do you know where geth has this implemented?

gakonst:
wonder if this is from a non-ethereum geth impl. i def tested it with real geth data when i wrote the feat

ethers-core/src/types/txpool.rs
mouseless-eth: Thank you for the tip, made the edit

mouseless-eth:
> interesting, do you know where geth has this implemented?

Seems to be implemented here at [go-ethereum/internal/ethapi/api.go](https://github.com/ethereum/go-ethereum/blob/8c5ce1107b3110c7cb735d8dfa91c9c701393c85/internal/ethapi/api.go#L235)

mouseless-eth:
> wonder if this is from a non-ethereum geth impl. i remember we had tested this with real geth data when the feature was implemented

Erigon does not have support for the `txpool_inspect` rpc (they chose to [comment it out](https://github.com/ledgerwatch/erigon/blob/76c36c2c0992ff2c1f410be03ddb3752434e410b/cmd/rpcdaemon/commands/txpool_api.go#L138) for some reason). Not sure about other node clients though

estarriolvetch:
Hey @mouseless-eth 

Thank you for the help!

I am trying to use your version to see if it works. However, I'm still getting the same result. Below is my code.

```main.rs
use std::io;
use std::io::prelude::*;
use std::ops::Mul;
use std::{thread, time::Duration};
use std::fs;
use std::{convert::TryFrom, path::Path, sync::Arc};
use ethers::{prelude::*};

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let provider = Provider::<Http>::try_from(
        // Test URL from quicknode docs
        "https://blue-light-gas.quiknode.pro/bc95fc4c322283e8b472d43143658e5673cbb631/",
    )?;

    let content = provider.txpool_inspect().await?;
    println!("{content:?}");
    Ok(())
}
```

```
[dependencies]
ethers = { git="https://github.com/mouseless-eth/ethers-rs", branch="fix/txpool_inspect" }
tokio = { version = "1.22", features = ["full"] }
```

estarriolvetch:
Actually, I'm getting a different error now.

```
Error: JsonRpcClientError(SerdeJson { err: Error("number too large to fit in target type", line: 1, column: 1550516), text: "{\"pe
```

estarriolvetch:
The full error is here:
https://gist.github.com/estarriolvetch/273d3188f7f0a61fbf2512740923746c

mouseless-eth:
> Actually, I'm getting a different error now.
> 
> ```
> Error: JsonRpcClientError(SerdeJson { err: Error("number too large to fit in target type", line: 1, column: 1550516), text: "{\"pe
> ```

Thanks for testing the branch and bringing the new error to attention

I isolated the error to this instance from the logs you shared
```
"0x307Ad9cbEB9ff143806cF163A79D26c9f13F8418": {
  "357": "0xab5558Fc3ac543433b15E70D84e6Ac11De74E512: 26000000000000000000 wei + 21100 gas × 10000000000 wei",
  "358": "0xab5558Fc3ac543433b15E70D84e6Ac11De74E512: 26000000000000000000 wei + 21100 gas × 9000000000 wei",
  "359": "0xab5558Fc3ac543433b15E70D84e6Ac11De74E512: 26000000000000000000 wei + 21100 gas × 9000000000 wei"
}
``` 

The error originates from the fact that the value sent (**26000000000000000000** wei) is greater than `u64::MAX` so when [`u64::from_str(value_split[0])`](https://github.com/gakonst/ethers-rs/blob/79e64a8378891df5afc6b83f75059c5afc587ae1/ethers-core/src/types/txpool.rs#L67) is called, it returns the error mentioned above

I bumped all u64 types to u128 for all variables that rely on `u64::from_str` to convert a string into a number (`value`, `gas`, `gas_price`), which should solve the problem (also bumped u64->u128 in tests just for consistency although makes no difference)

ethers-core/src/types/txpool.rs
```diff
@@ -60,13 +60,14 @@ impl<'de> Visitor<'de> for TxpoolInspectSummaryVisitor {
         let addr = match addr_split[0] {
             "" => None,
             "0x" => None,
+            "contract creation" => None,
             addr => {
                 Some(Address::from_str(&addr.to_string().split_off(2)).map_err(de::Error::custom)?)
```
DaniPopes, line 65:
```suggestion
                Some(Address::from_str(addr.trim_start_matches("0x")).map_err(de::Error::custom)?)
```

DaniPopes (changes requested):

ethers-core/src/types/txpool.rs
```diff
@@ -60,13 +60,14 @@ impl<'de> Visitor<'de> for TxpoolInspectSummaryVisitor {
...
             }
         };
-        let value = U256::from(u64::from_str(value_split[0]).map_err(de::Error::custom)?);
-        let gas = U256::from(u64::from_str(gas_split[0]).map_err(de::Error::custom)?);
-        let gas_price = U256::from(u64::from_str(gas_price_split[0]).map_err(de::Error::custom)?);
+        let value = U256::from(u128::from_str(value_split[0]).map_err(de::Error::custom)?);
+        let gas = U256::from(u128::from_str(gas_split[0]).map_err(de::Error::custom)?);
+        let gas_price = U256::from(u128::from_str(gas_price_split[0]).map_err(de::Error::custom)?);
```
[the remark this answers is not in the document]
mouseless-eth: Nice edits, much cleaner code.

You mention a good point however most other data types ([`TypedTransaction`](https://docs.rs/ethers/latest/ethers/types/transaction/eip2718/enum.TypedTransaction.html#method.gas), [`provider::get_gas_price`](https://docs.rs/ethers/latest/ethers/providers/struct.Provider.html#method.get_gas_price), [`ContractDeployer`](https://docs.rs/ethers/latest/ethers/contract/struct.ContractDeployer.html#method.gas), ...) use U256 for both `gas` and `gas_price` - might be better to keep U256 for consistency. But [`GethTrace`](https://docs.rs/ethers/latest/ethers/types/struct.GethTrace.html#structfield.gas) and [`StructLog`](https://docs.rs/ethers/latest/ethers/core/types/struct.StructLog.html) do use u64 for `gas`

gakonst (approved):
clean - good catch, thank you.

PR #1907, opened 2022-11-29
fix(contract): multicall decode error

DaniPopes:
## Motivation

If `Multicall` called an EOA or any method that didn't return anything but succeeded, it would fail to decode the empty bytes in `call_raw` and not parse it into tokens

## Solution

Return an empty Bytes token when the return data is empty

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog

### DIFF
--- a/ethers-contract/src/multicall/mod.rs
+++ b/ethers-contract/src/multicall/mod.rs
@@ -655,15 +655,19 @@ impl<M: Middleware> Multicall<M> {
                     .iter()
                     .zip(&return_data)
                     .map(|(call, bytes)| {
-                        let mut tokens: Vec<Token> = call
-                            .function
-                            .decode_output(bytes.as_ref())
-                            .map_err(ContractError::DecodingError)?;
-                        Ok(match tokens.len() {
-                            0 => Token::Tuple(vec![]),
-                            1 => tokens.remove(0),
-                            _ => Token::Tuple(tokens),
-                        })
+                        if bytes.len() == 0 {
+                            Ok(Token::Bytes(Default::default()))
+                        } else {
+                            let mut tokens = call
+                                .function
+                                .decode_output(bytes)
+                                .map_err(ContractError::DecodingError)?;
+                            Ok(match tokens.len() {
+                                0 => Token::Tuple(vec![]),
+                                1 => tokens.remove(0),
+                                _ => Token::Tuple(tokens),
+                            })
+                        }
                     })
                     .collect::<Result<Vec<Token>, M>>()?
             }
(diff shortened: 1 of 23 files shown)


ethers-contract/src/multicall/mod.rs
```diff
@@ -877,17 +861,21 @@ impl<M: Middleware> Multicall<M> {
...
-                contract_call = contract_call.block(block)
-            };
+            self.set_call_flags(contract_call).value(total_value)
+        }
+    }
 
-            if self.legacy {
-                contract_call = contract_call.legacy();
-            };
+    fn set_call_flags<D: Detokenize>(&self, mut call: ContractCall<M, D>) -> ContractCall<M, D> {
```
mattsse, line 870: please add some docs here

ethers-contract/src/multicall/mod.rs
```diff
@@ -674,14 +677,16 @@ impl<M: Middleware> Multicall<M> {
                 let return_data = call.call().await?;
                 self.calls
                     .iter()
-                    .zip(&return_data)
+                    .zip(return_data.into_iter())
                     .map(|(call, res)| {
-                        let ret = &res.return_data;
-                        let res_token: Token = if res.success {
+                        let bytes = &res.return_data;
+                        let res_token: Token = if bytes.len() == 0 {
```
mattsse, line 683:
```suggestion
                        let res_token: Token = if bytes.is_empty() {
```

ethers-contract/src/multicall/mod.rs
```diff
@@ -655,15 +654,19 @@ impl<M: Middleware> Multicall<M> {
...
-                        let mut tokens: Vec<Token> = call
-                            .function
-                            .decode_output(bytes.as_ref())
-                            .map_err(ContractError::DecodingError)?;
-                        Ok(match tokens.len() {
-                            0 => Token::Tuple(vec![]),
-                            1 => tokens.remove(0),
-                            _ => Token::Tuple(tokens),
-                        })
+                        if bytes.len() == 0 {
```
mattsse, line 657:
```suggestion
                        if bytes.is_empty() {
```

mattsse, line 657: please add a note here

ethers-contract/src/multicall/mod.rs
```diff
@@ -100,7 +98,7 @@ pub type Result<T, M> = std::result::Result<T, MulticallError<M>>;
 /// Helper struct for managing calls to be made to the `function` in smart contract `target`
 /// with `data`.
 #[derive(Clone, Debug)]
-pub struct Call {
+pub struct GenericCall {
```
mattsse, line 101: what's the reason behind renaming this?

can we not keep it as `Call`, because this is breaking, right?

mattsse (changes requested):

ethers-contract/src/multicall/mod.rs
DaniPopes: I think it's a bug with rust analyzer, there's multiple Call types even if they are aliased

DaniPopes: Found in the tg chat. I can rename it back if you want @mattsse

gakonst: let's avoid breaking unless we think that it's a basically unused api

gakonst (changes requested):

PR #1909, opened 2022-11-29
feat(core): Chain impls and refactoring

DaniPopes:
## Motivation

should not use u32 in conversions when Chain is a u64

## Solution

split into commits

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog


DaniPopes:
what do you think about using [num_enum](https://crates.io/crates/num_enum) for number conversions and strum for also the Display impl, as far as I can tell they'd do the same thing for those and then we don't have to update them manually @gakonst @mattsse

ethers-core/src/types/chain.rs
```diff
@@ -271,7 +119,7 @@ impl fmt::Display for Chain {
             Chain::AuroraTestnet => "aurora-testnet",
         };
 
-        write!(formatter, "{chain}")
+        f.pad(chain)
```
mattsse, line 122: this makes sure flags are respected

mattsse (approved):
lgtm

gakonst (approved):

PR #1910, opened 2022-11-29
feat(provider): check for serde error with missing req id

mattsse:
## Motivation
some providers send invalid JSON RPC in error case `id = null`.
this will handle the serde error and try to extract the JsonRpcError and whether to retry.

cc @mds1

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog


gakonst (approved):
ACK because we have encountered this enough times that it's an annoyance. RPC providers should do better though :)

Issue #1889, opened 2022-11-23
Add a timeout for WebSocket connections


timas130:
**Is your feature request related to a problem? Please describe.**
Some RPC services (like [QuickNode](https://www.quicknode.com)) completely drop WebSocket requests that go above the speed limit. They don't even respond with anything or terminate the connection - the request gets ignored. And when that happens, the library just waits for the response endlessly.

**Describe the solution you'd like**
An option to set the timeout for `Ws`.

**Describe alternatives you've considered**
Just wrapping the `Future` in a function like this:

```rust
async fn with_timeout<T: Future>(future: T) -> T::Output {
    tokio::select! {
        result = future => result,
        _ = tokio::time::sleep(std::time::Duration::from_secs(3)) => {
            panic!("timeout reached");
        }
    }
}
```

0xMelkor:
Hello,
In the past I had the same problem with Quicknode. I found out this working pattern

```rust
         // This loop iterates upon inactivity timeouts.
        'connection: loop {

             let client = Provider::<Ws>::connect("wss://your-uri").await?;
             let client = Arc::new(client);
             let filter = Filter::new().event("Transfer(address,address,uint256)");

             let mut stream = events.stream().await?;

            // We loop over stream.next(). If a timeout occurs we break the loop and connect again
            'listen: loop {
                let timeout = time::sleep(Duration::from_secs(30));
                tokio::pin!(timeout); // You need to pin the timeout future to use in tokio::select!

                tokio::select! {
                    filter = stream.next() => {
                      // Do some work
                    }
                    _ = &mut timeout => {
                        // Timeout reached
                        break 'listen;
                    }
                }
            }
        }
```

If you are interested there is a previous PR left open to cope with WS reconnection at the transport layer
[1675](https://github.com/gakonst/ethers-rs/pull/1675)

timas130:
@0xMelkor Thanks for the advice. That's basically what I'm using, just a bit more prettily written. :) Feel free to close the issue if  you're not going to fix this.

0xMelkor:
@timas130 I'm afraid I have not permissions to close this.
Can you share how you would rewrite this snippet? That's useful to me and maybe someone else in the future.
Thanks.

timas130:
@0xMelkor I mean that your snipped is easier to understand. The code I'm using is shown in the issue description.

Anyways, I think that even though it is possible to wrap the function like that, it should probably be included as a standard feature.

Issue #1815, opened 2022-10-28
WS Provider Panic


Tiergarten:
I am occasionally hitting the below panic in `ethers-prodivers/src/transports/ws.rs` impl `WsServer`

The error message I get is "upstream went away", I believe this is a network / reliability issue with my json-rpc provider however I would like to catch it and restart the listening thread - this panic is making that difficult.

Are there any recommendations on how to proceed?

```
    /// Spawns the event loop
    fn spawn(mut self)
    where
        S: 'static,
    {
        let f = async move {
            loop {
                if self.is_done() {
                    debug!("work complete");
                    break
                }
                match self.tick().await {
                    Err(ClientError::UnexpectedClose) => {
                        error!("{}", ClientError::UnexpectedClose);
                        break
                    }
                    Err(e) => {
                        panic!("WS Server panic: {}", e);
                    }
                    _ => {}
                }
            }
        };
```

Thanks,
Matt

cliff0412:
same issue here.
i am getting below error, but do not know how to capture and handle
2022-11-15T16:42:32.333490Z DEBUG tungstenite::protocol: Received close frame: Some(CloseFrame { code: Away, reason: "upstream went away" })    
2022-11-15T16:42:32.334704Z DEBUG tungstenite::protocol: Replying to close with Frame { header: FrameHeader { is_final: true, rsv1: false, rsv2: false, rsv3: false, opcode: Control(Close), mask: None }, payload: [3, 233, 117, 112, 115, 116, 114, 101, 97, 109, 32, 119, 101, 110, 116, 32, 97, 119, 97, 121] }

0xMelkor:
Hi guys,
The error is raised [here](https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/transports/ws.rs#L255) at transport layer. The library needs more robust handling of Websocket errors.

At the moment there is a pull request that is trying to fix this: [1675](https://github.com/gakonst/ethers-rs/pull/1675).

ibhagwan:
Getting a similar issue (for a different ws error):
```
2022-11-30 13:31:40.488 DEBG Received close frame: Some(CloseFrame { code: Away, reason: "CloudFlare WebSocket proxy restarting" })
[...]

PR #1915, opened 2022-12-01
Fix: handle panic on Ws error 

0xMelkor:
## Motivation
Some issues have been raised, reporting `WsServer` panics on upstream connection errors. 
#1889 #1815
https://github.com/gakonst/ethers-rs/blob/4b685621ed3a4b827276a13107f30457a19379db/ethers-providers/src/transports/ws.rs#L250-L256

## Solution
In order to have a simpler and more predictable behavior, I believe the application should directly handle reconnections (also stated [here](https://github.com/snapview/tokio-tungstenite/issues/101)), avoiding recovery mechanisms at the transport layer.

Instead of crashing the program, we gracefully close all active subscriptions. Clients will then observe the end of their streams and issue proper actions.

What do you guys think?

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,10 @@
 ## ethers-core
 
 ### Unreleased
+-   Graceful handling of WebSocket transport errors [#1889](https://github.com/gakonst/ethers-rs/issues/1889) [#1815](https://github.com/gakonst/ethers-rs/issues/1815)
 -   `MiddlewareBuilder` trait to instantiate a `Provider` as `Middleware` layers.
 -   An `Event` builder can be instantiated specifying the event filter type, without the need to instantiate a contract.
--   Add 'ethers_core::types::OpCode' and use in 'ethers_core::types::VMOperation' [1857](https://github.com/gakonst/ethers-rs/issues/1857)
+-   Add 'ethers_core::types::OpCode' and use in 'ethers_core::types::VMOperation' [#1857](https://github.com/gakonst/ethers-rs/issues/1857)
 -   Remove rust_decimals dependency for ethers-core
 -   Add support for numbers greater than 2^96 for `ethers_core::utils::parse_units` [#1822](https://github.com/gakonst/ethers-rs/issues/1822)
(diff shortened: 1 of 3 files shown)


mattsse (changes requested):
imo, it makes more sense to handle this gracefully and don't panic on the ws task.

needs changelog entry

0xMelkor:
Hi @mattsse, changelog updated.

gakonst (approved):
LGTM - let's iterate on this @mattsse so we unblock @0xMelkor

Tigranchick:
Sorry for the dumb question, but how can I catch this panic or error, i mean is there place to match it. I don't get messages from websocket often and it breaks with connection refused. Now I took one of the solutions with 2 loops and a timeout in tokio select! like there #1889 .

0xMelkor:
Hi @Tigranchick,

this PR removed the `panic!` statement in `ws.rs`, in favor of a graceful teardown of active subscriptions. The effect is that your stream  reaches the last element and you are up to handle reconnections.

```rust
    // This loop iterates over WebSocket disconnections
    'connection: loop {

        let client = Provider::<Ws>::connect("wss://your-uri").await?;
        let client = Arc::new(client);
        let filter = Filter::new().event("Transfer(address,address,uint256)");

        let mut stream = client.subscribe_logs(&filter).await?;

        while let Some(log) = stream.next().await {
                 // Do some work
        }
       
        // If you get here, your subscription has been canceled => Iterate over the main loop to handle reconnection
   }
```

### How can we improve ethers in my opinion
A cleaner approach would also propagate errors from Ws to the application. 
This requires some library changes along the stack, and the subscription code would become like:

```rust
        // The stream now emits `Result`instead of `Log`
        while let Some(result) = stream.next().await {
                 match result {
                     Ok(log) => // Do some work
                     Err(e) => // Handle the errror
                 }
        }
```

Tigranchick:
@0xMelkor, thanks you to explanation!

PR #1919, opened 2022-12-02
fix(core): always set p2p port in non-dev mode

Rjected:
## Motivation

If another process is already using the default p2p port, disabling discovery will cause the geth instance to crash because it will start with p2p enabled. Even if it doesn't crash, the p2p port will not be visible from the `GethInstance` created when geth is spawned. Instead, it would be `None`.

## Solution

Always set the p2p port if geth is in non-dev mode, using `unused_port` if the p2p port is not explicitly set.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog


mattsse (approved):
lgtm

PR #1933, opened 2022-12-08
fix(core): set `GethInstance` p2p_port in spawn

Rjected:
## Motivation

Previously, spawning Geth with disabled discovery would not actually set the p2p port in the returned GethInstance. This is because it would only pass the port to the geth command. If the value started as None, it would remain to be None in the GethInstance even though a port was passed to the command.

## Solution

Return the p2p port by the `match` statement that handles setting the p2p port if `Geth` is in `PrivateNetMode`. This port is then set and returned in the `GethInstance`.

 * adds tests for the existence of ports in GethInstance

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -666,7 +666,7 @@ dependencies = [
  "libc",
  "log",
  "matches",
- "nix",
+ "nix 0.13.1",
  "serde",
  "thiserror",
  "wasm-bindgen",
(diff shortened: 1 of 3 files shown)


Rjected:
the failing test has to do with a colliding datadir:
```
Fatal: Failed to create the protocol stack: datadir already used by another process
test utils::geth::tests::explicit_p2p_port ... ok
```

gakonst (approved):

PR #1934, opened 2022-12-09
fix(contract): allow 16 calls in multicall

onsails:
## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [x] Updated the changelog

### DIFF
--- a/ethers-contract/src/multicall/mod.rs
+++ b/ethers-contract/src/multicall/mod.rs
@@ -608,7 +608,7 @@ impl<M: Middleware> Multicall<M> {
     /// # }
     /// ```
     pub async fn call<D: Detokenize>(&self) -> Result<D, M> {
-        assert!(self.calls.len() < 16, "Cannot decode more than 16 calls");
+        assert!(self.calls.len() <= 16, "Cannot decode more than 16 calls");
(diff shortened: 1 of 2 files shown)


ethers-contract/src/multicall/mod.rs
```diff
@@ -608,7 +608,7 @@ impl<M: Middleware> Multicall<M> {
     /// # }
     /// ```
     pub async fn call<D: Detokenize>(&self) -> Result<D, M> {
-        assert!(self.calls.len() < 16, "Cannot decode more than 16 calls");
+        assert!(self.calls.len() <= 16, "Cannot decode more than 16 calls");
```
gakonst, line 611: OK this should be fine due to this line https://github.com/gakonst/ethers-rs/blob/master/ethers-core/src/abi/tokens.rs#L130

gakonst (approved):

ethers-contract/src/multicall/mod.rs
onsails: I have propotypes of both, will prepare another PR

onsails: https://github.com/gakonst/ethers-rs/pull/1941

PR #1940, opened 2022-12-15
refactor: examples

0xMelkor:
## Motivation
The examples directory is the first place for developers to onboard quickly on ethers-rs. Anyway the lack of a proper structure makes it difficult to find useful content. 

## Solution
I catched your desire to provide something similar to this [ebook repo](https://github.com/miguelmota/ethereum-development-with-go-book) and started restructuring the entire folder providing:

1. ToC README in the root dir
2. Refactor examples by topic 
3. Provide new examples where missing
4. In another PR create a CI step to export docs in convenient formats

I'll leave this in DRAFT till completion of the first 3 points.
@gakonst @mattsse @prestwich let me know if the ToC is missing some core topics.

Thank you guys

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


gakonst:
Absolutely love this. Strong yes, and we can use the feature checklist as a roadmap.

examples/big-numbers/README.md
```diff
@@ -0,0 +1,260 @@
+# Big numbers
+Ethereum uses big numbers (also known as "bignums" or "arbitrary-precision integers") to represent certain values in its codebase and in blockchain transactions. This is necessary because Ethereum uses a 256-bit numbering system, which is much larger than the number systems typically used in modern computers (such as base-10 or base-2). Using big numbers allows Ethereum to handle very large numbers, such as those that may be required to represent very large balances or quantities of a particular asset.
```
DaniPopes, line 2: > Ethereum uses a 256-bit numbering system, which is much larger than the number systems typically used in modern computers (such as base-10 or base-2).

Ethereum ([the EVM](https://ethereum.org/en/developers/docs/evm/)) operates on a *256-bit word size*, which is different from the usual 32-bit or 64-bit of modern machines (this has nothing to do with base-2 or base-10 which is how we choose to represent numbers). This was chosen for the ease of use with 256-bit cryptography (such as Keccak-256 hashes or secp256k1 signatures). (taken from [here](https://ethereum.org/en/developers/docs/evm/#evm-instructions))

examples/big-numbers/README.md
```diff
@@ -0,0 +1,260 @@
...
+### `format_units`
+```rust
+use ethers::{types::U256, utils::format_units};
+
+fn format_units_example() {
+    // 1 ETHER = 10^18 WEI
+    let one_ether = U256::from(1000000000000000000_u128);
+
+    let num: String = format_units(one_ether, "wei").unwrap();
+    assert_eq!(&*num, "1000000000000000000.0");
```
DaniPopes, line 224: I don't think the `&*` are necessary when doing `assert_eq!` for `String` with `str`, but I may be wrong; in that case the idiomatic way is to use `as_str`.

DaniPopes (changes requested):
This is amazing, thanks for taking this on!

Some early notes:
- I think it might make more sense if the READMEs contained only a few or no examples, leaving the job to the individual files to avoid having to maintain the same code twice and risking having the examples in the README be out of date.
- If a type implements a trait in `std::{cmp, ops}` (which is the case for `U256`), you can write the expression using the regular operator and you also won't need to bring the trait into scope (this is true only for the traits that are not already in scope from the [prelude](https://doc.rust-lang.org/std/prelude/index.html)). This is explained in more detail in the [`std::ops` docs](https://doc.rust-lang.org/stable/std/ops/). Example:
```diff
- use std::ops::Add;
let a = U256::from(100);
let b = U256::from(100);
- assert!(a.eq(&b));
- assert_eq!(a.add(&b), U256::from(200));
+ assert_eq!(a, b);
+ assert_eq!(a + b, U256::from(200));
```

examples/big-numbers/README.md
0xMelkor: Thank you for the hint. Now it's much cleaner.

Cargo.toml
```diff
@@ -184,3 +184,71 @@ required-features = ["rustls", "ws"]
 name = "subscribe_contract_events_with_meta"
 path = "examples/subscribe_contract_events_with_meta.rs"
 required-features = ["rustls", "ws"]
+
+[[example]]
```
gakonst, line 188: Can we avoid these explicit declarations somehow?

gakonst (commented):
gg this is a great start excited for when it lands

Cargo.toml
0xMelkor: Hi @gakonst, thanks for pointing this out. 

Having a cargo workspace makes it more complicated, since nested example files are not resolved automatically.

We could provide a `Cargo.toml` in the examples directory (also done by the [tokio](https://github.com/tokio-rs/tokio/tree/master/examples) team). In this case `examples` would become a new workspace member. That would offload the root `Cargo.toml`, at the cost of maintaining a new workspace member.

Can it work for you?  Any other advice is welcome.

Thank you

gakonst: Yep - I like doing that that way.

0xMelkor: Since examples are structured in folders, we can create a standalone crate for each. We then use the native cargo examples layout:

```bash
examples
├── big-numbers
│   ├── Cargo.toml
│   ├── README.md
│   └── examples
│       ├── bn_comparison_equivalence.rs
│       ├── bn_conversion.rs
│       ├── bn_create_instances.rs
│       ├── bn_math_operations.rs
│       └── bn_utilities.rs
├── contracts
│   ├── Cargo.toml
│   ├── README.md
│   └── examples
│       ├── IERC20.json
│       ├── contract_abigen.rs
│       ├── contract_events.rs
│       ├── contract_instances.rs
│       └── contract_methods.rs
```

We can run examples with this
`cargo run -p <example-crate-name> --example <example-name>`
i.e.
`cargo run -p examples-contracts --example contract_abigen`

Can it work for you?

0xMelkor: I restructured all examples with this approach:
* All files have been moved to their espective crate
* Examples can be run with `cargo run -p <example-crate-name> --example <example-name>`
* Build succeeds (locally)
* CI is broken, need to fix it 😆 

Can you guys take a look? @gakonst @DaniPopes

gakonst: Given they're grouped I am OK doing it that way

Cargo.toml
```diff
@@ -125,62 +126,33 @@ bytes = "1.3.0"
 # Tell `rustc` to optimize for small code size.
 opt-level = "s"
 
-[[example]]
-name = "abigen"
-path = "examples/abigen.rs"
-required-features = ["ethers-solc"]
+# [[example]]
```
gakonst, line 129: Remove

examples/contracts/examples/contract_deploy_anvil.rs
```diff
@@ -1,10 +1,11 @@
-use ethers::{
-    prelude::*,
-    solc::{Project, ProjectPathsConfig},
-    utils::Anvil,
-};
+use ethers_contract::{abigen, ContractFactory};
```
gakonst, line 1: We should communicate clearly that while these are the minimal crates, the ethers crate re exports everything and is easier to use unless you're trimming dependencies

examples/contracts/examples/contracts/contract.sol
```diff
@@ -6,7 +7,7 @@ contract SimpleStorage {
 
     string _value;
 
-    constructor(string memory value) public {
```
gakonst, line 9: Think this is required

examples/middleware/examples/builder.rs
```diff
@@ -0,0 +1,56 @@
...
+const RPC_URL: &str = "https://mainnet.infura.io/v3/c60b0bb42f8a4c6481ecd229eddaca27";
+const SIGNING_KEY: &str = "fdb33e2105f08abe41a8ee3b758726a31abdd57b7a443f470f23efce853af169";
+
+/// In ethers-rs, middleware is a way to customize the behavior of certain aspects of the library by
+/// injecting custom logic into the process of sending transactions and interacting with contracts
+/// on the Ethereum blockchain. The MiddlewareBuilder trait provides a way to define a chain of
+/// middleware that will be called at different points in this process, allowing you to customize
+/// the behavior of the Provider based on your needs.
+#[tokio::main]
+async fn main() {
```
gakonst, line 16: We should add an example of writing your own middleware. And an example showing how to pass middlewares around as M: Middleware. I've seen people get really confused at the type errors because they try to pass eg SignerMiddleWare<LocalWallet, Provider<Http>>

examples/middleware/examples/gas_oracle.rs
```diff
@@ -0,0 +1,95 @@
...
+async fn etherchain() {
+    let oracle = Etherchain::new().category(GasCategory::Standard);
+    match oracle.fetch().await {
+        Ok(gas_price) => println!("Gas price is {gas_price:?}"),
+        Err(e) => println!("Cannot estimate gas: {e:?}"),
+    }
+}
+
+async fn eth_gas_station() {
+    let api_key: Option<&str> = Some("YOUR-API-KEY");
```
gakonst, line 49: Let's make these environment variables so we run them on CI to check that these work?

examples/middleware/examples/policy.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    let policy = RejectEverything;
+    let policy_middleware = provider.wrap_into(|p| PolicyMiddleware::new(p, policy));
+
+    match policy_middleware.send_transaction(tx, None).await {
+        Err(e) => {
+            // Given the RejectEverything policy, we expect to execute this branch
+            assert!(matches!(e, PolicyMiddlewareError::PolicyError(())))
+        }
+        _ => (),
```
gakonst, line 33: Add a panic here to communicate that this won't get hit

gakonst (changes requested):
This is great. Love the new format. Let's make sure we can run the tests in CI, else can merge.

examples/contracts/examples/contracts/contract.sol
0xMelkor: You are right, this is required until Solidity v0.7.0 [(here)](https://docs.soliditylang.org/en/v0.7.0/070-breaking-changes.html#functions-and-events), and the contract is `>=0.4.24`

0xMelkor:
Hi guys,

CI works, but I have an open point on this:
What's the purpose of building examples? Do you expect to release them in the final artifact? 
Should I build examples along with their respective features (e.g. ipc feature for the provider:ipc example)?

Thank you!

prestwich:
The purpose of building examples is to ensure that the examples are still
correct through code drift and are updated alongside the underlying library
when the API changes. Same purpose as building doctests, basically.
Generally yes, examples should be built alongside their features

examples/big-numbers/README.md
```diff
@@ -0,0 +1,7 @@
+# Big numbers
+Ethereum uses big numbers (also known as "bignums" or "arbitrary-precision integers") to represent certain values in its codebase and in blockchain transactions. This is necessary because Ethereum ([the EVM](https://ethereum.org/en/developers/docs/evm/)) operates on a 256-bit word size, which is different from the usual 32-bit or 64-bit of modern machines (this has nothing to do with base-2 or base-10 which is how we choose to represent numbers). This was chosen for the ease of use with 256-bit cryptography (such as Keccak-256 hashes or secp256k1 signatures). (taken from [here](https://ethereum.org/en/developers/docs/evm/#evm-instructions))
```
DaniPopes, line 2:
```suggestion
Ethereum uses big numbers (also known as "bignums" or "arbitrary-precision integers") to represent certain values in its codebase and in blockchain transactions. This is necessary because [the EVM](https://ethereum.org/en/developers/docs/evm) operates on a 256-bit word size, which is different from the usual 32-bit or 64-bit of modern machines. This was chosen for the ease of use with 256-bit cryptography (such as Keccak-256 hashes or secp256k1 signatures).
```

DaniPopes (changes requested):
Looking good, just a few things to make big numbers cleaner

examples/middleware/examples/builder.rs
0xMelkor: Hi @gakonst
I provided an example in [custom_gas_middleware.rs](https://github.com/gakonst/ethers-rs/pull/1940/files#diff-6be7519191da8a934a30d0c69f6424106a3c1af298113f5db73d23d329af6575)

It provides a constructor that takes `inner: M` as an argument with a trait bound on M. 
If you remember specific cases where users got confused please let me know.

examples/contracts/examples/contract_deploy_anvil.rs
0xMelkor: I updated all manifests to reference the "ethers" crate. All imports follow the `ethers::` standard.

gakonst:
@0xMelkor This is looking good as a first step. 

RE: Examples build not passing
>  --> examples/contracts/examples/contract_events_with_meta.rs:4:38
>  --> examples/contracts/examples/contract_events_with_meta.rs:4:38
>   |
> 4 |     providers::{Provider, StreamExt, Ws},
>   |                                      ^^ no `Ws` in `providers`

You need to enable `features = ["ws"]` in the ethers-providers crate. Would be good to document that as well, as people have asked htat before.

You're also missing a `cargo fmt` on the lint job. Once we get these, we can merge and iterate! Thanks for the hard work.

0xMelkor:
Hi guys,
we are almost there, but I need to solve some points before merge:
1. Read RPC_URL from env
2. Update ToC
3. All examples are built with default features => update CI to provide special features if required

Most of READMEs in example crates are empty. Should we merge leaving them open for contribs?
Thanks!

gakonst:
Sounds good to me on all fronts! Thanks for taking this on.

examples/middleware/examples/gas_oracle.rs
0xMelkor: Suggestion applied. Just 2 points:
1. Etherchain [URL](https://www.etherchain.org/api/gasPriceOracle) is broken (http 404). I commented out the example.
2. I didn't find an API KEY for Blocknative. Anyway it's not required to call the API.

examples/middleware/examples/gas_escalator.rs
```diff
@@ -0,0 +1,58 @@
...
+    providers::{Http, Middleware, Provider},
+};
+use eyre::Result;
+
+/// The gas escalator middleware in ethers-rs is designed to automatically increase the gas cost of
+/// transactions if they get stuck in the mempool. This can be useful if you want to
+/// ensure that transactions are processed in a timely manner without having to manually adjust the
+/// gas cost yourself.
+#[tokio::main]
+#[allow(unused_must_use)]
```
DaniPopes, line 13: `.await.unwrap()` to use the results instead

examples/wallets/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+
+[[example]]
+name = "yubi"
+path = "examples/yubi.rs"
+required-features = ["yubi"]
+
+[[example]]
+name = "trezor"
+path = "examples/trezor.rs"
+required-features = ["trezor"]
```
DaniPopes, line 34: I think these `[[example]]` can be removed

DaniPopes (changes requested):
nits

I think we can also remove the prefix on example files:
```diff
- examples/big-numbers/examples/bn_conversion.rs
+ examples/big-numbers/examples/conversion.rs
```

examples/wallets/Cargo.toml
0xMelkor: You are right

0xMelkor:
> nits
> 
> I think we can also remove the prefix on example files:
> 
> ```diff
> - examples/big-numbers/examples/bn_conversion.rs
> + examples/big-numbers/examples/conversion.rs
> ```

Removed all prefixes. I also ensured that no broken references were left.

DaniPopes (approved):

mattsse (approved):
awesome

gakonst:
full send - nice work. we should take some time to write the book now :D cc @0xkitsune see these new examples

0xKitsune:
Awesome, thanks for the heads up. I'll check these out later today.

0xMelkor:
Hi guys, how should we contrib the book?
* I saw @0xKitsune provided a repo [here](https://github.com/0xKitsune/Tour-of-ethers-rs).
* Other guys suggested to inline the examples in the rust docs using nightly docs features
* Other ideas..

@gakonst @0xKitsune

gakonst:
I'd love to put stuff in the repo so PRs etc are merged in lockstep. I also want to respect Kit wanting to have their own repo, so I don't mind fragmentation of effort, although I'd obviously prefer if we made a new docs/book-type folder here and went ham there. I don't mind if it's an mdbook or the nightly thing.

0xKitsune:
>  I also want to respect Kit wanting to have their own repo, so I don't mind fragmentation of effort, although I'd obviously prefer if we made a new docs/book-type folder here and went ham there. I don't mind if it's an mdbook or the nightly thing.

As usual, I would be more than happy to combine efforts and work on the book/docs here. I made the Tour-of-ethers-rs repo before I knew that the official book was going to be developed. I was originally planning on writing something that walked through each part of the lib with examples along the way, kind of like the Rust Book (but a little less text heavy). Im definitely open to brainstorming on what the most effective resource might look like, feel free to add me to any conversations.

gakonst:
Sick - if you're down to combine then let's do it

PR #1942, opened 2022-12-16
Init devenv, gitignore .nlsp-settings

onsails:
## Motivation

[Nix](https://nixos.org/) is widely used in both rust and blockchain development as it enables reproducible builds which is crucial for dapps security and stability. This PR enables [direnv](https://devenv.sh/) setup for ethers-rs – it's a developer-friendly nix flakes wrapper. If accepted, this feature makes ethers-rs more accessible by contributions who rely on nix in their development cycle.
In a separate commit there is a git ignore of `.nlsp-settings` which is a directory for [project-local neovim lsp](https://github.com/tamago324/nlsp-settings.nvim) settings.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


gakonst (commented):
uh i don't know how to review this, @shazow can you maybe take a look?

shazow:
I haven't used direnv yet (just straight up nix) but I should try it! Will give it a try soon. Looks fairly straightforward.

onsails:
I can rework this on plain nix flakes which can also easily be integrated with direnv via `echo use_flake >> .envrc`. Devenv is just a dev-friendly wrapper around nix flakes.

shazow:
@onsails One of the nice things about having a flake is that automatically lets you use it as a dependency in another flake, such as for packaging foundry. I assume just having the direnv settings wouldn't be enough for that?

onsails:
Please see updated diff. Two birds one stone [via](https://devenv.sh/guides/using-with-flakes/).

.gitignore
```diff
@@ -2,3 +2,7 @@
 .vscode
 /.envrc
 .idea
+.nlsp-settings
+.devenv*
+devenv.local.nix
+.direnv
```
shazow, line 8: Is this redundant with `.devenv*`?

.gitignore
```diff
@@ -2,3 +2,7 @@
 .vscode
 /.envrc
 .idea
+.nlsp-settings
```
shazow, line 5: More up to @gakonst, but personally I just do `/.*` on my .gitignores and force-add whatever needs to be there (like .gitignore :sweat_smile:)

flake.nix
```diff
@@ -0,0 +1,58 @@
+{
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/release-22.11";
```
shazow, line 3: Is it worth pinning to a stable release? I usually omit specifying nixpkgs which is equivalent to using unstable by default.

I feel like rust development generally likes to track more cutting-edge versions of tooling, so unstable might be a better bet here. (In either case, the flake lock will pin to a specific revision still.)

flake.nix
```diff
@@ -0,0 +1,58 @@
...
+            };
+
+            # https://devenv.sh/pre-commit-hooks/
+            pre-commit.hooks = {
+              shellcheck.enable = true;
+
+              # nightly fmt is required here:
+              # https://github.com/gakonst/ethers-rs/blob/master/.github/workflows/ci.yml#L123
+              # while devenv doesn't yet support customizing it:
+              # https://github.com/cachix/devenv/issues/211
```
shazow, line 51: Thank you for adding these comments, really useful!

shazow (approved):
A few optional nits, but generally looks great to me. Appreciate that it works with plain `nix develop` as well as direnv! :)

Confirmed everything works and builds with:

```
$ nix develop
pre-commit-hooks.nix: updating /home/shazow/local/src/github.com/gakonst/ethers-rs repo
pre-commit installed at .git/hooks/pre-commit
(devenv)  $ cargo build
...
Finished dev [unoptimized + debuginfo] target(s) in 3m 06s
```

In the future might want to also add a `packages` output for using the flake as a dependency inside foundry, but this is great for now. Definitely makes my workflow easier.

@gakonst what do you think?

.gitignore
onsails: no, because it's d**IR**env :)

shazow: Ah good point. :)

flake.nix
onsails: agreed, force-pushed

onsails:
Also added .gitattributes to hide Cargo.lock and flake.lock diffs

onsails:
another force-push, enabled nightly rutstfmt and clippy pre-commit hooks via https://github.com/cachix/devenv/pull/228 (have to wait until it's merged)

shazow (approved):
Nice, looking good.

.gitignore
gakonst: Fine with me.as is!

gakonst (approved):
Defer to you guys on this, merging and hopefully addresses all your issues! Thanks for the patience.

shazow:
Oops yea that should have been ignored, gets installed automagically by the nix shell.

PR #1944, opened 2022-12-16
fix(middleware): oracles, tests

DaniPopes:
## Motivation

Old / non functioning API URLs

## Solution

- Replace gas_now url (old redirected to new)
- [deprecate eth_gas_station](https://twitter.com/ETHGasStation/status/1597341610777317376) and replace it in docs and tests with gas_now
- temporarily ignore etherchain tests which is failing in builds due to 404
- clean up parsing and gas calculation since every oracle was different
- add some `Default` impls and other derives
- export internal Response structs which were exposed in the query functions but were not exported at mod level
- add common Result type alias for gas_oracle
- add missing tests for oracles
- big tests diffs are just removing indentation from removing unnecessary `mod tests`
- add missing fields in gas_now Response
- make file structure consistent with each other (imports -> definitions -> [trait impls, impls])

*minor breaking changes:*
- make blocknative API key optional, which also makes the estimatedBaseFees response field optional since I think a key is required for that
- change some base oracles' methods from `request` to `query` for consistency
- change all oracle `query` methods to return just the Response struct and move parsing in other functions

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [x] Breaking changes

### DIFF
--- a/ethers-middleware/src/gas_oracle/etherchain.rs
+++ b/ethers-middleware/src/gas_oracle/etherchain.rs
@@ -30,7 +30,7 @@ pub struct Response {
 
 impl Response {
     #[inline]
-    fn gas_from_category(&self, gas_category: GasCategory) -> f64 {
+    pub fn gas_from_category(&self, gas_category: GasCategory) -> f64 {
         match gas_category {
             GasCategory::SafeLow => self.safe_low,
             GasCategory::Standard => self.standard,
--- a/ethers-middleware/src/gas_oracle/etherchain.rs
+++ b/ethers-middleware/src/gas_oracle/etherchain.rs
@@ -1,6 +1,6 @@
-use super::{GasCategory, GasOracle, GasOracleError, Result, GWEI_TO_WEI_U256};
+use super::{from_gwei_f64, GasCategory, GasOracle, GasOracleError, Result};
 use async_trait::async_trait;
-use ethers_core::types::{u256_from_f64_saturating, U256};
+use ethers_core::types::U256;
 use reqwest::Client;
 use serde::Deserialize;
 use url::Url;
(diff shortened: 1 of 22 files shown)


mattsse (changes requested):
tysm!

It looks like this includes a lot of valuable but separate contributions, including some breaking changes I think.

I'd rather merge this in separate PRs, so it would be easier to merge this step by step.

Could you please split this as you see fit?

DaniPopes:
What do you mean by easier to merge? I think a lot of these changes will overlap with each other which would have the opposite effect - I don't think there's much point in splitting this, it's mostly just making the individual oracles and file structure consistent + fixing a few bugs and adding a few tests

I finished listing all changes and breaking changes

mattsse:
> What do you mean by easier to merge? I think a lot of these changes will overlap with each other which would have the opposite effect - I don't think there's much point in splitting this

this is good to know.

I just wanted to check this out before looking at it more closely because it looks like it has a lot of valuable corrections in it.

ethers-middleware/src/gas_oracle/etherchain.rs
```diff
@@ -1,32 +1,63 @@
...
 use async_trait::async_trait;
+use ethers_core::types::U256;
 use reqwest::Client;
 use serde::Deserialize;
 use url::Url;
 
-use crate::gas_oracle::{GasCategory, GasOracle, GasOracleError, GWEI_TO_WEI};
-
-const ETHERCHAIN_URL: &str = "https://www.etherchain.org/api/gasPriceOracle";
+const URL: &str = "https://www.etherchain.org/api/gasPriceOracle";
```
gakonst, line 8: https://twitter.com/milancermak/status/1604441584808103936 maybe we can use this https://etherchain.org/api/gasnow, which redirects to https://beaconcha.in/api/v1/execution/gasnow

gakonst (approved):

PR #1945, opened 2022-12-16
fix(msrv): bump MSRV in .clippy.toml and add a comment in root Cargo.toml

DaniPopes:
## Motivation

.clippy.toml msrv was not bumped with the others.

.clippy.toml is for backwards compatibility since the rust-version key in the Cargo manifest is [ignored before 1.56.0](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) and clippy will display a warning if it differs from the manifests

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


mattsse (approved):
nice catch, the ci error seems unrelated, seems like an API is down.

PR #1946, opened 2022-12-17
fix(core): decode to correctly in Transaction

### LINKED ISSUE #1939: ethers_core::types::Transaction fails to decode from RLP when "to" is None
**Version**
├── ethers-core v1.0.2
├── ethers-core v1.0.2 (*)
    ├── ethers-core v1.0.2 (*)
├── ethers-core v1.0.2 (*)
│   │   │   ├── ethers-core v1.0.2 (*)
│   │   │   ├── ethers-core v1.0.2 (*)
│   │   │   ├── ethers-core v1.0.2 (*)
│   │   ├── ethers-core v1.0.2 (*)
│   │   │   ├── ethers-core v1.0.2 (*)
│   │   │   ├── ethers-providers v1.0.0
│   │   │   │   ├── ethers-core v1.0.2 (*)
├── ethers v1.0.2
│   ├── ethers-addressbook v1.0.2
│   │   ├── ethers-core v1.0.2 (*)
│   ├── ethers-contract v1.0.2
│   │   ├── ethers-contract-abigen v1.0.2
│   │   │   ├── ethers-core v1.0.2 (*)
│   │   ├── ethers-contract-derive v1.0.2 (proc-macro)
│   │   │   ├── ethers-contract-abigen v1.0.2 (*)
│   │   │   ├── ethers-core v1.0.2 (*)
│   │   ├── ethers-core v1.0.2 (*)
│   │   ├── ethers-providers v1.0.2
│   │   │   ├── ethers-core v1.0.2 (*)
│   ├── ethers-core v1.0.2 (*)
│   ├── ethers-etherscan v1.0.2
│   │   ├── ethers-core v1.0.2 (*)
│   ├── ethers-middleware v1.0.2
│   │   ├── ethers-contract v1.0.2 (*)
│   │   ├── ethers-core v1.0.2 (*)
│   │   ├── ethers-etherscan v1.0.2 (*)
│   │   ├── ethers-providers v1.0.2 (*)
│   │   ├── ethers-signers v1.0.2
│   │   │   ├── ethers-core v1.0.2 (*)
│   ├── ethers-providers v1.0.2 (*)
│   └── ethers-signers v1.0.2 (*)

But also is reproducible on master branch

**Platform**
5.15.0-52-generic #58-Ubuntu SMP Thu Oct 13 08:03:55 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

**Description**
ethers_core::types::Transaction RLP decoding fails in case when `to` is None. According to the Ethereum yellow paper missing `to` should be RLP encoded in a different way:

> The address hash Tt is slightly different: it is either a  20-byte address hash or, in the case of being a contract-creation transaction (and thus formally equal to ∅), it is the RLP empty byte sequence and thus the member of B0

The reason is because optional members in Transaction are encoded to RLP but during decoding they are assumed to always be Some. At least for "to" it is completely legit to be None.

I tried this code:

```
let tx = Transaction {
            block_hash: None,
            block_number: None,
            from: Address::from_str("c26ad91f4e7a0cad84c4b9315f420ca9217e315d").unwrap(),
            gas: U256::from_str_radix("0x10e2b", 16).unwrap(),
            gas_price: Some(U256::from_str_radix("0x12ec276caf", 16).unwrap()),
            hash: H256::from_str("929ff27a5c7833953df23103c4eb55ebdfb698678139d751c51932163877fada").unwrap(),
            input: Bytes::from(
                hex::decode("a9059cbb000000000000000000000000fdae129ecc2c27d166a3131098bc05d143fa258e0000000000000000000000000000000000000000000000000000000002faf080").unwrap()
            ),
            nonce: U256::zero(),
            transaction_index: None,
            value: U256::zero(),
            other: Default::default()
        };
assert!(Transaction::decode(RLP::new(tx.rlp())).is_ok());
```

I expected to see this happen: roundtrip to RLP works for transactions with `to` equal to `None`.

Instead, this happened: `decode` function returns error trying to decode `None` value.

Rjected:
## Motivation

Currently `Transaction`s do not decode correctly if they are contract creation transactions. This is when the `to` field is `None`, which is encoded in RLP as `0x80`. 

## Solution

Refactor the `decode_to` method to instead return `Result<Option<Address>, DecodeError>`, and use it when decoding `Transaction`s, not just transaction requests.

Add a test from #1393 that passes with the changes.

Fixes #1939

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/types/transaction/eip1559.rs
+++ b/ethers-core/src/types/transaction/eip1559.rs
@@ -208,7 +208,7 @@ impl Eip1559TransactionRequest {
         *offset += 1;
         tx.gas = Some(rlp.val_at(*offset)?);
         *offset += 1;
-        tx.to = decode_to(rlp, offset)?;
+        tx.to = decode_to(rlp, offset)?.map(NameOrAddress::Address);
         tx.value = Some(rlp.val_at(*offset)?);
         *offset += 1;
         let data = rlp::Rlp::new(rlp.at(*offset)?.as_raw()).data()?;
--- a/ethers-core/src/types/transaction/mod.rs
+++ b/ethers-core/src/types/transaction/mod.rs
@@ -75,7 +75,7 @@ fn decode_signature(
 fn decode_to(
     rlp: &rlp::Rlp,
     offset: &mut usize,
-) -> Result<Option<super::NameOrAddress>, rlp::DecoderError> {
+) -> Result<Option<super::Address>, rlp::DecoderError> {
     let to = {
         let to = rlp.at(*offset)?;
         if to.is_empty() {
--- a/ethers-core/src/types/transaction/request.rs
+++ b/ethers-core/src/types/transaction/request.rs
@@ -233,7 +233,7 @@ impl TransactionRequest {
             *offset += 1;
         }
 
-        txn.to = decode_to(rlp, offset)?;
+        txn.to = decode_to(rlp, offset)?.map(NameOrAddress::Address);
         txn.value = Some(rlp.at(*offset)?.as_val()?);
         *offset += 1;
 
--- a/ethers-core/src/types/transaction/response.rs
+++ b/ethers-core/src/types/transaction/response.rs
@@ -1,7 +1,7 @@
 //! Transaction types
 use super::{
-    decode_signature, eip2718::TypedTransaction, eip2930::AccessList, normalize_v, rlp_opt,
-    rlp_opt_list,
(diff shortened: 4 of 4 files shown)


mattsse (approved):
lgtm

gakonst (approved):

PR #1947, opened 2022-12-17
feat: make geth executable configurable

mattsse:
## Motivation
make it configurable whcih geth bin to use.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/utils/geth.rs
+++ b/ethers-core/src/utils/geth.rs
@@ -164,6 +164,7 @@ impl Default for PrivateNetOptions {
 /// ```
 #[derive(Clone, Default)]
 pub struct Geth {
+    program: Option<PathBuf>,
     port: Option<u16>,
     authrpc_port: Option<u16>,
     ipc_path: Option<PathBuf>,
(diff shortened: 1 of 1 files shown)


ethers-core/src/utils/geth.rs
```diff
@@ -180,6 +181,32 @@ impl Geth {
         Self::default()
     }
 
+    /// Creates an Geth builder which will execute `geth` at the given path.
```
onbjerg, line 184:
```suggestion
    /// Creates a Geth builder which will execute `geth` at the given path.
```

ethers-core/src/utils/geth.rs
```diff
@@ -180,6 +181,32 @@ impl Geth {
...
+    ///  let geth = Geth::at("../go-ethereum/build/bin/geth").spawn();
+    ///
+    ///  println!("Geth running at `{}`", geth.endpoint());
+    /// # }
+    /// ```
+    pub fn at(path: impl Into<PathBuf>) -> Self {
+        Self::new().path(path)
+    }
+
+    /// Sets the `path` to the `geth` cli
```
onbjerg, line 200:
```suggestion
    /// Sets the `path` to the `geth` executable
```

onbjerg (approved):

PR #1961, opened 2022-12-21
fix: Add `openssl` and `rustls` feature flags in ethers-middleware

oblique:
## Motivation

I was working on a project with requirement to use OpenSSL, so I added this:

```
ethers = { version = "1.0.2", default-features = false, features = ["ws", "openssl"] }
```

But I noticed that `rustls` was still building.

## Solution

The problem was that `ethers-middleware` was explicitly using `reqwest` with `rustls-tls` enabled, so I added `rustls` and `openssl` feature flags to it.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [x] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -341,6 +341,8 @@
 
 ### Unreleased
 
+- Added `openssl` and `rustls` feature flags
+  [#1961](https://github.com/gakonst/ethers-rs/pull/1961)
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm

PR #1964, opened 2022-12-21
feat(solc): create compilation unit for files that share solc version/ settings

0xalpharush:
## Motivation

Currently, the cache file stores the compiler version and its settings individually for each file, and it can be ambiguous what compilation artifacts are a "unit"; that is, they were determined to be compatible and/or necessary e.g. imports and then compiled with the same solc version and compiler settings. This is mainly a concern when multiple compiler versions are used and the artifacts can then contain duplicate id's in the source mappings (see https://github.com/crytic/slither/issues/1213). 

Ref https://github.com/foundry-rs/foundry/issues/3450
## Solution
This PR changes the data model of the artifacts to instead store a unique identifier representing that the source files were compiled together (`CompilationUnitId`). This identifier maps back to a `CompilationUnit` which stores the version, configuration, and source files' paths. Rather than redundantly store the `SolcConfig` for each `CacheEntry`, it is now looked up by its `CompilationUnitId`.  Now, the each compilation unit can be parsed individually and lookup where the artifact is in the cache entry's `files`. This [diff](https://github.com/0xalpharush/foundry/commit/61f054038329e7a2eae175993487bdf4c3ab6ef2) in Foundry is pretty small fwiw.

I'm new to Rust and also may have overlooked some design decision that my changes are incompatible with so I'm very open to feedback!

## PR Checklist

-   [x] Added Tests
-   [X] Added Documentation
-   [ ] Updated the changelog
-   [X] Breaking changes

### DIFF
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -284,7 +284,7 @@ impl From<CompilerInput> for StandardJsonCompilerInput {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Settings {
     /// Stop compilation after the given stage.
@@ -513,6 +513,17 @@ impl Hash for Settings {
             (revert_strings.unwrap_or_default() as u8).hash(state);
             debug_info.hash(state);
         }
+    }
+}
+
+impl PartialEq for Settings {
+    fn eq(&self, other: &Self) -> bool {
+        self.optimizer == other.optimizer &&
+            self.metadata == other.metadata &&
+            self.output_selection == other.output_selection &&
+            self.evm_version == other.evm_version &&
+            self.via_ir == other.via_ir &&
+            self.debug == other.debug
     }
 }
 
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -9,6 +9,7 @@ use serde::{de::Visitor, Deserialize, Deserializer, Serialize, Serializer};
 use std::{
     collections::{BTreeMap, HashSet},
     fmt, fs,
+    hash::{Hash, Hasher},
     path::{Path, PathBuf},
     str::FromStr,
 };
@@ -494,6 +495,24 @@ impl Default for Settings {
             model_checker: None,
         }
         .with_ast()
+    }
+}
+
+impl Hash for Settings {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.optimizer.enabled.hash(state);
+        self.optimizer.runs.hash(state);
+        if let Some(SettingsMetadata { bytecode_hash, cbor_metadata, .. }) = self.metadata {
+            (bytecode_hash.unwrap_or_default() as u8).hash(state);
+            cbor_metadata.hash(state);
+        }
+        self.output_selection.0.keys().cloned().collect::<String>().hash(state);
+        (self.evm_version.unwrap_or_default() as u8).hash(state);
+        self.via_ir.hash(state);
+        if let Some(DebuggingSettings { revert_strings, debug_info, .. }) = &self.debug {
+            (revert_strings.unwrap_or_default() as u8).hash(state);
+            debug_info.hash(state);
+        }
     }
 }
 
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -9,7 +9,6 @@ use serde::{de::Visitor, Deserialize, Deserializer, Serialize, Serializer};
 use std::{
     collections::{BTreeMap, HashSet},
     fmt, fs,
-    hash::{Hash, Hasher},
     path::{Path, PathBuf},
     str::FromStr,
 };
@@ -284,7 +283,7 @@ impl From<CompilerInput> for StandardJsonCompilerInput {
     }
 }
 
-#[derive(Clone, Debug, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Hash)]
 #[serde(rename_all = "camelCase")]
 pub struct Settings {
     /// Stop compilation after the given stage.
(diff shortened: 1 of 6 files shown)


plotchy:
The attached slither issue shows one of the most ubiquitous, difficult to diagnose bugs in slither. very nice job pursuing a fix

ethers-solc/src/artifacts/mod.rs
```diff
@@ -494,6 +495,35 @@ impl Default for Settings {
...
+        (self.evm_version.unwrap_or_default() as u8).hash(state);
+        self.via_ir.hash(state);
+        if let Some(DebuggingSettings { revert_strings, debug_info, .. }) = &self.debug {
+            (revert_strings.unwrap_or_default() as u8).hash(state);
+            debug_info.hash(state);
+        }
+    }
+}
+
+impl PartialEq for Settings {
```
mattsse, line 516: what's the reason for the manual impls here?

manual hash impl for settings and including the metadatahash directly makes sense tho

ethers-solc/src/cache.rs
```diff
@@ -25,11 +26,34 @@ use std::{
...
 /// hardhat This allows ethers-solc to detect if the cache file was written by hardhat or
 /// `ethers-solc`
-const ETHERS_FORMAT_VERSION: &str = "ethers-rs-sol-cache-3";
+const ETHERS_FORMAT_VERSION: &str = "ethers-rs-sol-cache-4";
 
 /// The file name of the default cache file
 pub const SOLIDITY_FILES_CACHE_FILENAME: &str = "solidity-files-cache.json";
 
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, Hash)]
+pub struct CompilationUnitId(String);
```
mattsse, line 35: should this take the hasher outcome u64 instead of String and print hex via Display impl?

mattsse, line 35: needs some docs

ethers-solc/src/cache.rs
```diff
@@ -25,11 +26,34 @@ use std::{
...
+    }
+}
+
+/// A unit of source files that are compiled together with the same compiler version and settings
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CompilationUnit {
+    pub solc_config: SolcConfig,
+    pub version: Version,
+    pub source_units: HashSet<PathBuf>,
```
mattsse, line 54: let's make this ordered?

```suggestion
    pub source_units: BTreeMap<PathBuf>,
```

ethers-solc/src/cache.rs
```diff
@@ -38,12 +62,18 @@ pub struct SolFilesCache {
     /// contains all directories used for the project
     pub paths: ProjectPaths,
     pub files: BTreeMap<PathBuf, CacheEntry>,
+    #[serde(rename = "compilationUnits")]
+    pub compilation_units: HashMap<CompilationUnitId, CompilationUnit>,
```
mattsse, line 66:
```suggestion
    pub compilation_units: BTreeMap<CompilationUnitId, CompilationUnit>,
```

ethers-solc/src/config.rs
```diff
@@ -745,6 +746,18 @@ impl From<SolcConfig> for Settings {
...
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.settings.hash(state);
+    }
+}
+
+impl PartialEq for SolcConfig {
+    fn eq(&self, other: &Self) -> bool {
+        self.settings == other.settings
+    }
+}
```
mattsse, line 759: can just youse derived impls instead?

ethers-solc/src/cache.rs
```diff
@@ -25,11 +26,34 @@ use std::{
...
 
 /// The file name of the default cache file
 pub const SOLIDITY_FILES_CACHE_FILENAME: &str = "solidity-files-cache.json";
 
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, Hash)]
+pub struct CompilationUnitId(String);
+
+impl CompilationUnitId {
+    /// Create a unique id for each compilation unit based on compiler version and settings
+    pub fn new(version: Version, solc_config: SolcConfig) -> Self {
```
mattsse, line 39: can take args as a reference?

mattsse (changes requested):
thanks for this.

Given that this will improve slither integration, I'm supportive of this @gakonst.

on thing I'm not certain of is, how this should/will affect partial recompilation on changes. 
Currently, we only include the necessary files needed by a modified file. would the compileIdunit still be valid if the settings did not change?

compiling the entire unit on a single change would not be feasible imo

gakonst:
> Given that this will improve slither integration, I'm supportive of this @gakonst.

Same. Great change, thanks for taking this on @0xalpharush.

> on thing I'm not certain of is, how this should/will affect partial recompilation on changes.
> Currently, we only include the necessary files needed by a modified file. would the compileIdunit still be valid if the settings did not change?
> compiling the entire unit on a single change would not be feasible imo

Agreed. Let's add a test for this to check what happens to partial recompilation?

gakonst (changes requested):
Would like an extra test for the partial recompilation case that Matt mentioned. Appreciate you tackling this.

ethers-solc/src/cache.rs
0xalpharush: https://github.com/gakonst/ethers-rs/pull/1964/commits/80c3e3e4959bcb975d43af313b292d6cd2c68d1e

ethers-solc/src/artifacts/mod.rs
0xalpharush: I derived Hash impls where necessary https://github.com/gakonst/ethers-rs/pull/1964/commits/94e9415b87d21e396ca4a1d1bce38dd3519e3237
> manual hash impl for settings and including the metadatahash directly makes sense tho
Are you referring to `SettingsMetadata`?

ethers-solc/src/cache.rs
0xalpharush: https://github.com/gakonst/ethers-rs/pull/1964/commits/28bf56149c98ed8a87319f0ba2869d3481225159

ethers-solc/src/cache.rs
0xalpharush: https://github.com/gakonst/ethers-rs/pull/1964/commits/4439d1435cf44b8cb76d7d82ebc52a051ee99e47

0xalpharush:
> Currently, we only include the necessary files needed by a modified file. would the compileIdunit still be valid if the settings did not change?

@mattsse AFAICT I didn't modify the detection of "clean" and "dirty" sources (there is an existing test for [this](https://github.com/gakonst/ethers-rs/blob/799f752e55e80aa8d6357321d4f5f7ebf66daa8a/ethers-solc/src/compile/project.rs#L726)). The only time an entire compilation unit should be recompiled is if the settings are changed for the entire unit e.g. an incompatible solidity version, creating another compilation unit derived from the incompatible settings. I added a [test case](https://github.com/gakonst/ethers-rs/pull/1964/commits/98d823c74a26c386d276d87dd0e74aae617de8dc) to show that the compilation unit id remains valid if only a file's contents are changed and that only the dirty source is recompiled.

gakonst:
@mattsse PTAL I think this is starting to look good

ethers-solc/src/cache.rs
mattsse: this is written to disk, so we want consistent order

mattsse (approved):
thanks for clarifying @0xalpharush 

changes and test look good, let's rebase this PR then the foundry companion, good to merge when the companion passes.

0xalpharush:
Rebased. I think the ethereum test failure is unrelated:
```
---- nonce_manager stdout ----
thread 'nonce_manager' panicked at 'called `Result::unwrap()` on an `Err` value: MiddlewareError(MiddlewareError(JsonRpcClientError(JsonRpcError(JsonRpcError { code: -32000, message: "replacement transaction underpriced", data: None }))))', ethers-middleware/tests/nonce_manager.rs:42:14
```
Updated the Foundry PR https://github.com/foundry-rs/foundry/pull/3939

gakonst (approved):
LGTM pending @mattsse approval

ethers-solc/src/cache.rs
```diff
@@ -411,8 +448,8 @@ pub struct CacheEntry {
     pub content_hash: String,
     /// identifier name see [`crate::utils::source_name()`]
     pub source_name: PathBuf,
-    /// what config was set when compiling this file
-    pub solc_config: SolcConfig,
+    /// what compilation unit this file was a part of
+    pub compilation_unit: CompilationUnitId,
```
0xalpharush, line 452: @gakonst One thing I realized chasing down #1995 is that each cache entry should have multiple compilation units instead of just one so maybe the compilation unit should be stored at the artifact level. This would allow the cache to handle having the same version being compiled with different compiler settings e.g. someone doesn't want to use via-IR while testing. I explore this in a separate branch but not sure it's a great solution (https://github.com/0xalpharush/ethers-rs/commit/c378b495f712aae2cc077d6d3e50ab4def1e67bd).

mattsse: this is a reasonable consideration and would give us a bit more flexibility I think.

can't think of any downsides to moving this to the artifact level

gakonst: SGTM.

gakonst:
@0xalpharush friendly ping here, no pressure again just dont want this to get stale

0xalpharush:
Will get back to this when I have the time. I started over holiday break and have been busy recently... If anyone finds a solution to https://github.com/gakonst/ethers-rs/issues/1995, that would be great to have for the time being.

gakonst:
ACK. -thx! #1995 up for grabs.

gakonst:
Closing as stale.

PR #1943, opened 2022-12-16
fix: rebase to master for onbjerg's ast

iFrostizz:
## Motivation

Following up on onbjerg's PR: https://github.com/gakonst/ethers-rs/pull/1567 to make the files play nicely with the master branch, the merge from master to the ast branch was deleting some files on the last commit.

## PR Checklist

-   [ ] Check for foundry dependencies (coverage?)

PR #1966, opened 2022-12-22
revert: Ast Artifact

mattsse:
## Motivation
This reverts parts of #1943 because that PR did not pass all checks and will likely cause a lot of issues in foundry just because there are a lot of different version-specific variants which are likely not covered.

My suggesting is:
1. partially revert #1943 breaking changes with this PR
2. fix tests first
3. more tests for all versions
4. reconsider replacing Ast with SourceUnit

Since the AST is not of interest in most cases, we ca also handle it similar to Metadata where we just use the `Json` and add a separate function to deserialize the `SourceUnit` this makes it less error prone for foundry use case and you can still get the deserialized object via an additional call if you need that.

wdyt @iFrostizz ?

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-solc/src/artifact_output/configurable.rs
+++ b/ethers-solc/src/artifact_output/configurable.rs
@@ -16,9 +16,9 @@ use crate::{
             BytecodeOutputSelection, ContractOutputSelection, EvmOutputSelection,
             EwasmOutputSelection,
         },
-        CompactContractBytecodeCow, DevDoc, Evm, Ewasm, FunctionDebugData, GasEstimates,
-        GeneratedSource, LosslessAbi, LosslessMetadata, Metadata, Offsets, Settings, SourceUnit,
-        StorageLayout, UserDoc,
+        Ast, CompactContractBytecodeCow, DevDoc, Evm, Ewasm, FunctionDebugData, GasEstimates,
+        GeneratedSource, LosslessAbi, LosslessMetadata, Metadata, Offsets, Settings, StorageLayout,
+        UserDoc,
     },
     sources::VersionedSourceFile,
     ArtifactOutput, SolcConfig, SolcError, SourceFile,
@@ -68,7 +68,7 @@ pub struct ConfigurableContractArtifact {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub ewasm: Option<Ewasm>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub ast: Option<SourceUnit>,
+    pub ast: Option<Ast>,
     /// The identifier of the source file
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub id: Option<u32>,
--- a/ethers-solc/src/artifacts/ast/lowfidelity.rs
+++ b/ethers-solc/src/artifacts/ast/lowfidelity.rs
@@ -1,0 +1,217 @@
+//! Bindings for solc's `ast` output field
+
+use crate::artifacts::serde_helpers;
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::{collections::BTreeMap, fmt, fmt::Write, str::FromStr};
+
+/// Represents the AST field in the solc output
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Ast {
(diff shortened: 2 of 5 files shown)


iFrostizz:
Sure, will do that. Sorry for the noise

iFrostizz:
Just to clarify @mattsse , so as you probably don't want me to use my fork of ethers-rs as a dep of foundry, should we pin it to a rev, then merge on ethers-rs, run all the tests to make sure that it works, merge, then open the PR on the foundry side, change the deps to latest master of ethers-rs and make sure that all tests are all good. Correct ?

gakonst:
Ya I think you maybe chose the wrong Rev?

iFrostizz:
> Ya I think you maybe chose the wrong Rev?

If you look at the rev `bb4af1c` is wasn't including the AST at all. It was the last commit on the master branch.
My guess is that you have merged the PR on ethers side first which is why everything was so broken.

gakonst:
Yeah the right way would've been to specify "branch =..." With your own git link instead of the Rev. It's ok no problem.

iFrostizz:
> Yeah the right way would've been to specify "branch =..." With your own git link instead of the Rev. It's ok no problem.

If the dep is pinned to latest rev there should not be any issue in case I do anything wrong with my branch so if we keep it to official repo it's safer probably

mattsse:
perhaps I'm misunderstanding, but the whole point of testing against foundry is using a specific ethers PR, which would be _your_ repository and branch, using rev on master branch defeats that purpose entirely.

See for example: https://github.com/foundry-rs/foundry/pull/3976

iFrostizz:
> perhaps I'm misunderstanding, but the whole point of testing against foundry is using a specific ethers PR, which would be _your_ repository and branch, using rev on master branch defeats that purpose entirely.
> 
> See for example: https://github.com/foundry-rs/foundry/pull/3976
> 

Sure we could do that. Just wanted to make sure that in case anything goes wrong on my branch, that won't make foundry user grumpy.
There is probably some misunderstanding, the initial PR would be to pin the ethers-rs to the last rev (without breaking ast), merge ethers-rs ast, and finally open a second PR on foundry to fix everything.
Does that sounds wrong ?

gakonst:
Yeah that's wrong.

The process is:
1. Open ethers PR, get it reviewed 
2. Open foundry PR, using the ethers PR (not latest master Rev), get it reviewed 
3. Once both PRs look good, we merge ethers PR and then switch foundry to use latest ethers master and merge the Foundry PR

This ensures that your Foundry PR uses the ethers changes. 

Hope that makes sense.

iFrostizz:
> Yeah that's wrong.
> 
> The process is:
> 1. Open ethers PR, get it reviewed 
> 2. Open foundry PR, using the ethers PR with the breaking changes (not latest master Rev), get it reviewed 
> 3. Once both PRs look good, we merge ethers PR and then switch foundry to use latest ethers master and merge the Foundry PR
> 
> This ensures that your Foundry PR uses the ethers changes. 
> 
> Hope that makes sense.
> 

It's pretty much the same idea, but in this case we would use my fork as a branch for a process with less friction correct ?

gakonst:
It's not the same process clearly because it resulted in us merging breaking changes 😅 you do need to pin to a commit but we do it via branch and you do it via the Rev which is fine.

iFrostizz:
I think that you merged it in the wrong order so foundry was relying on ethers with breaking ast

gakonst:
Maybe :)

mattsse:
I don't understand this, because this shouldn't have any effect if the foundry PR would be patched to the ethers PR that introduces breaking changes.

iFrostizz:
Now using your method. Pinning on my fork's rev and doing stuff there.
As your idea, I'll make it retro-compatible with the old Ast, in case some properties of the ast are forgot, during the serialization the whole coverage will break too this is not desirable. Also trying to add a function to serialize it to the typed Ast. Does that sound good ?
Additionally, I will try to add more constrained tests for the Ast, because for now it's just checking if it doesn't panic during serialization. Idk if it's possible to compare the raw view with the typed one but it would be a great detector of forgot nodes

PR #1976, opened 2022-12-27
feat: windows ipc provider (named pipe)

### LINKED ISSUE #1974: New "named_pipe" provider for intercommunication with Nethermind client on Windows
I want to use named pipe channel from Nethermind client on windows for intercommunication using jsonrpc messages ( the same behavior as Ipc provider)

New provider, ie. ethers_provider::named_pipe

-

**Additional context**
We can use https://docs.rs/tokio/latest/tokio/net/windows/named_pipe/ crate for named pipes functionality

DaniPopes:
## Motivation

Fixes #1974, cc @CodeFunta

## Solution

- Add a custom `tokio::net::windows::named_pipe::NamedPipeClient` wrapper to be compatible with `UnixStream`
  - might look intimidating but it's just re-implementing the same traits and the `split` method, only real implementation is in `connect`
- Bump server thread stack size from 64kb to 256kb

Successfully tested after a bump in thread stack size

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


DaniPopes:
P.S.: CI needs quite the revamp, it's currently only covering Windows with no features and Linux with either no or all features
@mattsse maybe just copy it over from Foundry?

mattsse (commented):
thanks for picking this up.

before we merge this, it'd be great to have additional ipc tests in anvil first to test the provider.

DaniPopes:
> thanks for picking this up.
> 
> before we merge this, it'd be great to have additional ipc tests in anvil first to test the provider.

the tests are the same as websockets, do you want me to change from geth to anvil?

mattsse:
we currently only do ipc e2e testing on unix for anvil

https://github.com/foundry-rs/foundry/blob/6ecebad62014ac46513099f03e8c0b80f5d9d4b6/anvil/tests/it/main.rs#L10-L11

with this we can remove the cfg and it should also work on windows, so would be great to test this first via a PR with patched ethers

DaniPopes:
Oh I see what you mean now, on it

ethers-providers/Cargo.toml
```diff
@@ -46,9 +46,12 @@ bytes = { version = "1.3.0", default-features = false, optional = true }
 once_cell = "1.16.0"
 hashers = "1.0.1"
 
+[target.'cfg(target_family = "windows")'.dependencies]
+winapi = "0.3"
+
```
mattsse, line 51: we only need this if ipc is enabled, right?
could try to make optional?

ethers-providers/src/transports/ipc.rs
```diff
@@ -1,48 +1,217 @@
...
+                }
+
+                sleep(Duration::from_millis(50)).await;
+            }
+        }
+
+        #[allow(unsafe_code)]
+        pub fn split(&mut self) -> (ReadHalf, WriteHalf) {
+            // SAFETY: ReadHalf cannot write but still needs a mutable reference for polling.
+            // NamedPipeClient calls its `io` using immutable references, but it's private.
```
mattsse, line 102: I believe this is correct, there can be two tasks that use the underlying `io` as long as one only reads and the other only writes.

`UnixStream` basically does the same but implements it slightly differently:

it uses immutable references in the halves

https://docs.rs/tokio/latest/src/tokio/net/unix/split.rs.html#51-53

and uses unsafe in separate poll functions that use unsafe

https://docs.rs/tokio/latest/src/tokio/net/unix/stream.rs.html#944-950

perhaps we mimick this as well?

mattsse (changes requested):
foundry tests lgtm.

could use the windows stream as is or replicate the Windows stream implementation, both are essentially the same.

ethers-providers/src/transports/ipc.rs
DaniPopes: the problem is the read/write trait poll methods need a mutable reference because self is `Pin<&mut self>` so that's why I'm storing a mut ref instead of immutable. Otherwise if the inner `io` was public I'd be calling that since it uses immutable references

mattsse: makes sense

mattsse (approved):
lgtm,

cross platform ipc support 🔥

gakonst:
nice work :)

CodeFunta:
Thank you team!

PR #1977, opened 2022-12-27
ethers-signers/aws: sighash on the inner/updated tx object

gyuho:
Signed-off-by: Gyuho Lee <gyuho.lee@avalabs.org>

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


gakonst (approved):
Good catch. Hard to test these -_-

PR #1980, opened 2022-12-28
fix: don't override user-set 1559 attributes

prestwich:
## Motivation

Allow user to set tip without setting base fee, and avoid overriding any user-set tip or basefee

## Solution

instead of overriding, use the `Option::get_or_insert`

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


prestwich:
CI failure is related. Need to change the set so that if max priority fee can't be higher than max fee

prestwich:
CI failure should be fixed now 🖖

gakonst (approved):

gakonst:
ACK - good find, let's give this a try

PR #1983, opened 2022-12-29
added `from_bytes` for `Wallet` type

0xKitsune:
## Motivation
There was no way to initialize a new Wallet from a byte slice.

## Solution
I added a `from_bytes` associated function for the `Wallet` type. The function  initializes a new `SignerKey` from a byte slice and uses the key to create a new wallet.

```rust
impl Wallet<SigningKey> {
    //--snip--
    pub fn from_bytes(bytes: &[u8]) -> Result<Self, WalletError> {
        let signer = SigningKey::from_bytes(bytes)?;
        let address = secret_key_to_address(&signer);
        Ok(Self { signer, address, chain_id: 1 })
    }
 }
 ```

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-signers/src/wallet/private_key.rs
+++ b/ethers-signers/src/wallet/private_key.rs
@@ -90,13 +90,20 @@ impl Wallet<SigningKey> {
         let address = secret_key_to_address(&signer);
         Self { signer, address, chain_id: 1 }
     }
+
+    /// Creates a new Wallet instance from a raw scalar value (big endian).
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, WalletError> {
+        let signer = SigningKey::from_bytes(bytes)?;
+        let address = secret_key_to_address(&signer);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: ethers-signers/src/wallet/private_key.rs [hunk 1]
```diff
@@ -90,6 +90,13 @@ impl Wallet<SigningKey> {
         let address = secret_key_to_address(&signer);
         Self { signer, address, chain_id: 1 }
     }
+
+    /// Creates a new Wallet instance from a raw scalar value (big endian).
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, WalletError> {
```
gakonst, line 95: nit: `impl AsRef<[u8]>` is slightly better but no need

gakonst (approved):

Issue #1984, opened 2022-12-29
Typo in docs regarding Multicall::call_array


akramhussein:
**Version**

`master` branch for all - `10310ce3`

**Description**

https://github.com/gakonst/ethers-rs/blob/10310ce3ad8562476196e9ec06f78c2a27417739/ethers-contract/src/multicall/mod.rs#L664

Should be `call_array` (the function it references below)

PR #1985, opened 2022-12-29
fix: doc typo regarding Multicall::call_array

akramhussein:
## Motivation

Fixes https://github.com/gakonst/ethers-rs/issues/1984 regarding `Multicall::call_array`

## Solution

Change `.call` to `call_array` which is the function the docs are referencing.

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


mattsse (approved):
nice catch

PR #1988, opened 2022-12-30
feat(core, contract): improve `determine_ethers_crates`

DaniPopes:
## Motivation

That weird hack in multicall_contract

## Solution

Improve crate resolution

This is tested by every abigen test and example compiling and testing successfully

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


ethers-core/src/macros/ethers_crate.rs
```diff
@@ -1,123 +1,310 @@
...
+/// The resulting `CARGO_CRATE_NAME` values will be:
+///
+/// |                  Path                  |          Value         |
+/// |:-------------------------------------- | ----------------------:|
+/// | benches/large-input.rs                 |            large-input |
+/// | benches/multi-file-bench/\*\*/\*.rs    |       multi-file-bench |
+/// | examples/simple.rs                     |                 simple |
+/// | examples/multi-file-example/\*\*/\*.rs |     multi-file-example |
+/// | tests/some-integration-tests.rs        | some-integration-tests |
+/// | tests/multi-file-test/\*\*/\*.rs       |        multi-file-test |
```
DaniPopes, line 282: This is not entirely true, but I couldn't really find a good way to get these values..

DaniPopes (commented):
The `multicall_contract` changes can be moved to a separate PR -- should we keep it as an already expanded abigen or use the macro inline (since it works now)?

ethers-core/src/macros/ethers_crate.rs
```diff
@@ -1,123 +1,310 @@
...
+#[inline]
+fn is_crate_root(manifest_dir: impl AsRef<Path>) -> bool {
+    let manifest_dir = manifest_dir.as_ref();
+    env::var_os("CARGO_TARGET_TMPDIR").is_none() &&
+        manifest_dir.components().all(|c| {
+            let s = c.as_os_str();
+            s != "examples" && s != "benches"
+        }) &&
+        !is_crate_name_in_dirs(manifest_dir)
+}
```
mattsse, line 241: this is great!

ethers-core/src/macros/ethers_crate.rs
```diff
@@ -1,123 +1,310 @@
...
+    EthersContractAbigen,
+    EthersContractDerive,
+    EthersCore,
+    EthersDeriveEip712,
+    EthersEtherscan,
+    EthersMiddleware,
+    EthersProviders,
+    EthersSigners,
+    EthersSolc,
+}
```
mattsse, line 132: I like this, this separates str matching from the detection logic, way cleaner

ethers-core/src/macros/ethers_crate.rs
```diff
@@ -1,123 +1,310 @@
...
-/// This process is a bit hacky, we run `cargo metadata` internally which
-/// resolves the current package but creates a new `Cargo.lock` file in the
-/// process. This is not a problem for regular workspaces but becomes an issue
-/// during publishing with `cargo publish` if the project does not ignore
-/// `Cargo.lock` in `.gitignore`, because then cargo can't proceed with
-/// publishing the crate because the created `Cargo.lock` leads to a modified
-/// workspace, not the `CARGO_MANIFEST_DIR` but the workspace `cargo publish`
-/// created in `./target/package/..`. Therefore we check prior to executing
-/// `cargo metadata` if a `Cargo.lock` file exists and delete it afterwards if
-/// it was created by `cargo metadata`.
```
mattsse, line 50: what happened to all the docs? imo these provide additional useful context

ethers-core/src/macros/mod.rs
```diff
@@ -1,2 +1,2 @@
 mod ethers_crate;
-pub use ethers_crate::{ethers_contract_crate, ethers_core_crate, ethers_providers_crate};
+pub use ethers_crate::*;
```
mattsse, line 2: please submit all unrelated clippy fixes etc separately.

mattsse (changes requested):
incredible, keep 'em coming

I think this is way cleaner.

left some nits. 

Now that we have standalone functions for `is_crate_name_in_dirs`, `is_crate_root` I'd like to see some basic tests for them

ethers-core/src/macros/ethers_crate.rs
DaniPopes: first paragraph is split up on the functions, i guess it's missing the explanation for why doing the entire thing (which is that we can have either ethers_core, ethers::core or both)

tbh second paragraph is excessive, deducible from the code (store .exist(), rm after)

DaniPopes:
> Now that we have standalone functions for is_crate_name_in_dirs, is_crate_root I'd like to see some basic tests for them

I don't know, they need a full cargo project to be used properly, with env vars etc, and I don't think the logic is that complicated that it's necessary

ethers-core/src/macros/ethers_crate.rs
gakonst: fine with removing

ethers-core/src/macros/ethers_crate.rs
gakonst: fine w/ me

gakonst (approved):

gakonst:
I am supportive of going the extra mile with some more tests here. Anytime we've refactored this code it's bitten us, esp in `cargo release` with unresolved paths, so would like to be on the safe side. Appreciate the patience / understanding, or we can merge if you feel strongly against it.

DaniPopes:
Refactored by adding an environment struct since I found myself passing the same variables through too many functions, and for easier testing -- should be the same as before

ethers-core/src/macros/ethers_crate.rs
```diff
@@ -14,6 +14,12 @@ use std::{
 };
 use strum::{EnumCount, EnumIter, EnumString, EnumVariantNames, IntoEnumIterator};
 
+#[cfg(test)]
+use rand::{
+    distributions::{Distribution, Standard},
+    Rng,
+};
```
mattsse, line 14: is this needed?

ethers-core/src/macros/ethers_crate.rs
```diff
@@ -143,6 +166,15 @@ impl fmt::Display for EthersCrate {
...
 }
 
+#[cfg(test)]
+impl Distribution<EthersCrate> for Standard {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> EthersCrate {
+        const RANGE: std::ops::Range<u8> = 0..EthersCrate::COUNT as u8;
+        // SAFETY: generates in the safe range
+        unsafe { std::mem::transmute(rng.gen_range(RANGE)) }
+    }
+}
```
mattsse, line 267: can move to test module?

mattsse (changes requested):
thanks for the test.

smol nit re test util

ethers-core/src/macros/ethers_crate.rs
DaniPopes: used in `test_names` -> `gen_unique`

mattsse (approved):

gakonst (approved):
send it - nice work sir

PR #1989, opened 2022-12-30
tests: use sepolia

DaniPopes:
## Motivation

goerli moment

## Solution

add sepolia to test provider and use it

fauceted from  https://web.getlaika.app/faucets

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


ethers-providers/src/lib.rs
```diff
@@ -743,11 +743,10 @@ pub mod test_provider {
...
 
-    pub static GOERLI: Lazy<TestProvider> = Lazy::new(|| TestProvider::new(INFURA_KEYS, "goerli"));
-    pub static MAINNET: Lazy<TestProvider> =
-        Lazy::new(|| TestProvider::new(INFURA_KEYS, "mainnet"));
-    pub static ROPSTEN: Lazy<TestProvider> =
-        Lazy::new(|| TestProvider::new(INFURA_KEYS, "ropsten"));
+    pub static MAINNET: Lazy<TestProvider> = Lazy::new(|| TestProvider::new("mainnet"));
+    pub static GOERLI: Lazy<TestProvider> = Lazy::new(|| TestProvider::new("goerli"));
+    pub static SEPOLIA: Lazy<TestProvider> = Lazy::new(|| TestProvider::new("sepolia"));
+    pub static ROPSTEN: Lazy<TestProvider> = Lazy::new(|| TestProvider::new("ropsten"));
```
mattsse, line 749: let's deprecate ropsten afterwards

.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ name: Tests
 env:
     ETHERSCAN_API_KEY_ETHEREUM: I5BXNZYP5GEDWFINGVEZKYIVU2695NPQZB
     ETHERSCAN_API_KEY_CELO: B13XSMUT6Q3Q4WZ5DNQR8RXDBA2KNTMT4M
-    GOERLI_PRIVATE_KEY: "fa4a1a79e869a96fcb42727f75e3232d6865a82ea675bb95de967a7fe6a773b2"
+    SEPOLIA_PRIVATE_KEY: "fa4a1a79e869a96fcb42727f75e3232d6865a82ea675bb95de967a7fe6a773b2"
```
mattsse, line 15: let's keep goerli as well

we could also switch to gh secrets

mattsse (changes requested):
having a sepolia provider makes total sense.

I think switching testing to sepolia, is probably also the right move for basic contract testing.

.github/workflows/ci.yml
DaniPopes: do you mean we shouldn't change to sepolia at all, or just keep the env var name as goerli?

mattsse: keep both

DaniPopes: problem is that tests are flaky with goerli

.github/workflows/ci.yml
```diff
@@ -1,288 +1,283 @@
 on:
```
gakonst, line 1: giga nit but pls configure your editor to not change the formatting on these files :D or let's do it in a separate PR intentionally

ethers-middleware/tests/nonce_manager.rs
```diff
@@ -44,12 +44,12 @@ async fn nonce_manager() {
     }
 
     // sleep a bit to ensure there's no flakiness in the test
-    std::thread::sleep(std::time::Duration::new(5, 0));
+    tokio::time::sleep(Duration::from_secs(10)).await;
```
gakonst, line 47: why are we changing these from thread sleeps to tokio ones and making them longer?

gakonst (changes requested):
sorry to be nitpicky but let's remove all changes except the addition of the Sepolia provider :D

ethers-providers/src/lib.rs
```diff
@@ -756,16 +758,14 @@ pub mod test_provider {
     }
 
     impl TestProvider {
-        pub fn new(keys: &'static [&'static str], network: &str) -> Self {
-            Self { keys: Mutex::new(keys.iter().cycle()), network: network.to_owned() }
+        pub fn new(network: impl Into<String>) -> Self {
```
gakonst, line 761: i'd like to keep this around so i can specify providers w/ non infura keys in the future

.github/workflows/ci.yml
DaniPopes: idk why it formats on search and replace, that was unintentional

ethers-middleware/tests/nonce_manager.rs
DaniPopes: std sleep blocks the entire tokio runtime or something like that, longer to hopefully make the provider find the txs since it keeps failing in ci

PR #1992, opened 2022-12-30
fix: gwei wei wrong u256 constant

mattsse:
## Motivation
Fixes https://github.com/foundry-rs/foundry/issues/3973

Introduced here https://github.com/gakonst/ethers-rs/pull/1944

## Solution
use correct U256 constant

cc @DaniPopes 

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


DaniPopes:
oops yeah, I thought I made that mistake somewhere but I couldn't find where

PR #2005, opened 2023-01-03
fixed issue#2004 parsing solc verison with trailing newlines

### LINKED ISSUE #2004: ethers-solc version parsing fails with newline suffix
**Version**
ethers-solc v1.0.2

**Platform**
Linux ash11 4.15.0-143-generic #147-Ubuntu SMP Wed Apr 14 16:10:11 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

**Description**
Using `solc` installed from `solc-select` the version cannot be parsed.

When parsing the solc version I get the error: `SemverError(Error("unexpected end of input while parsing major version number"))`

Running `solc --version` I get:
```
solc, the solidity compiler commandline interface
Version: 0.8.12+commit.f00d7308.Linux.g++


```

After some debugging of this code:
https://github.com/gakonst/ethers-rs/blob/master/ethers-solc/src/compile/mod.rs#L684-L697

I found that `solc --version` output includes two trailing newlines:

```
[src/compile/mod.rs:687] &output = Output {
  status: ExitStatus(
    unix_wait_status(
      0,
    ),
  ),
  stdout: "solc, the solidity compiler commandline interface\nVersion: 0.8.12+commit.f00d7308.Linux.g++\n\n",
  stderr: "",
}
[src/compile/mod.rs:689] output.stdout.lines().last() = Some(
  Ok(
    "",
  ),
)
```

The `last()` call evaluates to an empty string causing the version parsing to fail. 

I fixed locally by stripping newlines as follows:
```
fn version_from_output(output: Output) -> Result<Version> {
    if output.status.success() {
        let version =
            output.
            stdout.
            lines().
            **filter(|l| !l.as_ref().unwrap().is_empty()).**
            last()
            .ok_or_else(|| SolcError::solc("version not found in solc output"))?
            .map_err(|err| SolcError::msg(format!("Failed to read output: {err}")))?;
        // NOTE: semver doesn't like `+` in g++ in build metadata which is invalid semver
        //dbg!(&version);
        Ok(Version::from_str(&version.trim_start_matches("Version: ").replace(".g++", ".gcc"))?)
    } else {
        Err(SolcError::solc(String::from_utf8_lossy(&output.stderr).to_string()))
    }
}
```

elizabethdinella:
fixes #2004 

Need some guidance on creating test(s) and updating docs

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-solc/src/compile/mod.rs
+++ b/ethers-solc/src/compile/mod.rs
@@ -686,6 +686,7 @@ fn version_from_output(output: Output) -> Result<Version> {
         let version = output
             .stdout
             .lines()
+            .filter(|l| !l.as_ref().unwrap().is_empty())
             .last()
             .ok_or_else(|| SolcError::solc("version not found in solc output"))?
             .map_err(|err| SolcError::msg(format!("Failed to read output: {err}")))?;
--- a/ethers-solc/src/compile/mod.rs
+++ b/ethers-solc/src/compile/mod.rs
@@ -686,10 +686,10 @@ fn version_from_output(output: Output) -> Result<Version> {
         let version = output
             .stdout
             .lines()
-            .filter(|l| !l.as_ref().unwrap().is_empty())
+            .filter_map(|l| l.ok())
+            .filter(|l| !l.trim().is_empty())
             .last()
-            .ok_or_else(|| SolcError::solc("version not found in solc output"))?
-            .map_err(|err| SolcError::msg(format!("Failed to read output: {err}")))?;
+            .ok_or_else(|| SolcError::solc("version not found in solc output"))?;
         // NOTE: semver doesn't like `+` in g++ in build metadata which is invalid semver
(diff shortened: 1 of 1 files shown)


ethers-solc/src/compile/mod.rs
```diff
@@ -686,6 +686,7 @@ fn version_from_output(output: Output) -> Result<Version> {
         let version = output
             .stdout
             .lines()
+            .filter(|l| !l.as_ref().unwrap().is_empty())
```
mattsse, line 689: great,

now that I'm seeing this again, we can 
```suggestion
            .filter_map(Result::ok)
            .filter(|l| !l.trim().is_empty())
```
and get rid of the `map_err` entirely

mattsse (changes requested):

mattsse (approved):
lgtm, tysm!

gakonst (approved):

PR #2009, opened 2023-01-04
feat(providers): support personal account apis

Rjected:
## Motivation

In reth we are using a clique geth to perform end to end testing. To produce blocks, clique requires that geth manage a clique signing key. Geth needs the raw signing key, and must explicitly "unlock" (decrypt) the key in order to start mining. 

## Solution

Add support for the following `personal` APIs:
 - `personal_importRawKey`
 - `personal_unlockAccount`
 
The `personal` namespace is also enabled in `Geth`. A `insecure_unlock` method is added so we can use `personal_unlockAccount` with open HTTP RPCs.

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


mattsse (approved):
lgtm

Rjected:
rebased

PR #2010, opened 2023-01-04
feat(core): add ability to take geth stderr

Rjected:
## Motivation

The `GethInstance` logs are useful for understanding what geth is doing, and there is currently no way to extract the `pid.stderr` so the logs can be read.

## Solution

Implement a `stderr` method on `GethInstance` that uses `take()` to return a `ChildStderr`, which can then be read with a `BufReader` or other methods.

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/utils/geth.rs
+++ b/ethers-core/src/utils/geth.rs
@@ -5,7 +5,7 @@ use std::{
     fs::{create_dir, File},
     io::{BufRead, BufReader},
     path::PathBuf,
-    process::{Child, Command, Stdio},
+    process::{Child, ChildStderr, Command, Stdio},
     time::{Duration, Instant},
 };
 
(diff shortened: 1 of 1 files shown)


ethers-core/src/utils/geth.rs
```diff
@@ -76,7 +76,17 @@ impl GethInstance {
         &self.data_dir
     }
 
+    /// Takes the stderr contained in the child process.
+    ///
+    /// This leaves a `None` in its place, so calling methods that require a stderr to be present
+    /// will fail if called after this.
+    pub fn stderr(&mut self) -> Result<ChildStderr, GethInstanceError> {
+        self.pid.stderr.take().ok_or(GethInstanceError::NoStderr)
```
mattsse, line 84: could also do an Option but since we already have `NoStdErr`, let's use this result

mattsse (approved):

PR #2011, opened 2023-01-04
docs: mock-provider

0xMelkor:
## Motivation
Docs improvement

## Solution
- [x] Provided examples for the `MockProvider`
- [x] Extended the mdbook 

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/examples/providers/examples/mock.rs
+++ b/examples/providers/examples/mock.rs
@@ -1,7 +1,5 @@
 use ethers::prelude::*;
 
-type BoxErr = Box<dyn std::error::Error>;
-
 /// `MockProvider` is a mock Ethereum provider that can be used for testing purposes.
 /// It allows to simulate Ethereum state and behavior, by explicitly instructing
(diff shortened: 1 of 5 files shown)


examples/providers/examples/mock.rs
```diff
@@ -0,0 +1,82 @@
+use ethers::prelude::*;
+
+type BoxErr = Box<dyn std::error::Error>;
```
gakonst, line 3: let's use `eyre::Error` everywhere and `eyre::Result`

gakonst (changes requested):

examples/providers/examples/mock.rs
0xMelkor: Ok let's keep `eyre::Result` (I wanted to introduce some variations)

PR #2012, opened 2023-01-05
abigen adopt hardhat generated bytecode

zsluedem:
## Motivation

The [hardhat compiled artifacts](https://hardhat.org/hardhat-runner/docs/advanced/artifacts) are like below
```json
{
  "abi":[],
  "bytecode": "0x00000000"
}
```

And the current ethers-rs could only support `{"abi": [..], "bin": "..."}`  or `{"abi": [..], "bytecode": {"object": "..."}}`.

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog
-   [ ] Breaking changes


gakonst (approved):
Am OK with this - @mattsse ?

mattsse (changes requested):
~~while this adds support for hardhat, it no longer supports the nested object format, right?~~

~~instead of removing the `"bytecode": {"object": "..."}` case, we should keep it and just add `bytecode": "0x00000000"` as an additional variant instead.~~

mattsse (approved):
nvm I totally misread that, that's already the case, sorry for the confusion

lgtm

PR #2013, opened 2023-01-05
nonce_manager: remove current nonce load when initialized

gyuho:
## Motivation

If nonce was just initialized, no need for store and load. Just return the stored value without load.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


mattsse (approved):
fair, even though this is the constructor, this gets rid of an atomic load

PR #2015, opened 2023-01-05
fix(abigen): tests

DaniPopes:
## Motivation

Some tests broke due to `ethers` paths now being global (prefixed by `::`)

## Solution

Prefix expected paths in tests with `::`

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


mattsse (approved):
ah, these are assert_quote! tests,

nice catch

Issue #1963, opened 2022-12-21
Support workspace dependencies


DaniPopes:
Cargo supports [inheriting dependencies from the workspace](https://doc.rust-lang.org/stable/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace) since v1.64, which is now our MSRV. We should consider supporting this to pin common dependencies to the same version and features, and make them more maintainable.

I'm aware we have dependabot but it looks like it only updates one crate's manifest and then the root lock file, creating different pinned versions, see examples.

I'm not sure these different versions have any effect when importing the main crate `ethers`, but it might when importing the individual crates, not sure since they might still use the version in the lock file (?)

I'm happy to implement this if it's something we want

## Examples

`syn`:

![syn](https://user-images.githubusercontent.com/57450786/208918201-2b0e6cc2-c153-4916-829e-a6f3a85427f6.png)

`serde_json`:

![serde_json](https://user-images.githubusercontent.com/57450786/208917959-5a66a3df-53a3-466e-9976-615f284dced9.png)

```bash
#!/bin/bash
crates=(ethers-addressbook ethers-contract ethers-contract/ethers-contract-abigen ethers-contract/ethers-contract-derive ethers-core ethers-core/ethers-derive-eip712 ethers-etherscan ethers-middleware ethers-providers ethers-signers ethers-solc)
for c in "${crates[@]}"
do
    res="$(cat "$c/Cargo.toml" | grep "$@")"
    [ "$?" -eq 0 ] && printf "$c:\n$res\n\n"
done
```

PR #2016, opened 2023-01-05
refactor(abigen): source

DaniPopes:
## Motivation

Parse test was not working

Crate was always built with reqwest, tokio even when most of the time it is not used. Ex: build script depending on `ethers-contract-abigen` that only wants to build local files now also gets all of reqwest, and tokio

TODO: upstream this change also in `ethers-contract` -- would be nice if we had #1963.

## Solution

Split into normal and online modules

Reqwest depends on Tokio so there's no extra dependencies

There should be no breaking changes + tests work now

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


ethers-contract/ethers-contract-abigen/Cargo.toml
```diff
@@ -15,23 +15,30 @@ keywords = ["ethereum", "web3", "celo", "ethers"]
 
 [dependencies]
 ethers-core = { version = "^1.0.0", path = "../../ethers-core", features = ["macros"] }
+ethers-etherscan = { path = "../../ethers-etherscan", default-features = false, optional = true }
```
mattsse, line 18: why do we need etherscan here?

mattsse (changes requested):
I'm a bit confused.

does this PR fix a feature issue and also add support for etherscan support to the contract crate?

if so, please split them into separate PRs, otherwise, this is very confusing.

> Parse test was not working

which test?

> Crate was always built with reqwest, tokio even when most of the time it is not used.

which crate? if ethers-contract, this makes sense since reqwest was default?

ethers-contract/ethers-contract-abigen/Cargo.toml
DaniPopes: we have duplicate code for parsing the get request when we could just use the specialized crate in the same project

DaniPopes:
> which test?

There's 1 test for `Source`, `parse_sources`, and it didn't work

> does this PR fix a feature issue and also add support for etherscan support to the contract crate?

> which crate? if ethers-contract, this makes sense since reqwest was default?

`ethers-contract-abigen` fails/panics at runtime in many cases if reqwest is not enabled or is wasm. In my opinion this whole thing is currently not handled well and that is what this PR aims to do. 
I also think that `online` features (with tons of extra dependencies) should not be enabled by default, with offline being an opt in, but the other way around. But this can be another issue/pr

mattsse:
I see, still let's do this in two steps?

DaniPopes:
Not sure how to split this? Since it's not really adding support for Etherscan, it's just moving to using ethers-etherscan instead + refactoring the whole thing in the process

gakonst (approved):
good with me. the ethers_etherscan import is OK not against it

PR #2017, opened 2023-01-05
refactor(abigen): Abigen, ContractBindings

DaniPopes:
## Motivation

See: ea41d210500471ea6f3aac71b06995d8bda3e7b2

Some inefficiencies in parsing

## Solution

impl correct traits

Better docs

deprecate `event_derive` for just `derive`

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


mattsse (changes requested):
great.

marking event_derives as deprecated and rename it to just `derives` is fine by me.

But I'd like to split this PR as well, 
1. one for all the parsing perf improvements
2. all changes related to `event_derives` so we can discuss and changes in 1. are not blocked

DaniPopes:
Alright

PR #2018, opened 2023-01-05
docs(abigen): rename `event_derives` to `derives`

DaniPopes:
## Motivation

`event_derives` implies that the derives are applied only to the smart contract events, which is not true because they get applied to all structs and enums of the contract

## Solution

rename EventDerive, event_derives to Derive and derives respectively (only internal use)
deprecate add_event_derive for add_derive

Add more docs

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


gakonst (changes requested):
Is this a breaking change? can we keep `event_derives` around as an alias that we give a `deprecated` warning for?

DaniPopes:
> Is this a breaking change? can we keep `event_derives` around as an alias that we give a `deprecated` warning for?

the only public items are:
- `add_event_derive` method, which is deprecated and doc(hidden) in favor of `add_derive`
- `event_derive(...)` argument in the abigen! macro, which is kept with added `derive(...)`

Any other items are private so no breaking changes

PR #2021, opened 2023-01-06
fix: disable native-tls on rusoto

rkrasiuk:
Disable default native-tls feature on rusoto

### DIFF
--- a/ethers-signers/Cargo.toml
+++ b/ethers-signers/Cargo.toml
@@ -34,8 +34,8 @@ trezor-client = { version = "0.0.7", optional = true, default-features = false, 
 ] }
 
 # aws
-rusoto_core = { version = "0.48.0", optional = true }
-rusoto_kms = { version = "0.48.0", optional = true }
+rusoto_core = { version = "0.48.0", default-features = false, optional = true }
+rusoto_kms = { version = "0.48.0", default-features = false, optional = true }
 tracing = { version = "0.1.37", optional = true }
 spki = { version = "0.6.0", optional = true }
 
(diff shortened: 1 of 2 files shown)


mattsse (approved):

ethers-signers/Cargo.toml
```diff
@@ -60,5 +60,5 @@ futures = ["futures-util", "futures-executor"]
 celo = ["ethers-core/celo"]
 ledger = ["coins-ledger", "futures", "semver"]
 yubi = ["yubihsm"]
-aws = ["rusoto_core", "rusoto_kms", "tracing", "spki"]
+aws = ["rusoto_core/rustls", "rusoto_kms/rustls", "tracing", "spki"]
```
gakonst, line 63: leave this unchanged

ethers-signers/Cargo.toml
```diff
@@ -34,8 +34,8 @@ trezor-client = { version = "0.0.7", optional = true, default-features = false,
 ] }
 
 # aws
-rusoto_core = { version = "0.48.0", optional = true }
-rusoto_kms = { version = "0.48.0", optional = true }
+rusoto_core = { version = "0.48.0", default-features = false, optional = true }
```
gakonst, line 37:
```suggestion
rusoto_core = { version = "0.48.0", default-features = false, optional = true, features = ["rustls"] }
```

gakonst: should be like this i think?

gakonst (changes requested):
i think the way you did it, it would actually not work

gakonst (approved):
looks like they are the same

PR #2027, opened 2023-01-08
feat(abigen): use prettyplease

### LINKED ISSUE #2026: Format Abigen code using `prettyplease`
Use [dtolnay/prettyplease](https://github.com/dtolnay/prettyplease) internally to format code generated by `Abigen`, instead of `rustfmt`:
- Allows for consistent results across different hosts, toolchains and `rustc` versions
- Doesn't rely on an external binary and all the hurdles that come with invoking one
- Virtually no compilation time overhead as it depends only on `syn` and `proc-macro2`
- [20 times the speed](https://github.com/dtolnay/prettyplease/blob/master/README.md#feature-matrix)
- Used by [cargo-expand](https://github.com/dtolnay/cargo-expand/), [prost](https://github.com/tokio-rs/prost), [yew](https://github.com/yewstack/yew) to name a few
- Made by dtolnay :pray:

More info in the [prettyplease README](https://github.com/dtolnay/prettyplease/blob/master/README.md):

> This library is designed using the simplest possible algorithm and data structures that can deliver about 95% of the quality of rustfmt-formatted output. In my experience testing real-world code, approximately 97-98% of output lines come out identical between rustfmt's formatting and this crate's. The rest have slightly different linebreak decisions, but still clearly follow the dominant modern Rust style.

> The tradeoffs made by this crate are a good fit for generated code that you will not spend your career staring at. For example, the output of `bindgen`, or the output of `cargo-expand`. In those cases it's more important that the whole thing be formattable without the formatter giving up, than that it be flawless.

DaniPopes:
## Motivation

Closes #2026

## Solution

Deprecate and rename `rustfmt` to `format`, no breaking changes

Users who have generated bindings in the project, and run `cargo fmt --check` for example in CI, might have to add a [rustfmt ignore config](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#ignore) (nightly only) to ignore the generated files because the formatting might not match the local rustfmt config.

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


ethers-contract/ethers-contract-abigen/src/lib.rs
```diff
@@ -147,14 +146,20 @@ impl Abigen {
...
-    /// Note that in case `rustfmt` does not exist or produces an error, the
-    /// unformatted code will be used.
     #[must_use]
+    #[deprecated = "Use format instead"]
+    #[doc(hidden)]
     pub fn rustfmt(mut self, rustfmt: bool) -> Self {
-        self.rustfmt = rustfmt;
+        self.format = rustfmt;
+        self
+    }
```
mattsse, line 155: let's just keep this instead of adding/renaming to `format`

mattsse (changes requested):
ty, smol nit

ethers-contract/ethers-contract-abigen/src/lib.rs
DaniPopes: are you sure? I think it can be misleading because someone might think that it's using the same rustfmt that they use in CI or editor but then the formatting differs

https://github.com/tokio-rs/prost/blob/14ac71ef3300d49d04392e7963d6afc2eaf6f2c4/prost-build/src/lib.rs#L871-L878

gakonst: let's go with Dani's suggestion, agreed

gakonst (approved):

PR #2029, opened 2023-01-08
etherscan: fix deserializing contract creation NormalTransaction objects

petertdavies:
## Motivation

The `"to"` field of the response to `get_transactions()` is set to `""` for contract creation transactions. This wasn't handled resulting in a deserialization failure.

## Solution

Added `#[json_string]` macro.


mattsse (approved):
nice catch

gakonst:
nice - thx

PR #2036, opened 2023-01-10
fix(abigen): ensure correct ABI in `From` impl

nickmyatt:
## Motivation

The `new` methods generated by `abigen!` create an underlying `Contract` with the correct ABI, but the generated `From` impls simply wraps a source `Contract`, which could have a completely different ABI. In effect this was an unsafe cast, and indeed was observed to trigger a "method not found (this should never happen)" panic for subsequent method lookups.

## Solution

Implementing `From` in terms of `new` fixes this, at the cost of an extra `Arc::clone()` (which I can't see how to eliminate without piercing the `ethers::contract::Contract` public API).

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/contract.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract.rs
@@ -150,7 +150,7 @@ impl Context {
                     /// client at the given `Address`. The contract derefs to a `ethers::Contract`
                     /// object
                     pub fn new<T: Into<#ethers_core::types::Address>>(address: T, client: ::std::sync::Arc<M>) -> Self {
-                        #ethers_contract::Contract::new(address.into(), #abi_name.clone(), client).into()
+                        Self(#ethers_contract::Contract::new(address.into(), #abi_name.clone(), client))
                     }
 
                     #deployment_methods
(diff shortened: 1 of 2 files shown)


ethers-contract/ethers-contract-abigen/src/contract.rs
```diff
@@ -150,7 +150,7 @@ impl Context {
                     /// client at the given `Address`. The contract derefs to a `ethers::Contract`
                     /// object
                     pub fn new<T: Into<#ethers_core::types::Address>>(address: T, client: ::std::sync::Arc<M>) -> Self {
-                        #ethers_contract::Contract::new(address.into(), #abi_name.clone(), client).into()
+                        Self(#ethers_contract::Contract::new(address.into(), #abi_name.clone(), client))
```
mattsse, line 153: being more explicit is good here

mattsse (approved):
lgtm

ethers-contract/tests/it/abigen.rs
```diff
@@ -750,3 +750,21 @@ fn can_handle_overloaded_function_with_array() {
     ]"#,
     );
 }
+
+#[tokio::test]
+async fn convert_uses_correct_abi() {
```
DaniPopes, line 755: doesn't need to be async
```suggestion
#[test]
fn convert_uses_correct_abi() {
```

DaniPopes (approved):

nickmyatt:
Can I help with anything further? No documentation needed and this is not a breaking change AFAICT. Should I add something to the changelog?

gakonst (approved):
nice catch - thx for the patience

PR #2037, opened 2023-01-10
feat(signers): Allow parsing of private key that has `0x` prefix

oblique:
## Motivation

I wanted to be able to use private keys that have `0x` prefix, without stripping it manually.

## Solution

Remove prefix in `Wallet<SigningKey>::from_str`.

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [x] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-signers/src/wallet/private_key.rs
+++ b/ethers-signers/src/wallet/private_key.rs
@@ -130,9 +130,26 @@ impl FromStr for Wallet<SigningKey> {
     type Err = WalletError;
 
     fn from_str(src: &str) -> Result<Self, Self::Err> {
+        let src = src.strip_prefix("0x").or_else(|| src.strip_prefix("0X")).unwrap_or(src);
         let src = hex::decode(src)?;
         let sk = SigningKey::from_bytes(&src)?;
         Ok(sk.into())
+    }
+}
+
+impl TryFrom<&str> for Wallet<SigningKey> {
+    type Error = WalletError;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        value.parse()
+    }
+}
+
+impl TryFrom<String> for Wallet<SigningKey> {
+    type Error = WalletError;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        value.parse()
     }
 }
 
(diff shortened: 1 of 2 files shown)


ethers-providers/src/transports/common.rs
```diff
@@ -191,7 +192,7 @@ pub enum Authorization {
 
 impl Authorization {
     pub fn basic(username: impl Into<String>, password: impl Into<String>) -> Self {
-        let auth_secret = base64::encode(username.into() + ":" + &password.into());
+        let auth_secret = BASE64_STANDARD.encode(username.into() + ":" + &password.into());
```
mattsse, line 195: this looks like a separate change that bumps base64

mind submitting this separately?

ethers-signers/src/wallet/private_key.rs
```diff
@@ -130,12 +130,29 @@ impl FromStr for Wallet<SigningKey> {
...
+    }
+}
+
+impl TryFrom<String> for Wallet<SigningKey> {
+    type Error = WalletError;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        value.parse()
+    }
+}
```
mattsse, line 154: there's no transient impl when there's `FromStr`?

I guess we can add these

mattsse (changes requested):
smol nit

ethers-providers/src/transports/common.rs
DaniPopes: see #2032

ethers-signers/src/wallet/private_key.rs
oblique: The reason I added this is to enable user to have something like this:

```rust
fn create_client(url: &str, priv_key: impl TryInto<LocalWallet>) -> SignerMiddleware<Provider<Http>, LocalWallet>> {
    let wallet = priv_key.try_into().unwrap();
    // ...
}
```

And be able to pass as argument `&str`, `String`, `SigningKey`, and `SecretKey<Secp256k1>`.

mattsse (approved):

gakonst (approved):

PR #2043, opened 2023-01-11
feat(core): add `strum::EnumIter` and `strum::EnumCount` to `Chain`

joshieDo:
## Motivation

For situations where we want to iterate through all `Chain` variants.

## Solution

We are already using `strum`, so just adding `EnumIter` & `EnumCount` is enough.

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-providers/src/transports/common.rs
+++ b/ethers-providers/src/transports/common.rs
@@ -1,5 +1,6 @@
 // Code adapted from: https://github.com/althea-net/guac_rs/tree/master/web3/src/jsonrpc
 
+use base64::Engine;
 use ethers_core::types::U256;
 use serde::{
     de::{self, MapAccess, Unexpected, Visitor},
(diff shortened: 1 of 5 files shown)


ethers-providers/src/transports/common.rs
```diff
@@ -191,7 +192,8 @@ pub enum Authorization {
 
 impl Authorization {
     pub fn basic(username: impl Into<String>, password: impl Into<String>) -> Self {
-        let auth_secret = base64::encode(username.into() + ":" + &password.into());
+        let auth_secret = base64::engine::general_purpose::STANDARD
+            .encode(username.into() + ":" + &password.into());
```
DaniPopes, line 196: pls see #2032

DaniPopes (approved):

Rjected (approved):
makes sense, yeah #2032 can probably be merged first

gakonst (approved):

PR #2044, opened 2023-01-11
feat(contract): return multicall pending transaction

DaniPopes:
## Motivation

2 year-old TODO comment from #107

```txt
TODO: Can we make this return a PendingTransaction directly instead?
Seems hard due to `returns a value referencing data owned by the current function`
```

The only data that `PendingTransaction` has to hold a reference to is `M::Provider`. which we can get with `Arc::as_ref` but this is currently not exposed

## Solution

add `client_ref` method that returns `Arc::as_ref(self.client)` for `Contract`
return `PendingTransaction` in `Multicall::send`

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [x] Breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -290,6 +290,8 @@
 
 ### Unreleased
 
+- Return pending transaction from `Multicall::send`
+  [#2044](https://github.com/gakonst/ethers-rs/pull/2044)
 - Add abigen to default features
   [#1684](https://github.com/gakonst/ethers-rs/pull/1684)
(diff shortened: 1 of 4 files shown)


mattsse (changes requested):
lgtm, but this is a breaking change and needs to be added to the changelog

I don't have strong feelings about this, perhaps a separate function that awaits the pedingtx would be useful?

gakonst (approved):

PR #2046, opened 2023-01-11
perf(utils): avoid unnecessary allocations

DaniPopes:
## Motivation

`Into<Bytes>` allocates to the heap in `get_create2_address[_from_hash]`

## Solution

Use `AsRef<[u8]>`, minor breaking change only if the type was manually specified wrapper struct / function

Push/Extend a `Vec<u8>` instead of `[...].concat()` / `format!("...")`

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog
-   [x] Breaking changes


ethers-core/src/utils/mod.rs
```diff
@@ -367,18 +362,24 @@ pub fn get_create2_address(
...
+    bytes.extend_from_slice(from.as_bytes());
+    bytes.extend_from_slice(salt);
+    bytes.extend_from_slice(init_code_hash);
 
     let hash = keccak256(bytes);
 
     let mut bytes = [0u8; 20];
     bytes.copy_from_slice(&hash[12..]);
-    Address::from(bytes)
+    H160(bytes)
```
mattsse, line 382: let's stick to `Address` please

ethers-core/src/utils/mod.rs
```diff
@@ -344,19 +340,18 @@ pub fn get_create2_address(
...
 ///     Token::Address(token1),
-///     Token::Uint(U256::from(fee)),
+///     Token::Uint(fee),
 /// ]);
 ///
 /// // keccak256(abi.encode(token0, token1, fee))
 /// let salt = keccak256(&input);
-/// let pool_address =
-///     get_create2_address_from_hash(factory, salt.to_vec(), UNISWAP_V3_POOL_INIT_CODE_HASH);
+/// let pool_address = get_create2_address_from_hash(factory, salt, init_code_hash);
```
mattsse, line 354: unrelated?

mattsse (approved):
ty, lgtm

suggesting re `H160` v `Address`.

this isn't actually breaking, right?

ethers-core/src/utils/mod.rs
DaniPopes: doc also doing unnecessary allocations :)

mattsse: fair enough ;)

gakonst (approved):

PR #2048, opened 2023-01-12
fix: revert to old version

Will-Smith11:
@gakonst  version 0.5.8 of num_enum was pulled so now all builds on the most recent version fail due to an invalid dep


Will-Smith11:
@gakonst can we merge this pls

Will-Smith11:
image

gakonst:
ACK sorry was AFK

Issue #1895, opened 2022-11-27
Cargo test Failed in ethers-solc


aewc:
**Version**
I pull the ethers-rs, and using the Commit b0ef1343dd71c28bdce74212aa136e827b9c54a8.

**Platform**

Darwin C27C.local 22.1.0 Darwin Kernel Version 22.1.0: Sun Oct  9 20:14:30 PDT 2022; root:xnu-8792.41.9~2/RELEASE_ARM64_T8103 arm64

**Description**
```sh
$ cd ethers-rs
$ ETHERSCAN_API_KEY=xxxxx GOERLI_PRIVATE_KEY=xxxxx cargo test
...
failures:

---- config::tests::can_autodetect_dirs stdout ----
thread 'config::tests::can_autodetect_dirs' panicked at 'assertion failed: `(left == right)`
  left: `"/private/var/folders/2y/70kmz0xn3j95ncb5fw0j737w0000gn/T/rootPom69r/artifacts/build-info"`,
 right: `"/var/folders/2y/70kmz0xn3j95ncb5fw0j737w0000gn/T/rootPom69r/artifacts/build-info"`', ethers-solc/src/config.rs:934:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    config::tests::can_autodetect_dirs

test result: FAILED. 73 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.74s

error: test failed, to rerun pass `-p ethers-solc --lib`
```
The path difference in `tests::can_autodetect_dirs` is only `/private`

mattsse:
ah this is a mac symlink issue

path paths point to the same file, but the way `std::fs::canonicalize` works on mac is that `/private/var/`  is resolved to `/var`, or the other way around, can't remember.

this could be fixed by used `std::fs::canonicalize` on the other variant in the assertEq I think.

Could you please try that?

aewc:
Yes, macOS have this weird linked dir: `std::fs::canonicalize` would add `/private` in `/var`

However, it has not been resolved. The root cause is that the `build-info` folder has not been created, resulting in
```rs
pub fn canonicalized(path: impl Into<PathBuf>) -> PathBuf {
     let path = path.into();
     canonicalize(&path).unwrap_or(path)
}
```
`unwrap_or` directly returns `path`.

I printed some information, as follows:
```sh
[...]

PR #2052, opened 2023-01-13
fix: failing can_autodetect_dirs solc test (#1895)

agostbiro:
## Motivation

Fix `can_autodetect_dirs` solc test failing on MacOS (#1895).

## Solution

The test was failing for two reasons on macOS:

1. The build info directory was not created during the test.
2. Temporary directories are created with symbolic links on macOS and the path canonicalization returns the given path if it doesn't exist.

This lead to a mismatch between the project root's canonical path (which did exist and thus was a real path after canonicalization) and the build info directory's canonical path (which did not exist and was a symbolic link after canonicalization).

The fix is to create the build info directory to make sure its canonical path matches the project root's canonical path.

## PR Checklist

-   [ ] Added Tests **(n/a)**
-   [ ] Added Documentation **(n/a)**
-   [ ] Updated the changelog **(n/a)**
-   [ ] Breaking changes **(n/a)**

### DIFF
--- a/ethers-solc/src/config.rs
+++ b/ethers-solc/src/config.rs
@@ -911,13 +911,13 @@ mod tests {
 
         let root = root.path();
         assert_eq!(ProjectPathsConfig::find_source_dir(root), src,);
-        std::fs::File::create(&contracts).unwrap();
+        std::fs::create_dir_all(&contracts).unwrap();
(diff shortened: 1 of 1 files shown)


mattsse (approved):
oh nice!

Issue #2053, opened 2023-01-14
derive EthAbiType would cause clippy::unwrap-used lint error


zsluedem:
**Version**
`ethers = { git = "https://github.com/gakonst/ethers-rs.git", rev = "3ed83d5dd38e7819605511631aaddcc3e472427e", features = ["ethers-solc"]  }`

**Platform**
Linux will-MS-7B89 5.15.0-58-generic #64-Ubuntu SMP Thu Jan 5 11:43:13 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

**Description**

I tried this code:

```rust
  #[derive(Clone, EthAbiType)]
  pub struct UserOperation {
      pub sender: Address,
  }
```

cargo clippy --  -D clippy::unwrap_used

Lint error would show
```
error: used `unwrap()` on `an Option` value
    |
  |         pub sender: Address,
    |         ^^^
    |
    = help: if you don't want to handle the `None` case gracefully, consider using `expect()` to provide a better panic message
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_used
    = note: requested on the command line with `-D clippy::unwrap-used`

error: could not compile `aa-bundler` due to previous error

```

mattsse:
It'd be great if you could attach the expanded code

zsluedem:
@mattsse I created a pr #2056  for it

```rust 
#![feature(prelude_import)]
#[prelude_import]
use std::prelude::rust_2021::*;
#[macro_use]
extern crate std;
use ethers::prelude::EthAbiType;
use ethers::types::{Address, Bytes};
pub struct UserOperation {
    pub sender: Address,
    pub paymaster_and_data: Bytes,
    pub signature: Bytes,
}
#[automatically_derived]
impl ::core::clone::Clone for UserOperation {
    #[inline]
    fn clone(&self) -> UserOperation {
        UserOperation {
            sender: ::core::clone::Clone::clone(&self.sender),
            paymaster_and_data: ::core::clone::Clone::clone(&self.paymaster_and_data),
            signature: ::core::clone::Clone::clone(&self.signature),
        }
    }
}
impl ::ethers::core::abi::AbiType for UserOperation {
    fn param_type() -> ::ethers::core::abi::ParamType {
        ::ethers::core::abi::ParamType::Tuple(<[_]>::into_vec(
            #[rustc_box]
            ::alloc::boxed::Box::new([
[...]

PR #2056, opened 2023-01-15
remove EthAbiType derive generated unwrap

zsluedem:
## Motivation

https://github.com/gakonst/ethers-rs/issues/2053
## Solution
Replace unwrap with safer method

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-contract/ethers-contract-derive/src/abi_ty.rs
+++ b/ethers-contract/ethers-contract-derive/src/abi_ty.rs
@@ -39,14 +39,7 @@ pub fn derive_tokenizeable_impl(input: &DeriveInput) -> proc_macro2::TokenStream
 
                 let assignments = fields.named.iter().map(|f| {
                     let name = f.ident.as_ref().expect("Named fields have names");
-                    let name_str = name.to_string();
-                    quote_spanned! { f.span() => #name: iter
-                        .next()
-                        .ok_or_else(|| #core_crate::abi::InvalidOutputType(::std::format!(
-                            "Struct field {:} doesn't have correspond token.", #name_str
-                        )))
-                        .and_then(|t| #core_crate::abi::Tokenizable::from_token(t))?
-                    }
+                    quote_spanned! { f.span() => #name: #core_crate::abi::Tokenizable::from_token(iter.next().expect("The iter is guaranteed to be something due to the size check"))? }
                 });
                 let init_struct_impl = quote! { Self { #(#assignments,)* } };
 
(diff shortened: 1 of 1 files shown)


ethers-contract/ethers-contract-derive/src/abi_ty.rs
```diff
@@ -39,7 +39,14 @@ pub fn derive_tokenizeable_impl(input: &DeriveInput) -> proc_macro2::TokenStream
...
                     let name = f.ident.as_ref().expect("Named fields have names");
-                    quote_spanned! { f.span() => #name: #core_crate::abi::Tokenizable::from_token(iter.next().unwrap())? }
+                    let name_str = name.to_string();
+                    quote_spanned! { f.span() => #name: iter
+                        .next()
+                        .ok_or_else(|| #core_crate::abi::InvalidOutputType(::std::format!(
+                            "Struct field {:} doesn't have correspond token.", #name_str
+                        )))
+                        .and_then(|t| #core_crate::abi::Tokenizable::from_token(t))?
+                    }
```
mattsse, line 49: I think we should just use an `expect` here, because there's already a len check so this is guaranteed to succeed, hence the unwrap.

mattsse (changes requested):

ethers-contract/ethers-contract-derive/src/abi_ty.rs
zsluedem: Although I don't have problems with `expect`, there is also https://rust-lang.github.io/rust-clippy/master/#expect_used lint kind which some people might want to warn about it.

mattsse: this error can never happen though

zsluedem: I know.Just want to pass the lint check for unwrap in my situation. If you want `expect`, then `expect` it is.

ethers-contract/ethers-contract-derive/src/abi_ty.rs
```diff
@@ -39,14 +39,7 @@ pub fn derive_tokenizeable_impl(input: &DeriveInput) -> proc_macro2::TokenStream
...
                     let name = f.ident.as_ref().expect("Named fields have names");
-                    let name_str = name.to_string();
-                    quote_spanned! { f.span() => #name: iter
-                        .next()
-                        .ok_or_else(|| #core_crate::abi::InvalidOutputType(::std::format!(
-                            "Struct field {:} doesn't have correspond token.", #name_str
-                        )))
-                        .and_then(|t| #core_crate::abi::Tokenizable::from_token(t))?
-                    }
+                    quote_spanned! { f.span() => #name: #core_crate::abi::Tokenizable::from_token(iter.next().expect("No token found"))? }
```
mattsse, line 42: the expect message should make it clear that this is guaranteed due the size check.

PR #2064, opened 2023-01-19
feat: add support for js tracer to geth trace, fix different return types

Vid201:
## Motivation

Geth debug tracing API accepts custom JavaScript tracers.

## Solution

Added support for the additional tracer.

Tracing API returns different things based on the used tracer type.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


ethers-core/src/types/trace/geth.rs
```diff
@@ -39,16 +40,66 @@ pub struct StructLog {
...
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CallFrame {
+    #[serde(rename = "type")]
+    pub typ: String,
+    pub from: Address,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub to: Option<NameOrAddress>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub value: Option<U256>,
+    #[serde(deserialize_with = "from_int_or_hex")]
```
mattsse, line 53: if these are skipped if they are none, does that mean that they could be missing in the json object geth serializes?

In that case all these options should also have the `serde(default,` attribute.

ethers-core/src/types/trace/geth.rs
```diff
@@ -39,16 +40,66 @@ pub struct StructLog {
...
+pub enum GethDebugBuiltInTracerType {
     #[serde(rename = "callTracer")]
     CallTracer,
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Deserialize, Serialize)]
+#[serde(untagged)]
+/// Available tracers
+///
+/// See <https://geth.ethereum.org/docs/developers/evm-tracing/built-in-tracers> and <https://geth.ethereum.org/docs/developers/evm-tracing/custom-tracer>
```
mattsse, line 94: style nit: flip derive and comment order.

ethers-core/src/types/trace/geth.rs
```diff
@@ -39,16 +40,66 @@ pub struct StructLog {
...
+    pub from: Address,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub to: Option<NameOrAddress>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub value: Option<U256>,
+    #[serde(deserialize_with = "from_int_or_hex")]
+    pub gas: U256,
+    #[serde(deserialize_with = "from_int_or_hex", rename = "gasUsed")]
+    pub gas_used: U256,
+    #[serde(serialize_with = "serialize_bytes")]
```
mattsse, line 57: Bytes already serializes as hex string, so this seems redundant?

ethers-core/src/types/trace/geth.rs
```diff
@@ -39,16 +40,66 @@ pub struct StructLog {
...
+    pub input: Bytes,
+    #[serde(skip_serializing_if = "Option::is_none", serialize_with = "serialize_bytes_opt")]
```
mattsse, line 59: same

ethers-core/src/types/trace/geth.rs
```diff
@@ -87,3 +138,14 @@ where
...
+fn serialize_bytes_opt<S, T>(x: &Option<T>, s: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+    T: AsRef<[u8]>,
+{
+    match x {
+        Some(x) => serialize_bytes(x, s),
+        None => s.serialize_none(),
+    }
+}
```
mattsse, line 151: this and the existing `serialize_bytes` function seem unnecessary because `Bytes` already does that I think

or do they differ? if so needs docs.

mattsse (changes requested):
ty, only a few nits

Vid201:
> ty, only a few nits

@mattsse Made the changes, thanks for the review

gakonst (approved):
lgtm

ethers-core/src/types/trace/geth.rs
```diff
@@ -21,34 +22,83 @@ pub struct GethTrace {
...
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub storage: Option<BTreeMap<H256, H256>>,
 }
 
+// https://github.com/ethereum/go-ethereum/blob/a9ef135e2dd53682d106c6a2aede9187026cc1de/eth/tracers/native/call.go#L37
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CallFrame {
+    #[serde(rename = "type")]
+    pub typ: String,
```
KuTuGu, line 47: @Vid201   Why use typ instead of type ;)

Vid201: @KuTuGu AFAIK `type` is a keyword in Rust and cannot be used for variable/property name.

PR #2067, opened 2023-01-20
feat(etherscan): Allow `ClientBuilder` to create `Client` without API key

oblique:
## Motivation

Etherscan allows 1 request per 5 seconds if the API key is not defined

## Solution

Make API key optional

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


mattsse (approved):
reasonable

PR #2068, opened 2023-01-21
fix(core): fix geth --init temp dir race condition

Rjected:
## Motivation

Previously, if multiple instances of geth were spawned concurrently, only one directory would be used to populate geth's `genesis.json`. This can lead to a race condition where the `genesis.json` would be re-written by a second instance, before the first instance reads the `genesis.json` for populating its db and genesis block with the `geth --init` command. this was possible because directory returned by `std::env::temp_dir()` is often shared.

## Solution

This fixes the race condition by using `tempfile::tempdir()`, which creates a unique directory per call to `tempdir()`.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/Cargo.toml
+++ b/ethers-core/Cargo.toml
@@ -34,6 +34,7 @@ hex = { version = "0.4.3", default-features = false, features = ["std"] }
 once_cell = { version = "1.17.0", optional = true }
 unicode-xid = "0.2.4"
 strum = { version = "0.24", features = ["derive"] }
+tempfile = { version = "3.3.0", default-features = false }
 
 # macros feature enabled dependencies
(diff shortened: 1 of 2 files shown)


mattsse (approved):
for some reason this caused the wasm job to fail, 
`unused dependency`, otherwise lgtm, nice catch

Rjected:
just added:
```toml
[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
tempfile = { version = "3.3.0", default-features = false }
```
to fix the unused dependency

gakonst (approved):

PR #2070, opened 2023-01-23
Keystore Generation & Decryption

crypdoughdoteth:
## Motivation

This is a functioning example of generating + encrypting a new keystore and decrypting a keystore using ethers-rs. This could particularly be helpful for applications that rely on local encrypted keys client side or server side.

### DIFF
--- a/examples/wallets/examples/keystores.rs
+++ b/examples/wallets/examples/keystores.rs
@@ -25,7 +25,7 @@ async fn main() -> Result<()>{
     //generates a brand new keystore with key
     //name your keystore file, set the path, set the password
     //*NOTE* -- pls don't store your passwords in a plain text file, this is just an example 
-    let key_store = LocalWallet::new_keystore(encryption_path, &mut thread_rng(), &password, Some("my_encrypted_keys"))
+    let (signing_key, _) = LocalWallet::new_keystore(encryption_path, &mut thread_rng(), &password, Some("my_encrypted_keys"))
         .expect("key store fail");
     
     let (your_signing_key, _) = key_store; 
--- a/examples/wallets/examples/keystores.rs
+++ b/examples/wallets/examples/keystores.rs
@@ -1,22 +1,12 @@
 use ethers::prelude::*;
 use ethers::core::rand::thread_rng;
 use ethers::signers::{LocalWallet};
-use dotenv::dotenv;
-
 use eyre::Result;
-
 
 #[tokio::main]
 async fn main() -> Result<()>{
 
-    dotenv().ok();
-
-    let api_key = std::env::var("API_KEY").expect("expected environmental variable");
-    let encryption_path = std::env::var("ENCRYPTION_PATH").expect("expected environmental variable");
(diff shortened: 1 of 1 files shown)


examples/wallets/examples/keystores.rs
```diff
@@ -0,0 +1,43 @@
...
+use eyre::Result;
+
+
+#[tokio::main]
+async fn main() -> Result<()>{
+
+    dotenv().ok();
+
+    let api_key = std::env::var("API_KEY").expect("expected environmental variable");
+    let encryption_path = std::env::var("ENCYPTION_PATH").expect("expected environmental variable");
```
roynalnaruto, line 15:
```suggestion
    let encryption_path = std::env::var("ENCRYPTION_PATH").expect("expected environmental variable");
```

examples/wallets/examples/keystores.rs
```diff
@@ -0,0 +1,43 @@
...
+    let ws = Ws::connect(api_key)
+        .await
+        .expect("no connection");
+
+    println!("Is Connected: {}", ws.ready());
+
+    //generates a brand new keystore with key
+    //name your keystore file, set the path, set the password
+    //*NOTE* -- pls don't store your passwords in a plain text file, this is just an example 
+    let key_store = LocalWallet::new_keystore(encryption_path, &mut thread_rng(), &password, Some("my_encrypted_keys"))
```
roynalnaruto, line 28:
```suggestion
    let (signing_key, _) = LocalWallet::new_keystore(encryption_path, &mut thread_rng(), &password, Some("my_encrypted_keys"))
```

examples/wallets/examples/keystores.rs
```diff
@@ -0,0 +1,43 @@
...
+        .expect("key store fail");
+    
+    let (your_signing_key, _) = key_store;
```
roynalnaruto, line 31: Then you can remove this line

roynalnaruto (commented):
Minor suggestions and nitpicks

examples/wallets/examples/keystores.rs
crypdoughdoteth: Thank you! I merged your suggestions ❤️

roynalnaruto: You might need to fix linting

crypdoughdoteth: Linting worked locally, I think the bot doesn't like dotenv crate. So, I removed it and put placeholder values in

roynalnaruto:
Actually for what's being done in this example, in my opinion the documentation above those methods should already be sufficient. But I believe @gakonst can decide if this example is a good to have!

gakonst (changes requested):
Just looking at the docs.rs https://docs.rs/ethers/latest/ethers/signers/struct.Wallet.html#method.decrypt_keystore, I would rather that we had these here, instead of in the example. 

This example also isn't buildingm so my recommendation would be to move these to a doc comment in the struct/function:
1. Please add one of the public RPC urls we use
2. Please use a path that's not windows specific

prestwich:
Closing as stale

PR #2072, opened 2023-01-24
fix(provider): Add Send bound to return type of JsonRpcClient::request

wigy-opensource-developer:
## Motivation

`#[async_trait]` implementations require return types to be `Send`. So except for trivial cases the `JsonRpcClient::request` demanded implementors to be impossibly generic.

## Solution

Add `Send` bound to the method and follow-up on its callers inside the monorepo.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-contract/tests/it/abigen.rs
+++ b/ethers-contract/tests/it/abigen.rs
@@ -752,6 +752,7 @@ fn can_handle_overloaded_function_with_array() {
 }
 
 #[test]
+#[allow(clippy::disallowed_names)]
 fn convert_uses_correct_abi() {
     abigen!(
         Foo, r#"[function foo()]"#;
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -76,7 +76,7 @@ pub trait JsonRpcClient: Debug + Send + Sync {
     async fn request<T, R>(&self, method: &str, params: T) -> Result<R, Self::Error>
     where
         T: Debug + Serialize + Send + Sync,
-        R: DeserializeOwned;
+        R: DeserializeOwned + Send;
 }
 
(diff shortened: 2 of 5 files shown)


gakonst:
> #[async_trait] implementations require return types to be Send. So except for trivial cases the JsonRpcClient::request demanded implementors to be impossibly generic.

Got it - could you please add a test to avoid regressions? cc @mattsse is this too restrictive?

wigy-opensource-developer:
Do you mean a test that breaks if someone removes this Send bound? I will, but they would probably remove the test as well then 😃

gakonst:
> Do you mean a test that breaks the build if someone removes this Send bound? I will, but they would probably remove the test as well then 😃

A test that would only compile if the Send bound is required!

mattsse (approved):
I think this is reasonable since we don't expect that the response type is not Send because it's also DeserializeOwned

gakonst:
Hm I guess that makes sense? `DeserializeOwned` is owned, so it's probably going to be `Send`? Surprised they haven't done that themselves on the trait level? Are there any cases when that's not the case? Probably w/ `unsafe`?

mattsse:
unsure, there's the `rc` feature which allows deserializing directly into an `Rc` which would no longer work after this PR, but imo this is perfectly fine.
I don't we there are any responses that are deserialized into a type !Send fields like Cell,Rc,etc.. so this should be okay imo. 

RawValue is Send

https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html

gakonst:
SGTM - let's send it and keep an eye open.

PR #2076, opened 2023-01-25
fix: add getrandom with js feature for wasm

mattsse:
## Motivation
getrandom needs to be enabled with js feature when for wasm target

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


gakonst (approved):
oof how was this not caught before -_-

PR #2081, opened 2023-01-27
feat(abi): add helper functions to access solidity types

mattsse:
## Motivation
Adds necessary functions to fix https://github.com/foundry-rs/foundry/issues/4039

all current getter functions return the rust type (pascal case) but we want the actual solidity type as well so we can use it to fix https://github.com/foundry-rs/foundry/issues/4039

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


gakonst (approved):
makes sense

PR #2082, opened 2023-01-27
refactor: make contract abstract over Borrow

prestwich:
## Motivation

Contracts currently hold `Arc<M>` and have bound `where M: Middleware` . This requires managing arcs across a codebase, which imposes unpleasant requirements on binaries

## Solution

- Modify the contract object to be abstract over `B: Borrow<M>` instead of holding an `Arc<M>`. This allows contracts to be used directly with Cows, non-ARCed middleware, Boxes, other smart pointers, and anything else that uses the `Borrow`  trait.
- rename the modified object `ContractInternal` and create a type alias under the old name, for backwards compatibility

Example use case: 
- you can now use a contract directly with a `Lazy<Provider<Http>>` without having to have an arc in the lazy first

## Inference Limitation

While `FunctionCall` and `ContractInstance` may be used with any type is `Borrow<M>`, there are sometimes multiple choices of borrowed type that satisfy the `M: Middleware` constraint. 

because `Middleware` has an `auto_impl(&, Box, Arc)`, references, boxes, and arcs have multiple `Borrow` options that satisfy the constraint. 
E.g. `Arc<M>` has these impls:
- `impl Borrow<Arc<M>> for Arc<M>`
- `impl Borrow<M> for Arc<M>`
- `impl Borrow<&M> for Arc<M>`

as a result, type inference fails when calling `ContractInstance::new(address, arc_mware)` users must instead specify `ContractInstance::<_, ConcreteMiddlewareType>(address, arc_mware)`

Removing the `auto_impl` will solve this. However, we don't know the extent of reliance on that auto_impl. As a result, using the backwards-compatibility aliases `Contract` and `ContractCall` is still recommended. the Followup work section describes a migration route to fix inference and deprecate the backwards-compatibility aliases

## Followup work

- Change the abigen derived Contract objects to use the generic `ContractInternal` 
- in 2.0.0, eliminate the type alias, and have the  generic version use the `Contract` name

Lib-wide refactoring plan:
1. list further abstraction targets
    - EthEvent
    - ContractDeployer
    - ContractFactory
    - Middlewares (should the stack be `MyMiddleware { inner: Borrow<M> } ` ?)
    - ProviderOracle
    - abigen! output code
3. refactor to use borrow
4. provide backwards compatible type aliases like `type Contract = ContractInstance<Arc<M>, M>`
5. remove `auto_impl(&, Box, Arc)` on `Middleware`
6. write migration guide
7. deprecate backwards-compatibility aliases
8. remove aliases

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog
-   [x] Breaking changes

### DIFF
--- a/ethers-contract/src/base.rs
+++ b/ethers-contract/src/base.rs
@@ -1,4 +1,4 @@
-use crate::Contract;
+use crate::contract::ContractInternal;
 
 pub use ethers_core::abi::AbiError;
 use ethers_core::{
@@ -8,10 +8,10 @@ use ethers_core::{
 use ethers_providers::Middleware;
 
 use std::{
+    borrow::Borrow,
     collections::{BTreeMap, HashMap},
     fmt::Debug,
     hash::Hash,
-    sync::Arc,
 };
 
 /// A reduced form of `Contract` which just takes the `abi` and produces
@@ -195,12 +195,12 @@ impl BaseContract {
     }
 
     /// Upgrades a `BaseContract` into a full fledged contract with an address and middleware.
-    pub fn into_contract<M: Middleware>(
-        self,
-        address: Address,
-        client: impl Into<Arc<M>>,
-    ) -> Contract<M> {
-        Contract::new(address, self, client.into())
+    pub fn into_contract<B, M>(self, address: Address, client: B) -> ContractInternal<B, M>
+    where
+        B: Borrow<M>,
+        M: Middleware,
+    {
+        ContractInternal::new(address, self, client)
     }
 }
 
--- a/ethers-contract/src/base.rs
+++ b/ethers-contract/src/base.rs
@@ -200,7 +200,7 @@ impl BaseContract {
         address: Address,
         client: impl Into<Arc<M>>,
     ) -> Contract<M> {
-        Contract::new(address, self, client)
+        Contract::new(address, self, client.into())
     }
 }
 
--- a/ethers-contract/src/base.rs
+++ b/ethers-contract/src/base.rs
@@ -1,4 +1,4 @@
-use crate::contract::ContractInternal;
+use crate::contract::ContractInstance;
 
 pub use ethers_core::abi::AbiError;
 use ethers_core::{
@@ -195,12 +195,12 @@ impl BaseContract {
     }
 
     /// Upgrades a `BaseContract` into a full fledged contract with an address and middleware.
-    pub fn into_contract<B, M>(self, address: Address, client: B) -> ContractInternal<B, M>
+    pub fn into_contract<B, M>(self, address: Address, client: B) -> ContractInstance<B, M>
     where
         B: Borrow<M>,
         M: Middleware,
     {
-        ContractInternal::new(address, self, client)
+        ContractInstance::new(address, self, client)
     }
 }
 
--- a/ethers-contract/src/call.rs
+++ b/ethers-contract/src/call.rs
@@ -255,7 +255,7 @@ where
 impl<B, M, D> IntoFuture for FunctionCall<B, M, D>
 where
     Self: 'static,
-    B: Borrow<M>,
+    B: Borrow<M> + Send + Sync,
     M: Middleware,
     D: Detokenize + Send + Sync,
 {
--- a/ethers-contract/src/call.rs
+++ b/ethers-contract/src/call.rs
@@ -15,12 +15,11 @@ use ethers_providers::{
 };
 
 use std::{
-    borrow::Cow,
+    borrow::{Borrow, Cow},
     fmt::Debug,
     future::{Future, IntoFuture},
     marker::PhantomData,
     pin::Pin,
-    sync::Arc,
 };
 
 use thiserror::Error as ThisError;
@@ -73,33 +72,45 @@ pub enum ContractError<M: Middleware> {
     ContractNotDeployed,
 }
 
+/// type alias for backwards compatibility
+pub type ContractCall<M, D> = ContractCallInternal<std::sync::Arc<M>, M, D>;
+
 #[derive(Debug)]
 #[must_use = "contract calls do nothing unless you `send` or `call` them"]
 /// Helper for managing a transaction before submitting it to a node
-pub struct ContractCall<M, D> {
+pub struct ContractCallInternal<B, M, D> {
     /// The raw transaction object
     pub tx: TypedTransaction,
     /// The ABI of the function being called
     pub function: Function,
     /// Optional block number to be used when calculating the transaction's gas and nonce
     pub block: Option<BlockId>,
-    pub(crate) client: Arc<M>,
+    pub(crate) client: B,
     pub(crate) datatype: PhantomData<D>,
+    pub(crate) _m: PhantomData<M>,
 }
 
-impl<M, D> Clone for ContractCall<M, D> {
+impl<B, M, D> Clone for ContractCallInternal<B, M, D>
+where
+    B: Clone,
+{
     fn clone(&self) -> Self {
-        ContractCall {
+        ContractCallInternal {
             tx: self.tx.clone(),
             function: self.function.clone(),
             block: self.block,
             client: self.client.clone(),
             datatype: self.datatype,
+            _m: self._m,
         }
     }
 }
 
-impl<M, D: Detokenize> ContractCall<M, D> {
+impl<B, M, D> ContractCallInternal<B, M, D>
+where
+    B: Borrow<M>,
+    D: Detokenize,
+{
     /// Sets the `from` field in the transaction to the provided value
     pub fn from<T: Into<Address>>(mut self, from: T) -> Self {
         self.tx.set_from(from.into());
@@ -145,8 +156,9 @@ impl<M, D: Detokenize> ContractCall<M, D> {
     }
 }
 
-impl<M, D> ContractCall<M, D>
+impl<B, M, D> ContractCallInternal<B, M, D>
 where
+    B: Borrow<M>,
     M: Middleware,
     D: Detokenize,
 {
@@ -157,7 +169,11 @@ where
 
     /// Returns the estimated gas cost for the underlying transaction to be executed
     pub async fn estimate_gas(&self) -> Result<U256, ContractError<M>> {
-        self.client.estimate_gas(&self.tx, self.block).await.map_err(ContractError::MiddlewareError)
+        self.client
+            .borrow()
+            .estimate_gas(&self.tx, self.block)
+            .await
+            .map_err(ContractError::MiddlewareError)
     }
 
     /// Queries the blockchain via an `eth_call` for the provided transaction.
@@ -170,8 +186,12 @@ where
     ///
     /// Note: this function _does not_ send a transaction from your account
     pub async fn call(&self) -> Result<D, ContractError<M>> {
-        let bytes =
-            self.client.call(&self.tx, self.block).await.map_err(ContractError::MiddlewareError)?;
+        let bytes = self
+            .client
+            .borrow()
+            .call(&self.tx, self.block)
+            .await
+            .map_err(ContractError::MiddlewareError)?;
 
         // decode output
         let data = decode_function_data(&self.function, &bytes, false)?;
@@ -202,7 +222,7 @@ where
     ///
     /// Note: this function _does not_ send a transaction from your account
     pub fn call_raw_bytes(&self) -> CallBuilder<'_, M::Provider> {
-        let call = self.client.provider().call_raw(&self.tx);
+        let call = self.client.borrow().provider().call_raw(&self.tx);
         if let Some(block) = self.block {
             call.block(block)
         } else {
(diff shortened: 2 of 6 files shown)


ethers-contract/src/lib.rs
```diff
@@ -5,6 +5,9 @@
 mod contract;
 pub use contract::Contract;
 
+// pub mod call2;
+// pub mod contract2;
```
prestwich, line 9: artifact of development. need to delete.

ethers-contract/src/contract.rs
```diff
@@ -150,32 +150,45 @@ use ethers_core::types::TransactionRequest;
...
+pub struct ContractInternal<B, M> {
     address: Address,
     base_contract: BaseContract,
-    client: Arc<M>,
+    client: B,
+    _m: PhantomData<M>,
 }
 
-impl<M> std::ops::Deref for Contract<M> {
+pub type Contract<M> = ContractInternal<std::sync::Arc<M>, M>;
```
prestwich, line 160: type aliases ensure backwards compatibility

mattsse (commented):
this makes it easier indeed, 

especially for http provider, which actually could be made `Clone`

ethers-contract/src/contract.rs
```diff
@@ -291,12 +336,13 @@ impl<M: Middleware> Contract<M> {
 
         let tx = tx.into();
 
-        Ok(ContractCall {
+        Ok(ContractCallInternal {
             tx,
-            client: Arc::clone(&self.client), // cheap clone behind the Arc
+            client: self.client.clone(),
```
prestwich, line 341: we could make `ContractCallInternal` borrow our B here, by introducing a lifetime to it

I am concerned about the practicality of the `B: Clone` bound, for borrow types _without_ clone. are there commonly used borrow types that are _NOT_ clone? 🤔

ethers-contract/src/base.rs
```diff
@@ -200,7 +200,7 @@ impl BaseContract {
         address: Address,
         client: impl Into<Arc<M>>,
     ) -> Contract<M> {
-        Contract::new(address, self, client)
+        Contract::new(address, self, client.into())
```
prestwich, line 203: should this be generic over B? I think so. I'll change it in a minute

ethers-contract/src/contract.rs
```diff
@@ -209,14 +229,14 @@ impl<M: Middleware> Contract<M> {
...
+    M: Middleware,
+{
     /// Creates a new contract from the provided client, abi and address
-    pub fn new(
-        address: impl Into<Address>,
-        abi: impl Into<BaseContract>,
-        client: impl Into<Arc<M>>,
-    ) -> Self {
-        Self { base_contract: abi.into(), client: client.into(), address: address.into() }
+    pub fn new(address: impl Into<Address>, abi: impl Into<BaseContract>, client: B) -> Self {
```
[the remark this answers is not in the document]
mattsse: this change makes more sense, 
because `impl Into Arc` is kinda dumb, especially if you need already have an `Arc<P>`

prestwich:
currently checking out the test breakage. looks like this will affect type inference

ethers-contract/src/contract.rs
DaniPopes: It can be the case when `B: Borrow<M>` is equal to `M` and `M` is not Clone, since `Borrow` has a blanket implemention for any `T: ?Sized`. Otherwise it would mostly be the standard smart pointers. So I'd say if all our `Middleware` implement `Clone` this is fine

prestwich:
```
error[E0283]: type annotations needed
   --> ethers-contract/tests/it/contract.rs:431:14
    |
431 |             .connect(client2.clone())
    |              ^^^^^^^ cannot infer type of the type parameter `N` declared on the associated function `connect`
    |
    = note: multiple `impl`s satisfying `Arc<ethers_providers::Provider<Http>>: Borrow<_>` found in the following crates: `alloc`, `core`:
            - impl<T> Borrow<T> for Arc<T>
              where T: ?Sized;
            - impl<T> Borrow<T> for T
              where T: ?Sized;
note: required by a bound in `ethers_contract::contract::ContractInternal::<B, M>::connect`
   --> /Users/james/devel/ethers-rs/ethers-contract/src/contract.rs:269:12
    |
269 |         C: Borrow<N>,
    |            ^^^^^^^^^ required by this bound in `ethers_contract::contract::ContractInternal::<B, M>::connect`
help: consider specifying the type arguments in the function call
    |
431 |             .connect::<C, N>(client2.clone())
    |                     ++++++++
```

the test failures are all driven by failed inference on `connect`. because `connect` is now generic over `B`, and `B: Borrow<M>` and `M: Borrow<M>`, both B and M are valid args for `connect()`. the compiler doesn't know whether it should be using the `B` or borrowing `M` from it and using `M`

this is only an issue when _switching_ the middleware connected to a contract, and the solution is to write out `connect::<Arc<_>, _>()`. I think this is acceptable (although it is a breaking change)

prestwich:
workaround:
- changed params to `connect` to use an `arc` to leave existing code un-broken
- added `connect_with` with the borrow api, which requires more type ascription

ethers-contract/src/contract.rs
prestwich: unfortunately, not all of our middleware implement Clone, IIRC. e.g. I think `SignerMiddleware<LedgerSigner, _>` does not 🤔

DaniPopes: Ah yeah right, I guess this will need to be specified in a doc somewhere that it is recommended wrapping `M` with a smart pointer or using references to avoid cloning data or for cases where that it is actually not possible

prestwich: I'm going to make a couple toy examples for when `B` is `M` or `&M`. (`&M` is `Clone` however, so that should be fine)

prestwich:
anyone have a better name than `ContractInternal` ? 🤔

ethers-contract/src/contract.rs
prestwich: Screenshot 2023-01-27 at 12 19 12 PM

prestwich: I've also confirmed that `NonceManagerMiddleware` is not Clone

prestwich: so the place where I user would run into this is:

- instantiating a contract with an _owned_ **non-clone** middleware

prestwich: slight adjustment to screenshot. 3rd example is wrong, as M is `&`. corrected version here:

Screenshot 2023-01-27 at 12 27 58 PM

ethers-contract/src/call.rs
```diff
@@ -73,33 +72,45 @@ pub enum ContractError<M: Middleware> {
...
 }
 
+/// type alias for backwards compatibility
+pub type ContractCall<M, D> = ContractCallInternal<std::sync::Arc<M>, M, D>;
+
 #[derive(Debug)]
 #[must_use = "contract calls do nothing unless you `send` or `call` them"]
 /// Helper for managing a transaction before submitting it to a node
-pub struct ContractCall<M, D> {
+pub struct ContractCallInternal<B, M, D> {
```
mattsse, line 81: What is the motivation for this renaming?
so that we can use `ContractCall` for backwards compat?

The "internal" postfix somewhat irritates me as well.
I guess, technically this is a `ContractFunctionCall` or `FunctionCall`?

ethers-contract/src/contract.rs
```diff
@@ -145,37 +145,73 @@ use ethers_core::types::TransactionRequest;
...
+    _m: PhantomData<M>,
 }
 
-impl<M> std::ops::Deref for Contract<M> {
+/// `Contract` is a [`ContractInternal`] object with an `Arc` middleware.
+/// This type alias exists to preserver backwards compatibility with
+/// non-abstract Contracts.
+///
+/// For full usage docs, see [`ContractInternal`]
+pub type Contract<M> = ContractInternal<std::sync::Arc<M>, M>;
```
mattsse, line 183: nit: type aliases at the beginning of the file/ above struct def, as for `ContractCall`

mattsse, line 183: should we perhaps mark the aliases as deprecated?

ethers-contract/src/contract.rs
```diff
@@ -145,37 +145,73 @@ use ethers_core::types::TransactionRequest;
...
+///
+/// We expect to fix this fully in a future version
+///
 /// [`abigen`]: macro.abigen.html
 /// [`Abigen` builder]: struct.Abigen.html
 /// [`event`]: method@crate::Contract::event
 /// [`method`]: method@crate::Contract::method
 #[derive(Debug)]
-pub struct Contract<M> {
+pub struct ContractInternal<B, M> {
```
mattsse, line 171: The same concerns re naming here.
`Internal` is likely confusing, but for this one, I don't have any suggestions at this point.

hmm, compared to `BaseContract` this type represents an `Existing`, `Deployed` contract since it has an `address`, so perhaps we include this in the name?

mattsse (changes requested):
I like this change, 
using `Borrow` should allow us to create a `Contract` that uses a reference to a provider (`&M`) right? do we have an example for this.

so it's possible to 
1. connect ws
2. create and consume `Contract` with `&ws`
3. return ownership of `ws`

so "Arc<Ws>" should no longer be mandatory for every "Contract" interaction, right?

ethers-contract/src/contract.rs
prestwich: Naming for `Contract` is inherited from ethers.js, and implies an interface with an address and a provider. I believe `BaseContract` was backfilled to make sense given `Contract` has a connection to the chain.

proposed better nomenclature:

- `Interface` -- currently called `BaseContract` provides encoding/decoding functionality
   - alias to `BaseContract` for backwards compat
   - maybe de-duplicate the hashmap via a global cache so we stop doing a bunch of allocs on clone?
- `ContractInstance` -- currently `ContractInternal`. Wraps `BaseContract` with an address and a connection
  - alias version with arc to `Contract` for backwards compat

ethers-contract/src/call.rs
prestwich: `FunctionCall` seems like a good name, as contract calls may not have functions (e.g. fallback or receive invocations)

ethers-contract/src/contract.rs
prestwich: adding a deprecated tag results in close to 200 warnings in the ethers codebase. To avoid noise in the PR, I am adding in a commented-out deprecated attribute. I'll uncomment it and change all of our internal usage as a last step.

ethers-contract/src/contract.rs
```diff
@@ -240,26 +289,50 @@ impl<M: Middleware> Contract<M> {
...
-        name: &str,
-        args: T,
-    ) -> Result<ContractCall<M, D>, AbiError> {
-        // get the function
-        let function = self.base_contract.abi.function(name)?;
-        self.method_func(function, args)
+    /// Returns a new contract instance using the provided client
+    ///
+    /// Clones `self` internally
+    pub fn connect<N>(&self, client: Arc<N>) -> ContractInstance<Arc<N>, N>
```
gakonst, line 295: Should this also be `#[must_use]`?

ethers-contract/src/contract.rs
gakonst: Is that a big problem? Can't we pass these Middlewares by reference, instead of owned? And otherwise, just Arc them?

prestwich: > Is that a big problem? Can't we pass these Middlewares by reference, instead of owned? And otherwise, just Arc them?

it's a not-immediately-obivous issue that users will run into and ask about. It compounds with the ambiguity caused by `auto_imp(&, Box, Arc)` to make this more-powerful API extremely difficult to use

if you call `ContractInstance::new(address, &provider)` you get a type ambiguity around `M`. is `M` `Provider` or `&Provider`. Both implement `Middleware` and both can be borrowed from `&Provider`. Same ambiguity for `Arc`

The follow up work (will copy this to original post) to get this usable by beginner rustaceans is something like:
```
1. list further abstraction targets
2. refactor to use borrow
3. provide backwards compatible type aliases like `type Contract = ContractInstance<Arc<M>, M>`
4. remove `auto_impl(&, Box, Arc)` on `Middleware`
5. write migration guide
6. deprecate backwards-compatibility aliases
7. remove aliases
```

ethers-contract/src/contract.rs
prestwich: update:

see top comment for future plans wrt this. Aliases will not be deprecated in this PR. We will deprecate after we have a straightforward and beginner-friendly migration plan

prestwich:
@gakonst @mattsse @DaniPopes 

Please see updated notes at top of comment thread. I believe this is ready for final review and merge 🎉

ethers-contract/src/call.rs
```diff
@@ -73,33 +72,49 @@ pub enum ContractError<M: Middleware> {
     ContractNotDeployed,
 }
 
+/// `ContractCall` is a [`FunctionCall`] object with an [`Arc`] middleware.
```
gakonst, line 75:
```suggestion
/// `ContractCall` is a [`FunctionCall`] object with an [`std::sync::Arc`] middleware.
```

gakonst (approved):
Supportive - pending @mattsse

DaniPopes (approved):

mattsse (approved):
I think this is step in the right direction.

I guess there could be some quirks wrt Arc and its auto_impls but we can figure things out when we encounter them.

PR #2085, opened 2023-01-27
fix: add serde(default) to v,r,s in `Transaction`

gakonst:
ref: https://github.com/foundry-rs/foundry/issues/4094

I am not sure if this is a really good change, but also think that the other ways to do it (eg with an Option) are too annoying.

cc @mattsse


mattsse (commented):
I have a feeling that this could cause some issues, especially with recovery etc

I wonder if converting this into an enum would be more appropriate?

gakonst:
Yeah real solution is to go itno `enum Transaction { Ethereum(..), Celo(..), Optimism }` I think. Let's do that?

PR #2088, opened 2023-01-29
Add Geth js tracer option

zsluedem:
## Motivation
In [aa-bundler](https://github.com/Vid201/aa-bundler), we need `debug_traceCall` javascript tracer option. So we add this js tracer config in making the request.

And the `CallTracer` return data is not correct which needs to be fixed, too.

## Solution

Add additional options.

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-providers/src/lib.rs
+++ b/ethers-providers/src/lib.rs
@@ -595,12 +595,15 @@ pub trait Middleware: Sync + Send + Debug {
     }
 
     /// Executes the given call and returns a number of possible traces for it
-    async fn debug_trace_call<T: Into<TypedTransaction> + Send + Sync>(
+    async fn debug_trace_call<
+        T: Into<TypedTransaction> + Send + Sync,
+        R: Serialize + DeserializeOwned + Debug + Send,
+    >(
         &self,
         req: T,
         block: Option<BlockId>,
         trace_options: GethDebugTracingCallOptions,
-    ) -> Result<GethTrace, ProviderError> {
+    ) -> Result<R, ProviderError> {
         self.inner().debug_trace_call(req, block, trace_options).await.map_err(FromErr::from)
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Unreleased
 
+- Add support for Geth javascript custom tracer request into `debug_traceCall` [#2088](https://github.com/gakonst/ethers-rs/pull/2088)
 - Add a `Send` bound to the `IntoFuture` implementation of `ContractCall` [#2083](https://github.com/gakonst/ethers-rs/pull/2083)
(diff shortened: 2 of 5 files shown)


ethers-providers/src/lib.rs
```diff
@@ -595,12 +595,15 @@ pub trait Middleware: Sync + Send + Debug {
     }
 
     /// Executes the given call and returns a number of possible traces for it
-    async fn debug_trace_call<T: Into<TypedTransaction> + Send + Sync>(
+    async fn debug_trace_call<
+        T: Into<TypedTransaction> + Send + Sync,
+        R: Serialize + DeserializeOwned + Debug + Send,
+    >(
```
mattsse, line 601: I'm not sure about this because this would allow deserializing into any type.

what types are possible here?

perhaps we can add some safety trait restrictions here. For example
introduce a trait that is 

```rust
trait CallTrace: DeserializeOwned + Debug + Send {}

impl CallTrace for /*All types we can return in `debug_trace_call`*/ {}

```

mattsse (changes requested):
this seems to add a `CallTracer`, similar to https://github.com/gakonst/ethers-rs/pull/2064
and modifies the `debug_trace_call` so it is generic and can be deserialized into a `DeserializeOwned`

I'd suggest we go with #2064 for the model changes and tackle the provider function in this PR?

ethers-providers/src/lib.rs
zsluedem: > I'm not sure about this because this would allow deserializing into any type.

> what types are possible here?

Here are some tricky things with JS tracer. For JS tracer, it could return any type depending on the JS. Look at the example here https://github.com/gakonst/ethers-rs/blob/54879a83484da7a196801ee3a734d7d6c5ed8304/examples/geth_trace_call.rs#L60.

Focus on the `result: function() { return {\"result\":\"HelloWorld\"};` here. For this JS tracer, it would return something like 
```rust
struct Custom{
  result: String
}
```

And you could define some other results like `result: function() { return {\"a\":\"HelloWorld\", \"b\": 123};`  then the return type would be 
```rust
struct Custom{
  a: String,
  b: u64
}
```


> perhaps we can add some safety trait restrictions here.

That's a good idea.

zsluedem:
> I'd suggest we go with https://github.com/gakonst/ethers-rs/pull/2064 for the model changes and tackle the provider function in this PR?
@mattsse Agree. Then I would just wait for #2064 to merge first.

ethers-providers/src/lib.rs
mattsse: I see, so adding another trait bound is likely unnecessary.

gakonst:
2064 is merged!

zsluedem:
Closing this since #2064 is adopted

PR #2089, opened 2023-01-29
refactor/feat(abigen,types): add bytes::Bytes static methods, refactor struct declaration

DaniPopes:
## Motivation

we can make abigen'd contracts' bytecode static slices instead of allocating it at runtime with Lazy, using bytes::Bytes::from_static

## Solution

see commit list;

technically a breaking change since `#bytecode_name` is not `Lazy<Bytes>` anymore but just `Bytes`, but the interaction should be the exact same since `Lazy` derefs into `Bytes`

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [x] Breaking changes


mattsse (approved):
good idea.
lgtm

gakonst (approved):

PR #2095, opened 2023-01-30
feat(core): more type parsing

DaniPopes:
## Motivation

Upstreaming from foundry-cli/cast helper parsers

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/types/block.rs
+++ b/ethers-core/src/types/block.rs
@@ -675,6 +675,7 @@ impl fmt::Display for BlockNumber {
 #[cfg(not(feature = "celo"))]
 mod tests {
     use super::*;
+    use crate::types::{Transaction, TxHash};
 
     #[test]
     fn can_parse_eip1898_block_ids() {
@@ -921,6 +922,7 @@ mod tests {
 #[cfg(feature = "celo")]
 mod celo_tests {
     use super::*;
+    use crate::types::Transaction;
 
     #[test]
     fn block_without_snark_data() {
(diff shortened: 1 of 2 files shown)


ethers-core/src/types/ens.rs
```diff
@@ -100,10 +100,14 @@ impl From<Address> for NameOrAddress {
...
+    type Err = <Address as FromStr>::Err;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        Ok(Self::Name(s.to_string()))
+        if s.starts_with("0x") {
+            s.parse().map(Self::Address)
+        } else {
+            Ok(Self::Name(s.to_string()))
+        }
     }
```
mattsse, line 111: this is a great change

ethers-core/src/types/block.rs
```diff
@@ -624,15 +639,22 @@ impl FromStr for BlockNumber {
...
+            "latest" => Ok(Self::Latest),
+            "finalized" => Ok(Self::Finalized),
+            "safe" => Ok(Self::Safe),
+            "earliest" => Ok(Self::Earliest),
+            "pending" => Ok(Self::Pending),
+            n => {
+                if let Ok(n) = n.parse::<U64>() {
+                    Ok(Self::Number(n))
+                } else if let Ok(n) = n.parse::<u64>() {
+                    Ok(Self::Number(n.into()))
```
mattsse, line 652: This is no longer compliant with the rpc spec for block number, but I guess this is fine?

mattsse (approved):
lgtm

ethers-core/src/types/block.rs
gakonst: Fine w/ me. More flexibility, and people create `BlockNumber` themselves, vs always parsing it out from the hex returned from RPC

gakonst (approved):

PR #2104, opened 2023-02-01
feat(abi): packed encoding

### LINKED ISSUE #1862: Add abi::encode_packed
There is currently no solution to have a packed encoding in Rust using ethers, while this is useful for many use cases on Solidity, for example while hashing.

an `abi::encode_packed()`method that takes a list of `Token` and returns the encoding.

Currently, we have to use another library, which define its own types to achieve it https://github.com/roberts-ivanovs/eth-encode-packed-rs/blob/21962ce04aa62f396ac517fcbea91263dfc373a5/src/lib.rs#L103

DaniPopes:
## Motivation

Closes #1862

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/abi/packed.rs
+++ b/ethers-core/src/abi/packed.rs
@@ -1,0 +1,423 @@
+use ethabi::Token;
+use thiserror::Error;
+use Token::*;
+
+/// An error thrown by [`encode_packed`].
+#[derive(Debug, Error)]
+pub enum EncodePackedError {
+    #[error("This token cannot be encoded in packed mode: {0:?}")]
+    InvalidToken(Token),
+
+    #[error("FixedBytes token length > 32")]
+    InvalidBytesLength,
+}
+
+/// Encodes the given tokens into an ABI compliant vector of bytes.
+///
+/// This function uses [non-standard packed mode][ref], where:
+/// - types shorter than 32 bytes are concatenated directly, without padding or sign extension;
+/// - dynamic types are encoded in-place and without the length;
+/// - array elements are padded, but still encoded in-place.
+///
+/// Note that this function has the same behaviour as its [Solidity counterpart][ref], and
+/// thus structs as well as nested arrays are not supported.
+///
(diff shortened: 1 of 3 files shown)


ethers-core/src/abi/packed.rs
```diff
@@ -0,0 +1,423 @@
...
+                out.extend_from_slice(&[0; 12]);
+            }
+            out.extend_from_slice(&addr.0)
+        }
+        Int(n) | Uint(n) => {
+            // TODO: Different (u)int* padding
+            let mut buf = [0; 32];
+            n.to_big_endian(&mut buf);
+            out.extend_from_slice(&buf);
+        }
```
DaniPopes, line 119: By using the same `Token` type as `ethabi`, we cannot differentiate between the different integer type sizes as `Token::(U)Int` stores only the  `U256` number and not its encoded size (since in normal abi-encoding this wouldn't have any effect).
Should this function use a different `Token`-like enum, or is there another way around this?

gakonst: You can find the smallest value for which `n >> BITS` is 0, and mark that as the size?

e.g.
255 >> 8  == 0 --> put it in uint8
whereas 256 >> 8 == 1, putting it in uint16
etc.

DaniPopes: Right, but how does one then cast the number as something higher than its minimum bits? Like uint256(0) would need to pad all 32 bytes? We can have this as a minor limitation for the time being I guess

DaniPopes: Implemented this using `U256::bits`. Any 0 padding needs to be manually added with something like this:

```rust
let n = U256::from(...);
let mut buf = [0; 32];
n.to_big_endian(&mut buf);
// uint32 = 4 bytes -> 32 - 4 .. 32
let uint32 = Token::Bytes(buf[28..32]);
```

gakonst:
LGTM - ty!!

PR #2116, opened 2023-02-04
chore(etherscan): add cloudflare captcha error

mattsse:
## Motivation
new cloudflare error unlocked.

this supports a new cloudflare error variant that occurs when cloudflare asks for a captcha upon an API request.

discovered here https://github.com/foundry-rs/foundry/actions/runs/4091046786/jobs/7054872187

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


DaniPopes:
Can't we early return if the response's status code is bad? or is that also inconsistent

mattsse:
so less specific error variant and just, NotOk?

I don't think this is more useful.

PR #2121, opened 2023-02-07
feat: add support for Geth built-in tracer and config

KuTuGu:
## Motivation
Now only `callTracer` is supported, add more `build-in` tracer and [config](https://geth.ethereum.org/docs/developers/evm-tracing/built-in-tracers#config) support.

## Solution

- feat: support other geth build-in tracers
- feat: support geth tracer config
- refactor: reconstructs the layout of geth tracer
- test: add unit tests for default, legacy, and tracer config

Refs:  #2064

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [x] Updated the changelog
-   [ ] Breaking changes


ethers-core/src/types/trace/geth/call.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub typ: String,
+    pub from: Address,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub to: Option<NameOrAddress>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub value: Option<U256>,
+    #[serde(default, deserialize_with = "from_int_or_hex")]
+    pub gas: U256,
+    #[serde(default, deserialize_with = "from_int_or_hex", rename = "gasUsed")]
+    pub gas_used: U256,
```
KuTuGu, line 20: The `gas` and `gas_used` fields are not compatible with [the legacy format](https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/internal/tracetest/testdata/call_tracer_legacy/simple.json#LL60C5-L60C5). Should we change it to `Option`, and which might be a break change?      
For now, to be compatible with previous versions, I've only added a default handling, which returns `U256::zero()` instead of `None`.

ethers-core/src/types/trace/geth/pre_state.rs
```diff
@@ -0,0 +1,92 @@
...
+// https://github.com/ethereum/go-ethereum/blob/91cb6f863a965481e51d5d9c0e5ccd54796fd967/eth/tracers/native/prestate.go#L38
+#[derive(Debug, PartialEq, Eq, Clone, Deserialize, Serialize)]
+#[serde(untagged)]
+pub enum PreStateFrame {
+    Default(PreStateMode),
+    Diff(DiffMode),
+}
+
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PreStateMode(pub BTreeMap<Address, AccountState>);
```
mattsse, line 17: I think this should probably b `#[serde(transparent)]`?

https://serde.rs/container-attrs.html#transparent

ethers-core/src/types/trace/geth/test_data/call_tracer/default.rs
```diff
@@ -0,0 +1,21 @@
+r#"{
+    "calls": [
+        {
+            "from": "0x3b873a919aa0512d5a0f09e6dcceaa4a6727fafe",
+            "gas": "0x6d05",
+            "gasUsed": "0x0",
+            "input": "0x",
```
mattsse, line 7: I'd prefer them to be the json objects and not json strings embedded in rust file.

mattsse (changes requested):
thanks,

left some nits.

ethers-core/src/types/trace/geth/test_data/call_tracer/default.rs
KuTuGu: get it

ethers-core/src/types/trace/geth/pre_state.rs
KuTuGu: Seems transparent is enabled for newtype by [default](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=226ffebdc45dc392ae006722284b62da)?

mattsse: TIL

mattsse (approved):
lgtm

ethers-core/src/types/trace/geth/call.rs
gakonst: `default` is fine w me

gakonst (approved):

PR #2122, opened 2023-02-07
feature: bubble up jsonrpc error response via trait

prestwich:
## Motivation

Currently there's no way to get JSON-RPC error response out of the provider or middleware error. They're type-erased in the ProviderError.

## Solution

This adds a `ClientError` trait which helps cast `ProviderError` to `JsonRpcError`. 

I consider this to be a bit of an unpleasant kludge. I'm not sure it should be merged. Hypothetically this would allow us to access and decode revert data from contracts tho, which is definitely a big user need.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -236,6 +236,8 @@
 
 ### Unreleased
 
+- Add `MiddlewareError` and `RpcError` traits to support access to deeply nested
+  RPC errors [#2122](https://github.com/gakonst/ethers-rs/pull/2122)
 - Convert provider errors to arbitrary middleware errors
   [#1920](https://github.com/gakonst/ethers-rs/pull/1920)
 - Add a subset of the `admin` namespace
--- a/ethers-contract/tests/it/contract.rs
+++ b/ethers-contract/tests/it/contract.rs
@@ -13,7 +13,9 @@ mod eth_tests {
         utils::{keccak256, Anvil},
     };
     use ethers_derive_eip712::*;
-    use ethers_providers::{Http, Middleware, PendingTransaction, Provider, StreamExt};
+    use ethers_providers::{
+        Http, Middleware, MiddlewareError, PendingTransaction, Provider, StreamExt,
+    };
     use ethers_signers::{LocalWallet, Signer};
     use std::{convert::TryFrom, iter::FromIterator, sync::Arc, time::Duration};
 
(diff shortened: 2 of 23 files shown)


prestwich:
okay this now provides 2 traits:

- `RpcError`, which covers `ProviderError` and all transports
    - exposes `as_error_response` for possible downcast to JsonRpcError 
- `MiddlewareError`  which covers `ProviderError` and up
    - replaces `FromErr` 
    - provides functions for converting to/from Inner
    - provides functions for converting to/from ProviderError
    - provides functions for downcasting to an error response

the core idea here is that `ProviderError` is a hingepoint in the lib. everything below it (except JsonRpcError responses) is implementation details and should be hidden. everything above it is MiddlewareError, and should be stacked like middleware are stacked

if the direction is good, I'll go through and update all the docs

prestwich:
CI failures are unrelated. This is ready for review cc @gakonst @mattsse

gakonst (approved):
Really nice and useful, basically the same pattern that middleware uses to give access to inner layers but for errors, to avoid the nested match abomination... 

(We could/shoudl've done the refactor in a separate PR but OK w/ me)

gakonst:
Main question is: what breaking changes does this have? Seems like..none? If so, let's go ahead

prestwich:
> (We could/shoudl've done the refactor in a separate PR but OK w/ me)

I can break these into 2 if you want, shoulda done that in the first place tbh

> Main question is: what breaking changes does this have? Seems like..none? If so, let's go ahead

the breaking changes are

- 3rd-party custom transports need to implement `RpcError` for their errors
- 3rd-party custom middleware need to change `FromErr` impls to `MiddlewareError` impls

gakonst:
> 3rd-party custom transports need to implement RpcError for their errors
> 3rd-party custom middleware need to change FromErr impls to MiddlewareError impls

Sounds good. We'll need to update `ethers-flashbots` and `ethers-fireblocks` but otherwise should be good. Let's send it.

prestwich:
re-organization has been moved into #2159

prestwich:
I went ahead and split, as it won't generate conflicts because that branch is based on this one 👍 

I can also easily clean up any conflicts so lfg

gakonst (approved):

mattsse (approved):
lgtm, does this affect how the RetryProvider needs to check for serde/reqwest error?

prestwich:
> lgtm, does this affect how the RetryProvider needs to check for serde/reqwest error?

it really doesn't affect it at all. But now that you mention it, we could add `as_serde_error() -> Option<&serde_json::Error>` 🤔

prestwich:
okay I added `as_serde_error()` and `is_*_error()` and a bunch of docs for the traits :)

prestwich:
failures are unrelated and this is good to go 🎉 

like other recent PRs, this is a breaking change, and will require action from 3rd-party crates

gakonst:
supreseded https://github.com/gakonst/ethers-rs/pull/2159

PR #2123, opened 2023-02-08
feat(ethers-solc): extend model checker setting field

wiasliaw:
## Motivation

- extend model checker setting's field to open `invariants` flag.
- [invariants docs](https://docs.soliditylang.org/en/v0.8.18/smtchecker.html#reported-inferred-inductive-invariants) 

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -1018,6 +1018,8 @@ pub struct ModelCheckerSettings {
     pub timeout: Option<u32>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub targets: Option<Vec<ModelCheckerTarget>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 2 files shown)


mattsse (changes requested):
when was this introduced?

we probably need a check here

https://github.com/gakonst/ethers-rs/blob/e20bb715699280402c155a0abbfc82656582242e/ethers-solc/src/artifacts/mod.rs#L103-L105

wiasliaw:
@mattsse 
Found in [0.8.10](https://blog.soliditylang.org/2021/11/09/solidity-0.8.10-release-announcement/).

mattsse (approved):

gakonst (approved):
nice - i have not yet seen any prod codebases which use these extensively, would love to have some examples in foundry

Issue #2129, opened 2023-02-08
Pascal case not applied to struct fields in Abigen


batconjurer:
**Version**
This error occurred with `ethers-contract` version `1.0.2`.

**Description**
When generating Rust types for a solidity contract, a solidity struct with the name `ERC20Transfer` led to the creation of a Rust struct called `Erc20Transfer`, because Abigen applies Pascal case formatting to struct names. 

However, suppose another solidity struct existed 
```
struct SomeStruct {
    ERC20Transfer someField;
}
``` 

Then the genereated Abigen code will produce a Rust struct like so:
```rust
struct SomeStruct {
     some_field: ERC20Transfer,
}
```
The obvious fix is to apply Pascal casing to field types when expanding struct definitions. change https://github.com/gakonst/ethers-rs/blob/e20bb715699280402c155a0abbfc82656582242e/ethers-contract/ethers-contract-abigen/src/contract/structs.rs#L599
 
to
```rust
 let ty = util::ident(&ty.name().to_pascal_case());
``` 

Running this on my solidity contracts produced the desired result.

mattsse:
too summarize:
* struct types are created like `Erc20Transfer` but when used as fields it is `ERC20Transfer`?

In that case, I believe this is the correct fix.

batconjurer:
That is correct.

mattsse:
ah yes, nice find.

we use pascal case for the rust type here

https://github.com/gakonst/ethers-rs/blob/e20bb715699280402c155a0abbfc82656582242e/ethers-contract/ethers-contract-abigen/src/contract/structs.rs#L303

Would you like to submit the fix?

batconjurer:
Sure, I have already forked the repo and made the change. 

However, as per the contributing document, I am finding that with running `cargo check --all-features` that the versions of the crates `thiserror` and `anyhow` are using unstable features. Should I attempt to fix this as well, or leave it?

PR #2130, opened 2023-02-09
[fix]: Use pascal casing for type names in structs generated by abigen

batconjurer:
This fixes Issue #2129. Additionally, the versions of crates `thiserror` and `anyhow` where downgraded a couple of minor versions so that `cargo check --all-features` passes. The versions in the lock file used unstable features.

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/contract/structs.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract/structs.rs
@@ -596,7 +596,7 @@ fn struct_type_projections(name: &str) -> Vec<String> {
 fn expand_struct_type(struct_ty: &StructFieldType) -> TokenStream {
     match struct_ty {
         StructFieldType::Type(ty) => {
-            let ty = util::ident(ty.name());
+            let ty = util::ident(&ty.name().to_pascal_case());
             quote! {#ty}
         }
         StructFieldType::Array(ty) => {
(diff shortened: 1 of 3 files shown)


ethers-contract/ethers-contract-abigen/src/contract/structs.rs
```diff
@@ -596,7 +596,7 @@ fn struct_type_projections(name: &str) -> Vec<String> {
 fn expand_struct_type(struct_ty: &StructFieldType) -> TokenStream {
     match struct_ty {
         StructFieldType::Type(ty) => {
-            let ty = util::ident(ty.name());
+            let ty = util::ident(&ty.name().to_pascal_case());
```
mattsse, line 599: lgtm

ethers-signers/Cargo.toml
```diff
@@ -16,7 +16,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 ethers-core = { version = "^1.0.0", path = "../ethers-core", features = ["eip712"] }
-thiserror = { version = "1.0.38", default-features = false }
+thiserror = { version = "1.0.36", default-features = false }
```
mattsse, line 19: this doesn't actually change the locked version. 
So I don't understand what exactly this fixes.

mattsse (commented):
the fix lgtm, unsure about the effects of the thiserror downgrade.

perhaps submit this separately?

ethers-signers/Cargo.toml
batconjurer: I temporarily used `version = "=1.0.36"` to change the lock. Once the lock file was updated, I removed the `=` sign since it struck me as quite restrictive.

batconjurer:
I have removed the changes to the lock file.

gakonst:
ty!

PR #2137, opened 2023-02-10
feat(examples): add basic event filtering example

bsh98:
## Motivation

Improve the `events` section of `examples` by adding a simple filtering example.

First-time contributor and also new to Rust, so any feedback is greatly appreciated!  I'm happy to add more examples, as I find them very helpful for familiarizing others with the project as well as being great exercises for learning Rust.

## Solution

Adds a basic event filtering example that also includes parsing log topics.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [x] Updated the changelog
-   [ ] Breaking changes


gakonst (approved):
nice! we should also expose these examples + contracts in the book

PR #2140, opened 2023-02-13
Start Geth with debug api endpoint 

zsluedem:
## Solution
Introduce additional config to enable debug api endpoint.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/utils/geth.rs
+++ b/ethers-core/src/utils/geth.rs
@@ -21,7 +21,7 @@ const GETH_STARTUP_TIMEOUT_MILLIS: u64 = 10_000;
 const GETH_DIAL_LOOP_TIMEOUT: Duration = Duration::new(20, 0);
 
 /// The exposed APIs
-const API: &str = "eth,net,web3,txpool,admin,personal,miner";
+const API: &str = "eth,net,web3,txpool,admin,personal,miner,debug";
 
 /// The geth command
 const GETH: &str = "geth";


ethers-core/src/utils/geth.rs
```diff
@@ -364,7 +372,8 @@ impl Geth {
         // Open the HTTP API
         cmd.arg("--http");
         cmd.arg("--http.port").arg(port.to_string());
-        cmd.arg("--http.api").arg(API);
+        let apis = if self.enable_debug { format!("{},debug", API) } else { API.to_owned() };
+        cmd.arg("--http.api").arg(apis);
```
mattsse, line 376: I would like there to be a `Vec<String>` for all additional modules. so we can also enable admin for example.

Should this also be enabled for `ws.api`?

mattsse (changes requested):

ethers-core/src/utils/geth.rs
zsluedem: Yes. You are right.

zsluedem: > I would like there to be a `Vec<String>` for all additional modules. so we can also enable admin for example.

Do you want me to add another method like `enable_admin`?

zsluedem: @mattsse I made some changes. Take a look when you are available. Thanks.

mattsse (approved):
lgtm, pending q regarding trailing comma

ethers-core/src/utils/geth.rs
```diff
@@ -364,7 +380,15 @@ impl Geth {
...
         cmd.arg("--http.port").arg(port.to_string());
-        cmd.arg("--http.api").arg(API);
+        let mut additional_apis = vec![];
+        if self.enable_debug {
+            additional_apis.push("debug".to_string())
+        };
+        if self.enable_admin {
+            additional_apis.push("admin".to_string())
+        };
+        let enable_apis = format!("{},{}", API, additional_apis.join(","));
```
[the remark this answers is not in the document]
zsluedem: Good catch. I didn't notice that. I just checked with geth and it worked fine with trailing comma.

ethers-core/src/utils/geth.rs
```diff
@@ -21,7 +21,7 @@ const GETH_STARTUP_TIMEOUT_MILLIS: u64 = 10_000;
 const GETH_DIAL_LOOP_TIMEOUT: Duration = Duration::new(20, 0);
 
 /// The exposed APIs
-const API: &str = "eth,net,web3,txpool,admin,personal,miner";
+const API: &str = "eth,net,web3,txpool,personal,miner";
```
zsluedem, line 24: @mattsse I forgot to remove the default admin here. This would change the default behaviour.

mattsse: oh my bad didn't realise this was in by default,

sorry about the confusion.

we want to keep admin by default. so the enable_admin should be removed, but we can keep the Vec, so custom namespaces can be enabled.

gakonst (commented):
no disagreement beyond not changing the defaults

PR #2145, opened 2023-02-13
fix(core): test hive genesis parsing

Rjected:
## Motivation

Hive tests require parsing many different geth `genesis.json`s. This adds tests for deserializing many of the genesis files contained in the hive test suite, revealing that we don't correctly parse all of them.

TODO:
 - [x] fix the genesis deserialization so that these tests pass

## Solution

Use `serde(default)` for all required genesis fields, and add [genesis test fields from geth](https://github.com/ethereum/go-ethereum/blob/03585ed7a932a6002569c382ab1d3b8d62819b07/core/genesis.go#L60-L66) to `Genesis`.

Support both hex and decimal decoding for `GenesisAlloc` fields.

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


ethers-core/src/utils/genesis.rs
```diff
@@ -31,13 +35,33 @@ pub struct Genesis {
...
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub gas_used: Option<U64>,
+
+    /// The block parent hash
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub parent_hash: Option<H256>,
+
+    /// The base fee
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub base_fee_per_gas: Option<U256>,
```
gakonst, line 64: I would consider putting this in a `hive: HiveParams` w/ `serde(flatten)` but we can leave for follow-up, we don't use the genesis type in other places either way

gakonst (approved):

gakonst:
CI issues unrelated

ethers-core/src/utils/genesis.rs
Rjected: ah, yeah that makes sense

PR #2161, opened 2023-02-17
fix(types): NameOrAddress parsing

DaniPopes:
## Motivation

Fixes foundry-rs/foundry#4378

Clap AutoParser priority is as follows:
```log
     = note: the following trait bounds were not satisfied:
             `opts::cast::A: ValueEnum`
             which is required by `&&&&&_AutoValueParser<opts::cast::A>: clap::builder::via_prelude::_ValueParserViaValueEnum`
             `opts::cast::A: ValueParserFactory`
             which is required by `&&&&&&_AutoValueParser<opts::cast::A>: clap::builder::via_prelude::_ValueParserViaFactory`
             `opts::cast::A: std::convert::From<OsString>`
             which is required by `&&&&_AutoValueParser<opts::cast::A>: clap::builder::via_prelude::_ValueParserViaFromOsString`
             `opts::cast::A: std::convert::From<&'s std::ffi::OsStr>`
             which is required by `&&&_AutoValueParser<opts::cast::A>: clap::builder::via_prelude::_ValueParserViaFromOsStr`
             `opts::cast::A: std::convert::From<std::string::String>`
             which is required by `&&_AutoValueParser<opts::cast::A>: clap::builder::via_prelude::_ValueParserViaFromString`
             `opts::cast::A: std::convert::From<&'s str>`
             which is required by `&_AutoValueParser<opts::cast::A>: clap::builder::via_prelude::_ValueParserViaFromStr`
             `opts::cast::A: FromStr`
```

So `From<String>` and `From<&str>` take priority over `FromStr`, and these implementation aren't that clear/useful anyway...

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/types/ens.rs
+++ b/ethers-core/src/types/ens.rs
@@ -75,23 +75,7 @@ impl<'de> Deserialize<'de> for NameOrAddress {
     }
 }
 
-impl From<&str> for NameOrAddress {
-    fn from(s: &str) -> Self {
-        Self::from_str(s).unwrap()
-    }
-}
-
-impl From<String> for NameOrAddress {
-    fn from(s: String) -> Self {
-        Self::Name(s)
-    }
-}
-
-impl From<&String> for NameOrAddress {
-    fn from(s: &String) -> Self {
-        Self::Name(s.clone())
-    }
-}
+// Do not implement `From<String,&str>`
 
 impl From<Address> for NameOrAddress {
     fn from(s: Address) -> Self {


mattsse (approved):
this is reasonable imo

DaniPopes:
Actually `Into<NameOrAddress>` is used only in `TransactionRequest::to`, I'm thinking about using `TryInto<NameOrAddress, Error = Infallible>` just so we can still remove the implementations
wdyt @mattsse

mattsse:
hmm oh yeah this would remove the convenience of `to()` and would be a breaking change. if we use TryInto fallible here then this would be the same behavior as before I think.

but perhaps the more straightforward fix for foundry would be a custom value parser?

PR #2163, opened 2023-02-17
Include deployed bytecode in abigen output 

FrankieIsLost:
## Motivation

Currently, abigen output includes `bytecode`. However, there are certain use cases when having the `deployedBytecode` is also useful. For example, when using [eth_call state overrides](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth), it's useful to inject `deployedBytecode` into an account before executing the call. A common use-case for this is writing a contract with view functions that allow you to batch-read some state onchain in order to reduce the number of RPC calls you make. 

## Solution

Add functionality to include `deployedBytecode` in the abigen output. 

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/contract.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract.rs
@@ -110,6 +110,9 @@ pub struct Context {
 
     /// Bytecode extracted from the abi string input, if present.
     contract_bytecode: Option<Bytes>,
+
+    /// Deployed bytecode extracted from the abi string input, if present.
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm
wdyt @DaniPopes ?

DaniPopes (approved):
Makes sense since we are already parsing it in `AbiObject` after #2024

PR #2164, opened 2023-02-18
feat(contract): improve Multicall result handling

DaniPopes:
## Motivation

Currently, if Multicall version is set to 2 or 3, `.call()`'s result will be a tuple of `(bool, <_>)` where bool is success. This is not the best API since it differs from version 1 and it is likely to throw and error when if the bool success variable is false, since the revert data almost never matches the success data.

This PR addresses this by making use of  the standard `Result` type to represent the call's result, where `Ok` is used when the `success` boolean is true, `Err` otherwise.

Breaking changes:
- `.call()`'s generic result type in all versions is now just the tuple of types, instead of a tuple of `(bool, <_>)` tuples
- `.raw_call()`'s concrete result type was changed from `Vec<Token>` to `Vec<Result<Token, Bytes>>`, explained above

## Solution

On top of the changes described above:
- made all of the `Multicall` struct's fields public
- changed `.call()` and `.call_array()`'s generic type trait bound from `Detokenize` to `Tokenizable`, because it is better suit for what these methods do, which is take a vector of tokens and map them to a Rust tuple (single token). This is not a breaking change since `Detokenize` is a supertrait of `Tokenizable`
- improved the `Detokenize` implementation
- changed `ethers-contract`'s edition to 2021 to match the other crates
- improved existing multicall test and reflected changes in the methods' and the `Multicall` struct's docs

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog
-   [x] Breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -301,6 +301,8 @@
 
 ### Unreleased
 
+- (Breaking) Improve Multicall result handling
+  [#2164](https://github.com/gakonst/ethers-rs/pull/2105)
 - (Breaking) Make `Event` objects generic over borrow & remove lifetime
   [#2105](https://github.com/gakonst/ethers-rs/pull/2105)
 - Make `Factory` objects generic over the borrow trait, to allow non-arc mware
--- a/ethers-contract/Cargo.toml
+++ b/ethers-contract/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "ethers-contract"
 version = "1.0.2"
-edition = "2018"
+edition = "2021"
 rust-version = "1.64"
 authors = ["Georgios Konstantopoulos <me@gakonst.com>"]
 license = "MIT OR Apache-2.0"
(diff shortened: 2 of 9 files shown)


gakonst (commented):
not sure how i feel about this. will think about it. cc @frankieislost PTAL at the change in the `call_raw` api and please share your thoughts on it

DaniPopes:
Extra context:
When a call fails (and it was allowed to), it will be returned as either `Token::Bytes` or `Token::String`, which is not always the same type as the "Ok" return type of the call, thus even if we allow it to revert, `call` and `call_array` will fail turning them from tokens to native types, and also the error is not recoverable in this case.
This is why I think using `Result` (as it's also quite natural given that there are an Ok and an Err return type...) is a necessary breaking change to have.

If it fails and it wasn't allowed to, `call_raw` early returns with a middleware error at `.call()` on the ContractCall

gakonst (approved):
synced w frankie offline, let's send it

PR #2169, opened 2023-02-20
Make order of types in shared_types deterministic

sveitser:
## Motivation
The order in which the types in `shared_types.rs` appear is not deterministic. This is can be frustrating if the bindings are committed to version control.

Unfortunately I don't have a simple repro or a test case right now.

The PR also includes some additions (gcc, anvil, geth) necessary to run the tests in the dev shell on NixOS. cc @onsails 

Unfortunately I still can't run clippy, which fails with 
```
/nix/store/ir58ws685pawbh0q3ys66aixy7hnv2vx-devenv-profile/bin/cargo-clippy: error while loading shared libraries: libstd-62c4894b82797b30.so: cannot open shared object file: No such file or directory
```

## Solution
Instead of using HashMap/HashSet use BTree so that items are iterated over in order.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/multi.rs
+++ b/ethers-contract/ethers-contract-abigen/src/multi.rs
@@ -5,7 +5,7 @@ use inflector::Inflector;
 use proc_macro2::TokenStream;
 use quote::quote;
 use std::{
-    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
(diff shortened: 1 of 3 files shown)


gakonst (approved):
ah yes of course. good catch. the flake lock file seems fine, should make using easier

PR #2171, opened 2023-02-20
chore: added canto network

syahrul12345:
## Motivation

Include Canto and CantoTestnet as a Chain in the ethers project
## Solution

Add in Canto and CantoTestnet chain ids (7700 and 740)

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

PR #2172, opened 2023-02-21
Expose contract revert errors in the ContractError struct

prestwich:
## Motivation

Users don't have access to revert data. That sucks. I'm fixing it.

## Solution

- Add logic to spelunk for json rpc revert data, inspired by ethers-rs
- Add logic to contract
- Break old usages (and direct instantiation) of middleware error and provider error variants of contract error enum

TODO: 
- [x] test this code

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog
-   [x] Breaking changes


ethers-contract/src/call.rs
```diff
@@ -72,6 +78,34 @@ pub enum ContractError<M: Middleware> {
...
+impl<M: Middleware> ContractError<M> {
+    pub fn decode_revert<Err: EthError>(&self) -> Option<Err> {
+        match self {
+            ContractError::Revert(data) => Err::decode(data).ok(),
+            _ => None,
+        }
+    }
+
+    /// Convert a Middleware Error to a `ContractError`
+    pub fn from_middleware_error(e: M::Error) -> Self {
```
prestwich, line 90: can't make `From<M::Error>` because it potentially conflicts with blanket `From<T> for T` in cases where `M::Error` and `ContractError<M>` are the same type

yes this is weird

ethers-contract/src/call.rs
```diff
@@ -54,12 +56,16 @@ pub enum ContractError<M: Middleware> {
     DetokenizationError(#[from] InvalidOutputType),
 
     /// Thrown when a middleware call fails
-    #[error("{0}")]
-    MiddlewareError(M::Error),
+    #[error("{e}")]
+    MiddlewareError { e: M::Error },
```
prestwich, line 60: deliberately broken 3rd party direct instantiation, so that users must use the `from_middleware_error` method

similar for provider error. broke it to force `From` usage

ethers-providers/src/rpc/transports/common.rs
```diff
@@ -21,6 +24,45 @@ pub struct JsonRpcError {
...
+    /// Attempt to extract revert data from the JsonRpcError be recursively
+    /// traversing the error's data field
+    ///
+    /// This returns the first hex it finds in the data object, and its
+    /// behavior may change with `serde_json` internal changes.
+    ///
+    /// Inspired by ethers-js logic:
+    /// https://github.com/ethers-io/ethers.js/blob/9f990c57f0486728902d4b8e049536f2bb3487ee/packages/providers/src.ts/json-rpc-provider.ts#L25-L53
+    pub fn as_revert_data(&self) -> Option<Bytes> {
+        self.is_revert().then(|| self.data.as_ref().and_then(spelunk_revert)).flatten()
```
prestwich, line 57: TIL about `bool::then`

ethers-contract/src/call.rs
gakonst: Yeah, this would be fixed with specialization, but alas

gakonst (commented):
feature :+1:, breaking change good w/ me

prestwich:
looks like linter error is related to #2171

prestwich:
cc @DaniPopes this touches your recent work 😅

ethers-contract/tests/it/contract.rs
```diff
@@ -722,19 +722,24 @@ mod eth_tests {
...
-        assert!(matches!(
-            multicall.call::<(String,)>().await.unwrap_err(),
-            MulticallError::ContractError(_)
-        ));
+        // don't allow revert
+        multicall
+            .clear_calls()
+            .add_call(get_value_reverting_call.clone(), false)
+            .add_call(get_value_call.clone(), false);
+        let res = multicall.call::<((bool, String), (bool, String))>().await;
```
DaniPopes, line 730: this should just be `(String, String)`

DaniPopes (commented):
@prestwich That's fine since you're integrating it with the general Contract error, my PR was mostly about having a better API for the multicall call methods

gakonst (approved):

PR #2177, opened 2023-02-22
feat: add filecoin chains

AmeanAsad:
## Motivation

Adding Filecoin chains into the `Chain` enum. Filecoin is launching its own virtual machine and adding the Filecoin chains into `ethers-rs` allows developers to use them in the rust web3 tooling. The main personal use I have is estimating chain gas costs when using `Foundry` to deploy smart contracts on the Filecoin virtual machine. 

## Solution

Added filecoin chains into the `Chain` enum and the other corresponding spots required for their definition. 

## PR Checklist

-   [ ] Added Tests (N/A) 
-   [ ] Added Documentation (N/A)
-   [ ] Updated the changelog 
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/types/chain.rs
+++ b/ethers-core/src/types/chain.rs
@@ -478,8 +478,21 @@ impl Chain {
             Canto | CantoTestnet => "BLOCKSCOUT_API_KEY",
 
             // Explicitly exhaustive. See NB above.
-            XDai | Chiado | Sepolia | Rsk | Sokol | Poa | Oasis | Emerald | EmeraldTestnet |
-            Evmos | EvmosTestnet | AnvilHardhat | Dev => return None,
+            XDai |
+            Chiado |
+            Sepolia |
+            Rsk |
+            Sokol |
+            Poa |
+            Oasis |
+            Emerald |
+            EmeraldTestnet |
+            Evmos |
+            EvmosTestnet |
+            AnvilHardhat |
+            Dev |
+            FilecoinMainnet |
+            FilecoinHyperspaceTestnet => return None,
         };
 
         Some(api_key_name)


ethers-core/src/types/chain.rs
```diff
@@ -299,6 +303,9 @@ impl Chain {
             Chiado => {
                 ("https://blockscout.chiadochain.net/api", "https://blockscout.chiadochain.net")
             }
+            FilecoinHyperspaceTestnet => {
+                ("https://api.hyperspace.node.glif.io/rpc/v1", "https://hyperspace.filfox.info")
+            }
```
mattsse, line 308: are these compatible with etherscan API?

AmeanAsad: They are not

prestwich:
we can't guarantee that ethers will behave properly for any virtualized VM 🤔

AmeanAsad:
@prestwich the filecoin virtual machine is EVM compatible (idk if that helps). The main reason for this request is to enable gas estimation in `Foundry` which uses ethers `Chain` enum. As of now that is the only functionality required from this change. Is that acceptable?

prestwich:
the FVM has a virtualized EVM, this is not necessarily the same thing as being EVM-compatible 😅

ethers-core/src/types/chain.rs
gakonst: Hmm then we cannot use them. Where can we read more about the filecoin evm spec?

AmeanAsad: @gakonst  Here are some resources regarding docs/specs:
- [Spec](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0030.md)
- [Official docs](https://docs.filecoin.io/developers/smart-contracts/filecoin-virtual-machine/)

DaniPopes:
I think this is ok, given that:
- it's on [Chainlist](https://chainlist.org/chain/314)
- they have docs for usage with [Metamask](https://docs.filecoin.io/developers/smart-contracts/how-tos/add-to-metamask/) and [Hardhat](https://docs.filecoin.io/developers/smart-contracts/hardhat/)

Although they have their own precompiles

gakonst (approved):
K - given no block explorer handlign im fine w this

PR #2181, opened 2023-02-23
RRR: Reconnection & Request Reissuance

prestwich:
Adds automatic reconnection and request-reissuance logic to WebSockets

## Motivation

Users keep making a fuss about it and also it's generally a good thing to have :)

## Solution

WS now consists of 2 tasks:
1. `WsBackend` - manages dispatching raw values to the WebSocket
2. `RequestManager` - manages caching request bodies until the request has resolved

`WsBackend` dispatches requests and routes responses and notifications. It also has a simple ping-based keepalive (when not compiled to wasm), to prevent inactivity from triggering server-side closes

`RequestManager` holds a `BackendDriver`, to communicate with the current backend. Reconnection is accomplished by instantiating a new `WsBackend` and swapping out the manager's `BackendDriver`

The `RequestManager` holds copies of all pending requests (as `InFlight`), and active subscriptions (as `ActiveSub`). When reconnection occurs, all pending requests are re-dispatched to the new backend, and all active subs are re-subscribed

In order to provide continuity of subscription IDs to the client, the `RequestManager` also keeps a `SubscriptionManager`. This struct manages the relationship between the u64 request ID, and U256 server-side subscription ID. It does this by aliasing the server ID to the request ID, and returning the Request ID to the caller (hiding the server ID in the `SubscriptionManager` internals.) Giving the caller a "fake" subscription id allows the subscription to behave consistently across reconnections

The behavior is accessed by the `WsClient` frontend, which implements `JsonRpcClient`. The `WsClient` is cloneable (no need for an arc 😄). It communicates to the request manager via a channel, and receives notifications in a shared map for the client to retrieve

The `RequestManager` shuts down and drops when all `WsClient` instances have been dropped (because all `UnboundedSender` instances will have dropped).

The `WsBackend` shuts down when instructed to by the `RequestManager` or when the `RequestManager` drops (because the inbound channel will close)

### Note on use of mutexes

This work does NOT use async mutexes, as we never need to hold past an await point. Because each use of mutex is fast (simple hashmap insert/remove), we're not concerned about blocking on locking

### Limitations: 

- reconnection may cause dropped events in subscriptions, as the new sub may resume at a later height. This seems unavoidable? 🤔 
- faking the subscription ID requires an extra deser and reser on each `eth_subscribe` call so that the manager can inspect the sub id
- memory usage will be higher, as we're keeping a copy of in-flight request params while they're in flight, and a permanent copy of the subscription params

### Follow-up work:

Because `BackendDriver` does not contain any WS-specific behavior, we can add a `RequestManager` in front of IPC channels to add the same reconnection + reissuance logic. This doesn't seem high priority, as WS is much more prone to ephemeral failures than IPC

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog
-   [x] Breaking changes


gakonst:
@mattsse PTAL - I'll cut a release after we get this PR in

prestwich:
going to move all previous WS behavior under `legacy_ws` and add a feature flag to use it

ethers-providers/src/rpc/transports/ws/backend.rs
```diff
@@ -0,0 +1,194 @@
...
+                select! {
+                    _ = keepalive => {
+                        #[cfg(not(target_arch = "wasm32"))]
+                        if let Err(e) = self.server.send(Message::Ping(vec![])).await {
+                            error!("WS connection error {e}");
+                            err = true;
+                            break
+                        }
+                        #[cfg(target_arch = "wasm32")]
+                        unreachable!();
```
gakonst, line 145: ah nice; smart. so the future never completes so obvi it won't reach that branch

cc @ncitron in case you're using the WS code in WASM - would love you to give this new WS code a try

ethers-providers/src/rpc/transports/ws/backend.rs
```diff
@@ -0,0 +1,194 @@
...
+                #[cfg(not(target_arch = "wasm32"))]
+                let keepalive = tokio::time::sleep(std::time::Duration::from_secs(10)).fuse();
+                #[cfg(not(target_arch = "wasm32"))]
+                tokio::pin!(keepalive);
+
+                // in wasm, we don't ping. as ping doesn't exist in our wasm lib
+                #[cfg(target_arch = "wasm32")]
+                let mut keepalive = futures_util::future::pending::<()>().fuse();
+
+                select! {
```
gakonst, line 136: tokio select is just GOAT

ethers-providers/src/rpc/transports/ws/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+impl JsonRpcClient for WsClient {
+    type Error = WsClientError;
+
+    async fn request<T, R>(&self, method: &str, params: T) -> Result<R, WsClientError>
+    where
+        T: Serialize + Send + Sync,
+        R: DeserializeOwned,
+    {
+        let params = serde_json::to_string(&params)?;
+        let params = RawValue::from_string(params)?;
```
gakonst, line 80: This sucks enormously much and I wonder if we could instead do some `unsafe` `transmute` stuff to get around it.

ethers-providers/src/rpc/transports/ws/manager.rs
```diff
@@ -0,0 +1,373 @@
...
+    WsClient, WsClientError,
+};
+
+pub type SharedChannelMap = Arc<Mutex<HashMap<U256, mpsc::UnboundedReceiver<Box<RawValue>>>>>;
+
+pub const DEFAULT_RECONNECTS: usize = 5;
+
+pub struct SubscriptionManager {
+    subs: BTreeMap<u64, ActiveSub>,
+    aliases: HashMap<U256, u64>,
```
gakonst, line 28: realize it's a chore but do you mind adding docs from the PR description on the motivation on why we need to wire together the internal representation of subscriptions to the IDs we get on each reconnect? OK to make it a follow-up or open an issue

gakonst (approved):
LGTM. Much needed feature. Looking for a pass from @mattsse before merging.

ethers-providers/src/rpc/transports/ws/mod.rs
```diff
@@ -0,0 +1,145 @@
+#![allow(missing_docs)]
+
+pub(crate) mod macros;
+
```
gakonst, line 4:
```suggestion
```

ethers-providers/src/rpc/transports/ws/manager.rs
prestwich: ack. I am adding all commentary from the pr description to the relevant structs.

ethers-providers/src/rpc/transports/ws/backend.rs
prestwich: the downside is that there's no keepalive on wasm ws connections. wasm doesn't support ping/pong so if we were to ever add a JS setTimeout based keepalive here, it would have to issue a real rpc request

ethers-providers/src/rpc/transports/ws/mod.rs
DaniPopes: What's wrong with it? RawValue::from_string is basically that already no?

ethers-providers/src/rpc/transports/ws/backend.rs
gakonst: yup

ethers-providers/src/rpc/transports/ws/backend.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+impl BackendDriver {
+    pub fn shutdown(self) {
+        // don't care if it fails, as that means the backend is gone anyway
+        let _ = self.shutdown.send(());
+    }
+}
+
+impl WsBackend {
```
mattsse, line 42: this impl block is out of order

ethers-providers/src/rpc/transports/ws/mod.rs
mattsse: this is very little overhead because this only checks for valid json and does not allocate.

ethers-providers/src/rpc/transports/ws/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+impl PubsubClient for WsClient {
+    type NotificationStream = mpsc::UnboundedReceiver<Box<RawValue>>;
+
+    fn subscribe<T: Into<U256>>(&self, id: T) -> Result<Self::NotificationStream, WsClientError> {
+        // due to the behavior of the request manager, we know this map has
+        // been populated by the time the `request()` call returns
+        let id = id.into();
+        self.channel_map.lock().unwrap().remove(&id).ok_or(WsClientError::UnknownSubscription(id))
```
mattsse, line 93: could use parking_lot mutex here instead.

ethers-providers/src/rpc/transports/ws/manager.rs
```diff
@@ -0,0 +1,400 @@
...
+                // We bias the loop so that we always handle messages before
+                // reconnecting, and always reconnect before
+                select_biased! {
+                    item_opt = self.backend.to_handle.next() => {
+                        match item_opt {
+                            Some(item) => self.handle(item),
+                            // Backend is gone, so reconnect
+                            None => if let Err(e) = self.reconnect().await {
+                                break Err(e);
+                            }
```
mattsse, line 370: I could totally be wrong here because it wasn't obvious to which component `to_handle` is wired here. but it looks like the other half belongs to `WsBackend`?

I believe there's an issue with how the `WsBackend` task shuts down, which could cause two reconnects if the task exits because of an error?
first the channel will be None, then the  `self.backend.error` receives the error message and reconnects again?

mattsse (commented):
if I understood this right, a WsClient now has two different tasks?
I haven't looked into it much, but that's kind of weird I guess.
manager, backend, backenddriver is a but confusing.

it would be great to have some docs on the various channel fields because there are like 4 different channels used in this setup.

this now auto-reconnects on any error error caused by ws connection?

ethers-providers/src/rpc/transports/ws/manager.rs
prestwich: > I believe there's an issue with how the WsBackend task resolves, which could cause two reconnects if the task exits because of an error?

No, only 1 select branch is reachable in any pass through the loop. So if the `to_handle` channel is dead, the WS will reconnect. Then by the time the loop restarts, the `BackendDriver` will be different, and the error channel will be fresh and good

prestwich: > I could totally be wrong here because it wasn't obvious to which component to_handle is wired here. but it looks like the other half belongs to WsBackend?

the naming convention is you send a chunk of work to the handler via `handler`. the handler accepts work that it should handle via `to_handle`

same convention with `dispatcher` and `to_dispatch`. you send to the dispatching task via a `dispatcher` channel. the dispatcher accepts the requests via `to_dispatch`

prestwich: you can think of the `BackendDriver` as control or ownership of the running task. It has the ability to shut down that task, send/receive messages from it, and dropping the driver causes the task to shut down automatically

The communication flow is

client ----instructions---> manager
manager -----requests-----> backend
backend <-----WS Protocol-----> server
backend -----PubSubItem----> manager
manager ---sub notifications---> sub channels
manager ---sub channels---> client
manager ---responses----> request future channels

this would go well in a swim lanes but I don't have time to make one rn

prestwich:
> if I understood this right, a WsClient now has two different tasks?

yeah, because reconnection requires the connection to be disposable, there must be >=2 tasks. One for the connection, and 1 for detecting and fixing connection failures.

> it would be great to have some docs on the various channel fields because there are like 4 different channels used in this setup.

added comments to all struct fields

> this now auto-reconnects on any error error caused by ws connection?

yes, up to a fixed number of retries. It also includes a keepalive loop when not running in wasm

ethers-providers/src/rpc/transports/ws/mod.rs
gakonst: meh should be nbd

ethers-providers/src/rpc/transports/ws/manager.rs
```diff
@@ -0,0 +1,409 @@
...
+            self.backend.shutdown();
+        };
+
+        #[cfg(target_arch = "wasm32")]
+        super::spawn_local(fut);
+
+        #[cfg(not(target_arch = "wasm32"))]
+        tokio::spawn(fut);
+    }
+}
```
gakonst, line 409: nit: could add a `reconnects(..)` builder method to override the default, but can do in followup

gakonst (approved):
@mattsse merging cautiously; if anybody experiences issues they can use the `legacy-ws` feature.

shrimpynuts:
@prestwich Is it reasonable to try to keep a subscription through `WsClient` alive indefinitely?

If yes, would the best way be to just instantiate a new `WsClient` every time the `RequestManger` reaches the reconnect limit?

prestwich:
> Is it reasonable to try to keep a subscription through WsClient alive indefinitely?

Yes it's pretty reasonable

> If yes, would the best way be to just instantiate a new WsClient every time the RequestManger reaches the reconnect limit?

You should just set an arbitrarily high reconnection count :) u64::max should be fine

wminshew:
> reconnection may cause dropped events in subscriptions, as the new sub may resume at a later height. This seems unavoidable? 🤔

dropping events seems fine to me but out of curiosity why is it unavoidable? couldn't `SubscriptionManager` keep note of the most recent block each sub has received an event from & poll from there on re-connect?

PR #2182, opened 2023-02-24
feature: contract revert trait

prestwich:
## Motivation

Followup work to #2172. Allows easier decoding of error types for generated contracts

## Solution

- Introduce the `ContractRevert` trait
- Auto-derive `ContractRevert` on the error enum in `abigen!` contract modules
- Add `RevertString(String)` to contract error enum in `abigen!` contract modules
- Add `ContractError::decode_contract_revert<Err: ContractRevert(&self) -> Option<Err>`
- (drive-by) add missing `FromIterator` impls to `ethers::core::types::Bytes`

This is a breaing change to `abigen!` generated bindings (again) but i promise this one is also really good :)

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [x] Breaking changes

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/contract/errors.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract/errors.rs
@@ -112,6 +112,9 @@ impl Context {
             impl #ethers_core::abi::AbiDecode for #enum_name {
                 fn decode(data: impl AsRef<[u8]>) -> ::core::result::Result<Self, #ethers_core::abi::AbiError> {
                     let data = data.as_ref();
+                    // NB: This implementation does not include selector information, and ABI encoded types
+                    // are incredibly ambiguous, so it's possible to have bad false positives. Instead, we default
+                    // to a String to minimize amount of decoding attempts
                     if let Ok(decoded) = <::std::string::String as #ethers_core::abi::AbiDecode>::decode(data) {
                         return Ok(Self::RevertString(decoded))
                     }
--- a/ethers-contract/ethers-contract-abigen/src/contract/errors.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract/errors.rs
@@ -104,7 +104,7 @@ impl Context {
             #[derive(Clone, #ethers_contract::EthAbiType, #derives)]
             pub enum #enum_name {
                 #( #variants(#variants), )*
-                RevertString(String),
+                RevertString(::std::string::String),
             }
 
             impl #ethers_core::abi::AbiDecode for #enum_name {
@@ -136,10 +136,11 @@ impl Context {
             impl #ethers_contract::ContractRevert for #enum_name {
                 fn valid_selector(selector: [u8; 4]) -> bool {
                     match selector {
+                        // Error(string) -- standard solidity revert
+                        [0x08, 0xc3, 0x79, 0xa0] => true,
                         #(
                             _ if selector == <#variants as #ethers_contract::EthError>::selector() => true,
                         )*
-                        _ if selector == <::std::string::String as #ethers_contract::EthError>::selector() => true,
                         _ => false,
                     }
                 }
--- a/ethers-contract/ethers-contract-abigen/src/contract/errors.rs
+++ b/ethers-contract/ethers-contract-abigen/src/contract/errors.rs
@@ -104,11 +104,15 @@ impl Context {
             #[derive(Clone, #ethers_contract::EthAbiType, #derives)]
             pub enum #enum_name {
                 #( #variants(#variants), )*
+                RevertString(String),
             }
 
             impl #ethers_core::abi::AbiDecode for #enum_name {
                 fn decode(data: impl AsRef<[u8]>) -> ::core::result::Result<Self, #ethers_core::abi::AbiError> {
                     let data = data.as_ref();
+                    if let Ok(decoded) = <::std::string::String as #ethers_core::abi::AbiDecode>::decode(data) {
+                        return Ok(Self::RevertString(decoded))
+                    }
                     #(
                         if let Ok(decoded) = <#variants as #ethers_core::abi::AbiDecode>::decode(data) {
                             return Ok(Self::#variants(decoded))
@@ -124,17 +128,38 @@ impl Context {
                         #(
                             Self::#variants(element) => #ethers_core::abi::AbiEncode::encode(element),
                         )*
+                        Self::RevertString(s) => #ethers_core::abi::AbiEncode::encode(s),
                     }
                 }
             }
+
+            impl #ethers_contract::ContractRevert for #enum_name {
+                fn valid_selector(selector: [u8; 4]) -> bool {
+                    match selector {
+                        #(
+                            _ if selector == <#variants as #ethers_contract::EthError>::selector() => true,
+                        )*
+                        _ if selector == <::std::string::String as #ethers_contract::EthError>::selector() => true,
+                        _ => false,
+                    }
+                }
+            }
+
 
             impl ::core::fmt::Display for #enum_name {
                 fn fmt(&self, f: &mut ::core::fmt::Formatter<'_>) -> ::core::fmt::Result {
                     match self {
                         #(
-                            Self::#variants(element) => ::core::fmt::Display::fmt(element, f)
-                        ),*
+                            Self::#variants(element) => ::core::fmt::Display::fmt(element, f),
+                        )*
+                        Self::RevertString(s) => ::core::fmt::Display::fmt(s, f),
                     }
+                }
+            }
+
+            impl ::core::convert::From<::std::string::String> for #enum_name {
+                fn from(value: String) -> Self {
+                    Self::RevertString(value)
                 }
             }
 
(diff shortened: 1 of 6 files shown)


ethers-contract/src/error.rs
```diff
@@ -6,6 +6,29 @@ use ethers_core::{
 use ethers_providers::JsonRpcError;
 use std::borrow::Cow;
 
+/// A trait for enums unifying [`EthError`] types.
+///
+/// We do not recommend manual implementations of this trait. Instead, use the
+/// automatically generated implementation in the [`crate::abigen`] macro
+pub trait ContractRevert: AbiDecode + AbiEncode + Send + Sync {
+    /// Decode the error from EVM revert data including an Error selector
+    fn decode_with_selector(data: &Bytes) -> Option<Self> {
```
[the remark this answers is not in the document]
DaniPopes: Also applies to `EthError::decode_with_selector` below

DaniPopes (changes requested):

ethers-contract/ethers-contract-abigen/src/contract/errors.rs
```diff
@@ -124,20 +128,41 @@ impl Context {
...
-                        ),*
+                            Self::#variants(element) => ::core::fmt::Display::fmt(element, f),
+                        )*
+                        Self::RevertString(s) => ::core::fmt::Display::fmt(s, f),
                     }
                 }
             }
 
+            impl ::core::convert::From<::std::string::String> for #enum_name {
+                fn from(value: String) -> Self {
```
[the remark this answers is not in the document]
gakonst: probably should be `core::alloc::String`?

ethers-contract/ethers-contract-abigen/src/contract/errors.rs
```diff
@@ -124,20 +128,41 @@ impl Context {
...
+            }
+
+            impl #ethers_contract::ContractRevert for #enum_name {
+                fn valid_selector(selector: [u8; 4]) -> bool {
+                    match selector {
+                        #(
+                            _ if selector == <#variants as #ethers_contract::EthError>::selector() => true,
+                        )*
+                        _ if selector == <::std::string::String as #ethers_contract::EthError>::selector() => true,
+                        _ => false,
```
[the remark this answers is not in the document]
prestwich: it would save some number of keccak invocations, at the cost of my sanity 😬

ethers-contract/ethers-contract-abigen/src/contract/errors.rs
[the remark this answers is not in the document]
DaniPopes: `core::alloc` is just the Allocator API. it would be `::alloc::string::String` but I don't think we'll go into `no-std` territory any time soon

prestwich:
will address these comments tomorrow. ledger issue has worn my brain out

ethers-contract/ethers-contract-abigen/src/contract/errors.rs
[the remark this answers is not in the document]
prestwich: moving the String match to the top and hardcoding its selector, as that'll save keccaks in a lot of cases

ethers-contract/ethers-contract-abigen/src/contract/errors.rs
```diff
@@ -104,11 +104,15 @@ impl Context {
...
                 #( #variants(#variants), )*
+                RevertString(::std::string::String),
             }
 
             impl #ethers_core::abi::AbiDecode for #enum_name {
                 fn decode(data: impl AsRef<[u8]>) -> ::core::result::Result<Self, #ethers_core::abi::AbiError> {
                     let data = data.as_ref();
+                    if let Ok(decoded) = <::std::string::String as #ethers_core::abi::AbiDecode>::decode(data) {
+                        return Ok(Self::RevertString(decoded))
+                    }
```
gakonst, line 115: Shouldn't this be below the `if let Ok` in the macro below? So that we only decode as string as a fallback vs prioritizing it?

ethers-contract/ethers-contract-abigen/src/contract/errors.rs
```diff
@@ -124,20 +128,42 @@ impl Context {
...
                         )*
+                        Self::RevertString(s) => #ethers_core::abi::AbiEncode::encode(s),
+                    }
+                }
+            }
+
+            impl #ethers_contract::ContractRevert for #enum_name {
+                fn valid_selector(selector: [u8; 4]) -> bool {
+                    match selector {
+                        // Error(string) -- standard solidity revert
```
mattsse, line 139: can we please add this as hex to the docs as well?

ethers-contract/src/error.rs
```diff
@@ -1,11 +1,34 @@
...
 };
 use ethers_providers::JsonRpcError;
 use std::borrow::Cow;
 
+/// A trait for enums unifying [`EthError`] types.
+///
+/// We do not recommend manual implementations of this trait. Instead, use the
+/// automatically generated implementation in the [`crate::abigen`] macro
+pub trait ContractRevert: AbiDecode + AbiEncode + Send + Sync {
+    /// Decode the error from EVM revert data including an Error selector
```
mattsse, line 14: is this only derived for enums?
because this clashes with `EthError`

ethers-contract/src/error.rs
```diff
@@ -1,11 +1,34 @@
...
+    fn decode_with_selector(data: &[u8]) -> Option<Self> {
+        if data.len() < 4 {
+            return None
+        }
+        let mut selector = [0u8; 4];
+        selector.copy_from_slice(&data[..4]);
```
mattsse, line 20: can use try_into().expect() here as well.

ethers-contract/ethers-contract-abigen/src/contract/errors.rs
prestwich: This implementation has never been safe for users to access directly, as it does not include selector information, and ABI encoded types are incredibly ambiguous. It has always thrown bad false positives.

My belief is that strings will be more common than other error types, even for contracts that do not revert, so putting a string in the first attempt results in the fewest decode attempts

ethers-contract/src/error.rs
prestwich: yes, this trait is only for unifying multiple potential ethereum errors into 1 type. Basically this should be thought of as "the errors associated with a specific contract's ABI". 

I thought about breaking out a super-trait that had the shared methods on it, and decided that it was too complex for users

prestwich: i also thought about modifying AbiDecode to include `decode_with_selector(selector: Selector, bytes: &[u8])` but didn't want to get into lower level changes

prestwich: you can also consult the docstring, btw

```
/// A trait for enums unifying [`EthError`] types.
///
/// We do not recommend manual implementations of this trait. Instead, use the
/// automatically generated implementation in the [`crate::abigen`] macro```

File: ethers-contract/ethers-contract-abigen/src/contract/errors.rs [hunk 8]
```diff
@@ -104,11 +104,17 @@ impl Context {
...
                 #( #variants(#variants), )*
+                /// The standard solidity revert string, with selector
+                /// Error(string) -- 0x08c379a0
+                RevertString(::std::string::String),
             }
 
             impl #ethers_core::abi::AbiDecode for #enum_name {
                 fn decode(data: impl AsRef<[u8]>) -> ::core::result::Result<Self, #ethers_core::abi::AbiError> {
                     let data = data.as_ref();
+                    if let Ok(decoded) = <::std::string::String as #ethers_core::abi::AbiDecode>::decode(data) {
```
gakonst, line 115:
```suggestion
                    // NB: This implementation does not include selector information, and ABI encoded types
                    // are incredibly ambiguous, so it's possible to have bad false positives. Instead, we default
                    // to a String to minimize amount of decoding attempts
                    if let Ok(decoded) = <::std::string::String as #ethers_core::abi::AbiDecode>::decode(data) {
```

ethers-contract/ethers-contract-abigen/src/contract/errors.rs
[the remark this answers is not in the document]
gakonst: kk just wanted to be consistent cuz we do core::convert above

gakonst (approved):
good w/ me pending any more matt feedback

prestwich:
this is now based on latest master 👍

gakonst:
@prestwich seems to have broken in the Seaport codegen

```

---- abigen::can_generate_seaport_1_0 stdout ----
error: test failed, to rerun pass `-p ethers-contract --test it`
thread 'abigen::can_generate_seaport_1_0' panicked at 'checked by len: TryFromSliceError(())', /home/runner/work/ethers-rs/ethers-rs/ethers-contract/src/error.rs:32:40
```

prestwich:
@gakonst seaport is fixed now 👍

PR #2197, opened 2023-02-25
feat(providers): add must_use to PendingTransaction

DaniPopes:
## Motivation

I think this is fitting since sometimes you might forget calling await a second time (`.send_transaction().await.unwrap()` -> `PendingTransaction` -> `.await.unwrap().unwrap()` -> `Receipt`)

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-middleware/src/signer.rs
+++ b/ethers-middleware/src/signer.rs
@@ -499,6 +499,9 @@ mod tests {
                 None,
             )
             .await
+            .unwrap()
+            .await
+            .unwrap()
             .unwrap();
         let client = SignerMiddleware::new_with_provider_chain(provider, key).await.unwrap();
 
--- a/ethers-middleware/tests/gas_escalator.rs
+++ b/ethers-middleware/tests/gas_escalator.rs
@@ -30,9 +30,9 @@ async fn gas_escalator_live() {
     let tx = TransactionRequest::pay(Address::zero(), 1u64).gas_price(10_000_000);
 
     // broadcast 3 txs
-    provider.send_transaction(tx.clone().nonce(nonce), None).await.unwrap();
-    provider.send_transaction(tx.clone().nonce(nonce + 1), None).await.unwrap();
-    provider.send_transaction(tx.clone().nonce(nonce + 2), None).await.unwrap();
+    let _ = provider.send_transaction(tx.clone().nonce(nonce), None).await.unwrap();
+    let _ = provider.send_transaction(tx.clone().nonce(nonce + 1), None).await.unwrap();
+    let _ = provider.send_transaction(tx.clone().nonce(nonce + 2), None).await.unwrap();
 
     // Wait a bunch of seconds and refresh etherscan to see the transactions get bumped
     tokio::time::sleep(Duration::from_secs(100)).await;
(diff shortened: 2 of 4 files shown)


mattsse (commented):
disagree here because perhaps you're not even interested in the receipt and just want to send it.

wdyt @prestwich

ethers-middleware/src/signer.rs
```diff
@@ -499,6 +499,9 @@ mod tests {
                 None,
             )
             .await
+            .unwrap()
+            .await
```
prestwich, line 503: producing a receipt without checking anything about it seems unnecessay

ethers-middleware/tests/gas_escalator.rs
```diff
@@ -30,9 +30,9 @@ async fn gas_escalator_live() {
     let tx = TransactionRequest::pay(Address::zero(), 1u64).gas_price(10_000_000);
 
     // broadcast 3 txs
-    provider.send_transaction(tx.clone().nonce(nonce), None).await.unwrap();
-    provider.send_transaction(tx.clone().nonce(nonce + 1), None).await.unwrap();
-    provider.send_transaction(tx.clone().nonce(nonce + 2), None).await.unwrap();
+    drop(provider.send_transaction(tx.clone().nonce(nonce), None).await.unwrap());
```
prestwich, line 33: I think `let _ = ` is more idiomatic than `drop()`

ethers-middleware/tests/transformer.rs
```diff
@@ -133,6 +133,9 @@ async fn ds_proxy_code() {
         .legacy()
         .send()
         .await
+        .unwrap()
+        .await
```
prestwich, line 137: same as above. if we produce the receipt, it should be for testing something

DaniPopes:
> disagree here because perhaps you're not even interested in the receipt and just want to send it.

Yeah but it's the same for a Result, if you don't care about the error you have to explicitly state that you don't care by dropping it, in case you forgot

ethers-middleware/tests/gas_escalator.rs
DaniPopes: `let _ = impl Future;`  produces [`clippy::let_underscore_future`](https://rust-lang.github.io/rust-clippy/master/#let_underscore_future)

ethers-middleware/tests/transformer.rs
DaniPopes: we are checking the resolution of the tx later so it has to confirm first

prestwich:
okay here's my actual hot take, we should split up the `PendingTx` from the actual future type. Then we can have the must use on the future, and allow 

``` 
// New struct with same name and deref behavior & builder behavior for the future
pub struct PendingTransaction { .. };

impl PendingTransaction {
    fn wait_for_receipt(self) -> TxAwaiter;
}

// Current pending transaction struct
#[must_use]
pub struct TxAwaiter { .. } 

impl std::future::IntoFuture for PendingTransaction {
    type IntoFuture = TxAwaiter;

    fn into_future(self) -> Self::IntoFuture {
        self.wait_for_receipt()
    }
}
```

gakonst:
I think let's save this for 2.0. Agree with the basis that I might want to skip polling the receipt.

PR #2203, opened 2023-02-27
docs: fix broken links, update documentation

DaniPopes:
## Motivation
commits:
1. Allows running: `cargo doc --workspace --all-features --no-deps --document-private-items`
2. While at it, updated all crates to have README.md + using `#![doc = include_str!("../README.md")]` in lib.rs

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-solc/src/lib.rs
+++ b/ethers-solc/src/lib.rs
@@ -1,5 +1,4 @@
 #![deny(rustdoc::broken_intra_doc_links)]
-#![allow(rustdoc::private_intra_doc_links)]
 
 pub mod artifacts;
 pub mod sourcemap;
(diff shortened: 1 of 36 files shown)


ethers-solc/src/lib.rs
```diff
@@ -24,7 +24,7 @@ pub use compile::{
 };
 
 mod config;
-pub use config::{AllowedLibPaths, PathStyle, ProjectPathsConfig, SolcConfig};
+pub use config::{AllowedLibPaths, PathStyle, ProjectPaths, ProjectPathsConfig, SolcConfig};
```
DaniPopes, line 27: ProjectPaths is referenced and exposed in `ProjectPathsConfig`, but not exported at the crate level, so it failed to document

src/lib.rs
```diff
@@ -78,56 +73,38 @@
...
+pub use ethers_solc as solc;
 
-/// Etherscan bindings
-pub mod etherscan {
-    pub use ethers_etherscan::*;
-}
-
-pub use crate::core::{abi, types, utils};
+#[doc(inline)]
+pub use ethers_core::{abi, types, utils};
```
DaniPopes, line 105: This inlines the modules' docs as if we declared them in this scope. This is probably what the previous interface was going for, but this doesn't require duplicating docs and declaring new modules

mattsse (approved):
lgtm

PR #2218, opened 2023-03-01
i256: Fix logical errors in doc comments for is_negative and is_zero

harpocryptes:
This should be less confusing ;)


gakonst (approved):
good catch

PR #2220, opened 2023-03-01
test: simplify test cleanup

mattsse:
get rid of some weird IO errors:
https://github.com/gakonst/ethers-rs/actions/runs/4305262960/jobs/7507450686

no need for close: executed in drop impl
no need for drop: dropped when out of scope


rkrasiuk:
@mattsse 
> no need for close: executed in drop impl

these aren't equivalent. see https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#method.close

PR #2221, opened 2023-03-01
fix: docs.rs build

DaniPopes:
## Motivation

Hopefully

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


mattsse (approved):
this is due to intra doc links via include! readme that we use both in ethers main and sub crates?

I also tried solving it like this but failed initially.

DaniPopes:
No, I was going based off of this: <https://docs.rs/crate/ethers/1.5.0/builds/755205>

PR #2222, opened 2023-03-01
chore(deps): bump and use workspace dependencies

### LINKED ISSUE #1963: Support workspace dependencies

DaniPopes:
## Motivation

Closes #1963

## Solution

See commit log

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


gakonst:
sorry need to bump versions to 2.0 everywhere

.github/workflows/ci.yml
```diff
@@ -147,9 +165,6 @@ jobs:
             - uses: Swatinem/rust-cache@v2
             - name: Check
               run: cargo check --workspace --target wasm32-unknown-unknown
-            # TODO: [#2191](https://github.com/gakonst/ethers-rs/issues/2191)
```
gakonst, line 150: is this no longer a problem?

ethers-providers/src/rpc/transports/retry.rs
```diff
@@ -338,9 +335,7 @@ where
                 trace!("retrying and backing off for {:?}", next_backoff);
 
                 #[cfg(target_arch = "wasm32")]
-                wasm_timer::Delay::new(next_backoff)
-                    .await
-                    .map_err(|_| RetryClientError::TimerError)?;
+                futures_timer::Delay::new(next_backoff).await;
```
gakonst, line 338: is this correct? if so, should we just use it in both cases?

ethers-solc/src/compile/mod.rs
```diff
@@ -70,7 +70,7 @@ pub(crate) fn take_solc_installer_lock() -> std::sync::MutexGuard<'static, ()> {
 /// A list of upstream Solc releases, used to check which version
 /// we should download.
 /// The boolean value marks whether there was an error accessing the release list
-#[cfg(all(feature = "svm-solc"))]
+#[cfg(all(feature = "svm-solc", not(target_arch = "wasm32")))]
```
gakonst, line 73: Does anything in ethers-solc work inside of WASM, given it's calling the binary? Should we just make the crate not compile with ethers-rs if wasm is not configured?

examples/anvil/Cargo.toml
```diff
@@ -1,10 +1,14 @@
 [package]
```
gakonst, line 1: Can we make `cargo release` not go over the examples crates if possible? Right now it thinks all the examples should be published

ethers/Cargo.toml
```diff
@@ -0,0 +1,89 @@
+[package]
+name = "ethers"
+authors = ["Georgios Konstantopoulos <me@gakonst.com>"]
+readme = "../README.md"
+description = "A complete Ethereum and Celo Rust library"
```
gakonst, line 5: Are we OK with pulling ethers back down a directory? I initially had pulled it up to surface the examples, but seems fine now given we have separate examples crates

gakonst (approved):
fine w me - want +1 from @mattsse

examples/anvil/Cargo.toml
DaniPopes: doesn't `publish = false` do the trick? What's the command you run to release

.github/workflows/ci.yml
DaniPopes: hmm i forget why i removed that, it still is with the features

ethers/Cargo.toml
DaniPopes: I think so, also this way it doesn't pull in every folder like `.github` when packaging the `ethers` crate for release

gakonst (approved):
ok let's do this

PR #2235, opened 2023-03-06
fix: docsrs builds final

DaniPopes:
## Motivation

For some reason (I believe because the build process isn't normal and maybe our resolution doesn't work with the given environment) the docs.rs builds always return `ethers` as the crate name (default). Tested multiple times on a local docs.rs instance and I can 99.9999999% confirm that this PR solves it by doing the following:

- re-export the internal crate as `ethers`
- export core, contract and providers at the top level
- rustc flags also have to be set to --cfg docsrs (for `#[cfg(docsrs)]` on the re-exports)
  - this is technically not necessary now because I removed the `cfg`s, just to be safe
  
this needs to be done in every (release) crate that uses a macro that uses `ethers_core::macros::*`, which right now includes contract and middleware.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ default-members = [
 
 [package.metadata.docs.rs]
 all-features = true
+rustc-args = ["--cfg", "docsrs"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [package.metadata.playground]
--- a/ethers-middleware/src/lib.rs
+++ b/ethers-middleware/src/lib.rs
@@ -40,3 +40,18 @@ pub use timelag::TimeLag;
 /// [`Middleware`](ethers_providers::Middleware) in a concise way
 pub mod builder;
 pub use builder::MiddlewareBuilder;
+
+// For macro expansions only, not public API.
+
+#[doc(hidden)]
+#[allow(unused_extern_crates)]
+extern crate self as ethers;
+
+#[doc(hidden)]
+pub use ethers_contract as contract;
+
+#[doc(hidden)]
+pub use ethers_core as core;
+
+#[doc(hidden)]
+pub use ethers_providers as providers;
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -38,6 +38,7 @@ default-members = [
 
 [package.metadata.docs.rs]
 all-features = true
+rustc-args = ["--cfg", "docsrs"]
```
mattsse, line 41: why do we need this?

ethers-middleware/src/lib.rs
```diff
@@ -40,3 +40,18 @@ pub use timelag::TimeLag;
...
+extern crate self as ethers;
+
+#[doc(hidden)]
+pub use ethers_contract as contract;
+
+#[doc(hidden)]
+pub use ethers_core as core;
+
+#[doc(hidden)]
+pub use ethers_providers as providers;
```
mattsse, line 57: hmm, I'm not sure I like that these are now always there.

mattsse (commented):
I'm a bit skeptical that messing with module alias just to get docs working is the right way to fix this.

Cargo.toml
DaniPopes: cfg is a rustc arg not a rustdoc

https://docs.rs/about/builds#cross-compiling

ethers-middleware/src/lib.rs
DaniPopes: if you have a better way to find ethers crate names then go ahead, this is a patch for builds when that is not possible, and only for our internal crates

mattsse: I don't have a better solution for this, and seems like this is a proper fix. 
But let's document why we're doing this

gakonst (approved):
agree with Matt that it sucks we need to do this, but let's roll with it

PR #2238, opened 2023-03-06
added keys and mnemonic to wallet

Andy53:
I have added the public key, private key and mnemonic to the Wallet struct so they are easily accessible by users. 

## Motivation

I think it is generally a good idea to give users access to these values so they can easily control their assets.

## Solution

This is fairly straight forward. I have added 3 option<string> to the wallet struct and populated them when a wallet is generated.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

This is not a breaking change. I have added a changeling entry.

### DIFF
--- a/ethers-signers/src/wallet/mod.rs
+++ b/ethers-signers/src/wallet/mod.rs
@@ -71,12 +71,21 @@ pub struct Wallet<D: DigestSigner<Sha256Proxy, RecoverableSignature>> {
     pub(crate) address: Address,
     /// The wallet's chain id (for EIP-155)
     pub(crate) chain_id: u64,
+    // The wallet's mnemonic
+    pub mnemonic: Option<String>,
+    // The wallet's private key
+    pub private_key: Option<String>,
+    // The wallet's public key
+    pub public_key: Option<String>
 }
 
 impl<D: DigestSigner<Sha256Proxy, RecoverableSignature>> Wallet<D> {
     /// Construct a new wallet with an external Signer
     pub fn new_with_signer(signer: D, address: Address, chain_id: u64) -> Self {
-        Wallet { signer, address, chain_id }
+        let mnemonic = None;
+        let private_key = None;
+        let public_key = None;
+        Wallet { signer, address, chain_id, mnemonic, private_key, public_key }
     }
 }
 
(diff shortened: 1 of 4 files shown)


ethers-signers/src/wallet/mod.rs
```diff
@@ -71,12 +71,21 @@ pub struct Wallet<D: DigestSigner<Sha256Proxy, RecoverableSignature>> {
     pub(crate) address: Address,
     /// The wallet's chain id (for EIP-155)
     pub(crate) chain_id: u64,
+    // The wallet's mnemonic
+    pub mnemonic: Option<String>,
+    // The wallet's private key
+    pub private_key: Option<String>,
+    // The wallet's public key
+    pub public_key: Option<String>
```
mattsse, line 79: I don't think this is a good idea.
the `Wallet` only requires that the signer type is able to sign data and the signer may not even have access to the private key directly.
and if it does (LocalWallet) then private_key/public_key is redundant here.

Making them public also makes it possible to modify them which could cause all sorts of issues.

mattsse (commented):
> I think it is generally a good idea to give users access to these values so they can easily control their assets.

While this sounds reasonable at first, it would potentially cause problems, such as unintentionally leaking them via println

ethers-signers/src/wallet/mod.rs
Andy53: Thank you for the feedback.

Some things to think about with this pull request:

1. I haven't fundamentally changed the way wallets work. If you change these values all that will happen is the these values will change. They are never actually used anywhere. I appreciate they could be in the future and making them constants would probably help remediate this.

2. I'm not sure if I have misunderstood your concern about leaking these values via println but two things I would think of here are that if someone can modify the library then they could just make the changes I made and if you're concerned about someone implementing this accidentally leaking a private key via println could they not also leak the same data via the method I used to assign the private key variable just as easily? (`format!("0x{:02X?}", signer.to_bytes()).replace(", ", "").replace("[", "").replace("]", "")`)

prestwich:
this is a concept NACK from me. Users can and should use `.signer()` to get access to the underlying signer info

prestwich:
Closing as stale, given discussion has stalled

PR #2243, opened 2023-03-08
feat(core): add implementations to Opcode

DaniPopes:
## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


mattsse (approved):
lgtm,
I think it is ok to add them here and not reexport revm opcode so we have greater flexibility with derives etc...

PR #2247, opened 2023-03-10
refactor(abigen): keep and use parsed spans

DaniPopes:
## Motivation

Keep spans around instead of unparsing into Strings

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


ethers-contract/ethers-contract-abigen/src/lib.rs
```diff
@@ -66,34 +66,34 @@ use std::{collections::HashMap, fmt, fs, io, path::Path};
...
+    pub method_aliases: HashMap<String, String>,
 
     /// Manually specified event name aliases.
-    event_aliases: HashMap<String, String>,
+    pub event_aliases: HashMap<String, String>,
 
     /// Manually specified error name aliases.
-    error_aliases: HashMap<String, String>,
+    pub error_aliases: HashMap<String, String>,
+
```
mattsse, line 85: why do we need pub access now?

mattsse (changes requested):
looks great.

but I want to get rid of Abigen pub fields. don't think they need to be mutable or do they?

DaniPopes:
They do for the builder, allows direct access to maps

mattsse:
since this is a public type, I'd rather have immutable getter functions.

DaniPopes:
Clippy unrelated, looks like a new lint dropped

gakonst (approved):

PR #2268, opened 2023-03-16
fix: accept ethlive as a chain name

CodeSandwich:
## Motivation

I'm coming from Foundry. The result of `cast chain` for Ethereum mainnet returns `ethlive`. This value is then rejected as a `--chain` argument for `forge verify-contract`, maybe other commands too. The correct value for `--chain` is `mainnet`.

## Solution

This PR adds support for parsing the `ethlive` name of a chain. This name was also used in dapptools, maybe also somewhere else, but it doesn't seem to be a bug on Foundry side to use it, it's just a less known value.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [x] Updated the changelog
-   [x] Breaking changes


ethers-core/src/types/chain.rs
```diff
@@ -47,6 +47,7 @@ pub type ParseChainError = TryFromPrimitiveError<Chain>;
 #[strum(serialize_all = "kebab-case")]
 #[repr(u64)]
 pub enum Chain {
+    #[strum(serialize = "ethlive")]
```
DaniPopes, line 50: strum is weird and you have to this

```suggestion
    #[strum(to_string = "mainnet", serialize = "ethlive")]
```

mattsse, line 50: this will change replace "mainnet" whith ethlive.

imo mainnet is more meaningful.

we could add this as alias, but dunno how with strum tbh.

mattsse (changes requested):
I'd prefer an alias, and a test for this.

DaniPopes (changes requested):

ethers-core/src/types/chain.rs
DaniPopes: yeah it's weird, I learnt this not too long ago: you can have only one "to_string" attr, which is the main one. Then aliases use serialize = ...
I'll open a pr to add that to the comments above

CodeSandwich: That's right, the proposed PR breaks the string representation. I've fixed it by adding another `serialize` parameter, it matches what other enum variants do. I've added a test, checking that, but I see that in #2270 you've solved the tests coverage once and for all, so I'll remove them.

CodeSandwich: Maybe it makes sense to switch multiple `serialize` parameters with implicitly the last one being the `to_string` representation with the explicit `to_string` parameters?

Issue #2038, opened 2023-01-10
Chain has Serialization/Deserialization mismatch


plotchy:
**Description**
Using serde_json to save serializations of Chains within files, and later deserializing the files back into Chain objects.

Ran into an issue with the Chain::BinanceSmartChain deserialization failing. Below is a test showing the issue:
```rust
#[tokio::test]
    async fn test_chain_deser() {
        let chain = Chain::Mainnet;
        let chain_str = serde_json::to_string(&chain).unwrap(); // "\"mainnet\""
        dbg!(&chain_str);
        let chain_deser: Chain = serde_json::from_str(&chain_str).unwrap(); // Mainnet
        dbg!(&chain_deser);
        assert_eq!(chain, chain_deser); // passes

        let chain = Chain::BinanceSmartChain;
        let chain_str = serde_json::to_string(&chain).unwrap(); // "\"bsc\""
        dbg!(&chain_str);
        let chain_deser: Chain = serde_json::from_str(&chain_str).unwrap(); // panic. "bsc" is not a valid chain
        assert_eq!(chain, chain_deser); // not reached
    }
```

Here is the output of the test:
```
running 1 test
[src/lib.rs:288] &chain_str = "\"mainnet\""
[src/lib.rs:290] &chain_deser = Mainnet
[src/lib.rs:295] &chain_str = "\"bsc\""
[...]

PR #2270, opened 2023-03-16
feat(core): roundtrip serde + to/from strings

DaniPopes:
## Motivation

Refs #2038, https://github.com/gakonst/ethers-rs/pull/2268#discussion_r1138931364

Doesn't fix it because you still need to replace kebab-case to snake_case

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### Unreleased
 
-- Add `other: OtherFields` to `TransactionReceipt` [#2209[](https://github.com/gakonst/ethers-rs/pull/2209)
+- Make `Chain` more round-trip friendly [#2270](https://github.com/gakonst/ethers-rs/pull/2270)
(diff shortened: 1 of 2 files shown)


CodeSandwich:
From tests I wrote when adding `ethlive` I've found out that the **last** `serialize` parameter is the one which is used when doing `to_string`. In this PR you're promoting the **first** one, which will break a lot of downstream.

gakonst:
Does this supersede #2268?

DaniPopes:
> From tests I wrote when adding `ethlive` I've found out that the **last** `serialize` parameter is the one which is used when doing `to_string`. In this PR you're promoting the **first** one, which will break a lot of downstream.

Yes. I don't consider it breaking since this is only changes the `Display`/`ToString` implementation and previous implementations can parse this "new" one because it was aliased. I also think it wasn't intended in the first place that the *last* item in the list is the selected one for Display.

> Does this supersede https://github.com/gakonst/ethers-rs/pull/2268?

No

Issue #2271, opened 2023-03-16
Abigen produces bindings with dependency issues


FrankieIsLost:
**Version**
439a0c7de0fb8569178d56933b000d8783805c2f

**Platform**
MacOS

**Description**
https://github.com/gakonst/ethers-rs/pull/2222 seems to have affected binding generation. `forge bind` generates bindings which produce the following errors 

```
error[E0432]: unresolved import `futures_util::try_join`
   |
35 | use futures_util::{lock::Mutex, try_join};
   |                                 ^^^^^^^^ no `try_join` in the root

error: cannot determine resolution for the macro `try_join`
    |
789 |             try_join!(self.resolve_field(ens_name, "avatar"), self.resolve_name(ens_name))?;
    |             ^^^^^^^^
    |
    = note: import resolution is stuck, try simplifying macro imports

    Checking ethers-contract-abigen v2.0.0 (https://github.com/gakonst/ethers-rs#439a0c7d)
```

Relatedly, should the generated `Cargo.toml` file pin ethers to a specific version?

DaniPopes:
How can I reproduce this? And where is the error occurring?

DaniPopes:
Oh I see, it breaks on building the repo, even if you cd into it and build it there it's fine, I'm not sure what this is.
Can you manually set `[package] ... resolver = "1"` in `Cargo.toml`?

PR #2272, opened 2023-03-16
fix(abigen): pin bindings to latest crate version

DaniPopes:
## Motivation

refs #2271
we should be switching off of git deps for downstream too anyway

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/multi.rs
+++ b/ethers-contract/ethers-contract-abigen/src/multi.rs
@@ -13,7 +13,11 @@ use std::{
 use toml::Value;
 
 /// The default ethers dependency to generate.
-const DEFAULT_ETHERS_DEP: &str = "ethers = { git = \"https://github.com/gakonst/ethers-rs\", default-features = false, features = [\"abigen\"] }";
+const DEFAULT_ETHERS_DEP: &str = concat!(
+    "ethers = { version = \"",
+    env!("CARGO_PKG_VERSION"),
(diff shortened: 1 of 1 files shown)


gakonst:
Will this still generate a binding that depends on git if there's another crate around? Should we make`forge bind` take a `--version` or `--git` argument to make this user-configured instead of trying to guess ourselves for all cases?

ethers-contract/ethers-contract-abigen/src/multi.rs
```diff
@@ -13,7 +13,11 @@ use std::{
 use toml::Value;
 
 /// The default ethers dependency to generate.
-const DEFAULT_ETHERS_DEP: &str = "ethers = { git = \"https://github.com/gakonst/ethers-rs\", default-features = false, features = [\"abigen\"] }";
+const DEFAULT_ETHERS_DEP: &str = concat!(
+    "ethers = { version = \"",
+    env!("CARGO_PKG_VERSION"),
+    "\", default-features = false, features = [\"abigen\"] }"
+);
```
mattsse, line 20: this will cause issues with foundry because foundry uses ethers master to generate bindings that then depend on crates.

so until we don't use ethers from crates in foundry, I recommend against this.

DaniPopes: makes sense, closing until then

PR #2275, opened 2023-03-17
fix: use to_string in mainnet chain variant

CodeSandwich:
## Motivation

A followup of https://github.com/gakonst/ethers-rs/pull/2268 in the context of https://github.com/gakonst/ethers-rs/pull/2270. I didn't trust myself with the serde and strum configurations so I've added a test checking if they match.

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/types/chain.rs
+++ b/ethers-core/src/types/chain.rs
@@ -616,4 +616,13 @@ mod tests {
             }
         }
     }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):
the test is already checked in rountrip_string but that's fine

PR #2280, opened 2023-03-19
feat(solc): improve error diagnostic

mattsse:
## Motivation
Closes https://github.com/foundry-rs/foundry/issues/4594

checks if the file's version is valid

cc @aathan

## Solution
add check that the requirement is satisfied by at least one known version

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


mattsse:
clippy false positive

PR #2281, opened 2023-03-19
Contracts chapter

0xJepsen:
I noticed that the mdbook was missing some chapters. I followed the existing format from the previous chapters and linked to the relevant examples. I know that the example in `contracts/creating-instances.md` is missing, and I am unfamiliar if the examples are up to date. If they are out of date, if someone could point me in the direction of the breaking changes so that i can update them id be happy to take a stab. 

Contributes to #2282


gakonst:
Merging in the interest of having something but we should iterate on this^^

PR #2283, opened 2023-03-20
fix(core): re-export CallLogFrame from geth types

jqphu:
## Motivation

The CallLogFrame was not being exported from geth which made it
difficult to store all the logs using the ethers type.

## Solution

Export CallLogFrame so it's accessible publicly via `ethers::types`.
Also export the internal fields.

### DIFF
--- a/ethers-core/src/types/trace/geth.rs
+++ b/ethers-core/src/types/trace/geth.rs
@@ -4,7 +4,7 @@ mod noop;
 mod pre_state;
 
 pub use self::{
-    call::{CallConfig, CallFrame},
+    call::{CallConfig, CallFrame, CallLogFrame},
     four_byte::FourByteFrame,
     noop::NoopFrame,
     pre_state::{PreStateConfig, PreStateFrame},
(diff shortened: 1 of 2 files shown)


mattsse (approved):

Cargo.toml
```diff
@@ -81,11 +81,11 @@ ethers-derive-eip712 = { version = "2.0.0", path = "ethers-core/ethers-derive-ei
...
-futures = { version = "0.3.27", default-features = false, features = ["std"] }
-futures-core = "0.3.27"
-futures-util = "0.3.27"
-futures-executor = "0.3.27"
-futures-channel = "0.3.27"
+futures = { version = "0.3.25", default-features = false, features = ["std"] }
+futures-core = "0.3.25"
+futures-util = "0.3.25"
+futures-executor = "0.3.25"
+futures-channel = "0.3.25"
```
DaniPopes, line 88: what is this for?

DaniPopes (changes requested):

Cargo.toml
jqphu: Wrong fork.. sorry! Removing.

jqphu:
Was just testing this locally and realized I didn't make the internal fields public. Fixed and good to go now!

prestwich (approved):

PR #2284, opened 2023-03-21
Prestwich/gas escalator dangle

### LINKED ISSUE #2269: Gas escalator middleware can lead to file descriptor leaks
**Version**
List the versions of all `ethers-rs` crates you are using. The easiest way to get
this information is using `cargo-tree`.

2.0.0

**Platform**
The output of `uname -a` (UNIX), or version and 32 or 64-bit (Windows)

> Linux ip-10-0-67-197 5.15.0-1030-aws #34~20.04.1-Ubuntu SMP Tue Jan 24 15:16:46 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

**Description**
Enter your issue details here.
One way to structure the description:

I am creating the provider with the gas escalator dynamically, to dynamically connect to chain endpoints. For every gas escalator middleware creation `GasEscalatorMiddleware::new`, it spawns an asynchronous task to poll the gas price:

https://github.com/gakonst/ethers-rs/blob/304fe0988d23302694a63da47ce683af19305357/ethers-middleware/src/gas_escalator/mod.rs#L135-L156

After some hours, we found that this likely caused the file descriptor leaks when we create `GasEscalatorMiddleware` for each request (even if we are done processing the request, the asynchronous task is still running).

One way is for us to reuse gas escalator, instead of creating a new one every time. Or `ethers-rs` exposes a way to shut down this asynchronous task.

Would appreciate any feedback. Thanks!

prestwich:
Fixes #2269 

## Motivation

Task spawn was causing dangling tasks when GasEscalator dropped, as no messaging notified the task to close

## Solution

GasEscalatorMiddleware is split into 3 things:
- `GasEscalatorMiddleware`: a simple wrapper around an arc for safe cloning
- `GasEscalatorMiddlewareInternal`: Holds the inner mware, the shutdown channel, and a ref to the tx vec
- `EscalatorTask`: the running escalation task. Periodically attempts to escalate the transactions

What I fixed:
- dangling tasks
- a very long mutex guard hold
- gas escalator now always cheaply cloneable
- remove clone bounds from some functions

What I didn't fix:
- if the escalation task crashes, the mware will **still** become unusable, but not inform you of this or error. This is not a change from previous behavior, however, it sucks. It would require a much larger fix tho

What this introduces
- if the user drops all middleware, escalation will cease. this seems like good, expected behavior

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Updated the changelog
-   [ ] Breaking changes


gakonst (approved):
One idea I had is that maybe we pull `SignerMiddleware` to `ethers-signers` and we otherwise pull out `ethers-middleware` to a separate repo, as these are starting to be a maintenance overhead and are not part of the "core" evolving suite.

PR #2290, opened 2023-03-21
fix(solc): features

DaniPopes:
## Motivation

removed by accident during refactoring

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (approved):
I think this is the reason I saw some issues when running foundry locally,

PR #2291, opened 2023-03-21
fix: ethers-contract circular dep on ethers-signers

gakonst:
As title, ethers-contract depended on signers which messes up the auto-release process, fixes by removing the dep and pulling the EIP712 tests which require both signers + contract + eip712 derive to the top level crate tests


gakonst:
In a bit of a rush to fix the release flow and start automating it, so merging.

PR #2294, opened 2023-03-21
fix: enable doc_cfg feature for docsrs

DaniPopes:
## Motivation

This doesn't actually fix the doc problem, since it has to be enabled in `generic-array`, but it prevents this happening in the future for our crates

## Solution

Adds `#![cfg_attr(docsrs, feature(doc_cfg))]` to all workspace crates (minus examples)

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (approved):
good idea!

PR #2314, opened 2023-03-29
perf: early return in EthCodec derive

DaniPopes:
## Motivation

Currently we go straight into decoding which can become expensive when called continuously from a big enum, eg foundry HEVM or HardhatConsole. We have the ParamType to decode so we can calculate very cheaply a minimum bound in bytes to check against

## Solution

- Add a minimum bound check for the given bytes to early return for in `derive(EthCodec)`
- (drive by) Fix tuple implementations of `Tokenizable`
  - these would panic if given a `Token::Tuple` of the wrong length
  - also we have ownership of the `Vec` so `into_iter` is better than `drain`

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-contract/ethers-contract-derive/src/abi_ty.rs
+++ b/ethers-contract/ethers-contract-derive/src/abi_ty.rs
@@ -83,16 +83,16 @@ pub fn derive_tokenizeable_impl(input: &DeriveInput) -> Result<TokenStream, Erro
             // can't encode an empty struct
             // TODO: panic instead?
             quote! {
-                #ethers_core::abi::Token::Tuple(Vec::new())
+                #ethers_core::abi::Token::Tuple(::std::vec![])
             },
         ),
         _ => {
+            let err_format_string = format!("Expected {params_len} tokens, got {{}}: {{:?}}");
             let from_token = quote! {
                 if let #ethers_core::abi::Token::Tuple(tokens) = token {
                     if tokens.len() != #params_len {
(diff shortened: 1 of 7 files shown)


ethers-contract/ethers-contract-derive/src/abi_ty.rs
```diff
@@ -205,45 +201,44 @@ fn tokenize_enum<'a>(
...
+                into_tokens.extend(quote! {
+                    #enum_name::#var_ident(element) => element.into_token(),
+                });
+            }
+            _ => {
+                return Err(Error::new(
+                    variant.span(),
+                    "EthAbiType cannot be derived for enum variants with multiple fields",
+                ))
+            }
```
DaniPopes, line 235: This was written weirdly, with also an unreachable block (length == 1 guarantees an item, but an `else` block was present)

ethers-contract/ethers-contract-derive/src/abi_ty.rs
```diff
@@ -205,45 +201,44 @@ fn tokenize_enum<'a>(
 
     let mut into_tokens = TokenStream::new();
     let mut from_tokens = TokenStream::new();
-    for (idx, variant) in variants.into_iter().enumerate() {
+    let last = variants.size_hint().1.unwrap().saturating_sub(1);
```
DaniPopes, line 204: consume in the last call instead of cloning

mattsse (approved):
lgtm

PR #2316, opened 2023-03-30
add scroll alpha testnet

### LINKED ISSUE #2310: Add support for `Scroll Alpha Testnet`.
chain-id - 534353
rpc:https://alpha-rpc.scroll.io/l2
api: https://blockscout.scroll.io/api
explorer: https://blockscout.scroll.io/

salman01zp:
## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

---
- closes #2310


ethers-core/src/types/chain.rs
```diff
@@ -412,6 +414,10 @@ impl Chain {
                 ("https://blockscout.com/xdai/mainnet/api", "https://blockscout.com/xdai/mainnet")
             }
 
+            ScrollAlphaTestnet => {
+                ("https://blockscout.scroll.io/api", "https://blockscout.scroll.io/")
+            }
```
mattsse, line 419: is this right?

kinda weird, because this looks more like the mainnet URL?

salman01zp: yeah.. but it is for scroll-alpha-testnet [https://blockscout.scroll.io/](https://blockscout.scroll.io/)

You can also confirm with the scroll official page [here](https://scroll.io/alpha)

mattsse: very cursed imo

will cause issues once they replace it with mainnet or they add a `mainnet` to the path or domain, which would be even more confusing -.-

but looks like this is correct (at least for now), so lgtm

mattsse (approved):

prestwich:
Please rebase onto latest main and then we can merge 👍

PR #2317, opened 2023-03-31
chore: use ethers 2 by default when generating bindings

mattsse:
## Motivation
no that we have auto-releases, we can switch to using crates dep by default for generated bindings.

we can also depend on ethers from crates in foundry from now

wdyt @DaniPopes (harcoding the current major release is probably the easiest solution)

cc @leonardoalt 

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


DaniPopes:
Yeah I think that's fine if we switch foundry off of git as well

PR #2279, opened 2023-03-18
feat: syn 2.0, merge eip712 crate into ethers-contract-derive

DaniPopes:
## Motivation

https://github.com/dtolnay/syn/releases/tag/2.0.0

Blocked by:
- ~prettyplease (abigen formatting)~

Haven't updated yet (doesn't block):
- async-trait
- auto_impl
- clap_derive (from criterion and svm-rs)
- ctor (dev; from pretty_assertions)
- derive_more (from ethereum-types)
- futures-macro
- impl-trait-for-tuples (from ethbloom)
- num_enum_derive
- open-fastrlp-derive
- parity-scale-codec-derive
- phf_macros (from phf <- solang-parser)
- pin-project-internal (from pin-project)
- proc-macro-error (from auto_impl, clap_derive; seen above)
- rlp-derive
- serde_derive
- serial_test_derive (dev)
- signature_derive
- strum_macros
- synstructure
- thiserror-impl
- tokio-macros
- tracing-attributes
- tracing-test-macro
- wasm-bindgen-backend
- wasm-bindgen-macro-support
- zeroize_derive

## Solution

Fix breaking changes, mainly:
- `input.parse_terminated` from turbofish to argument
- attribute parsing

Additionally, moves `ethers-derive-eip712` to `ethers-contract-derive` due to duplicated code and useless duplicated dependencies on ethers-core. Also ~removes~ deprecates the feature since it doesn't gate dependencies anymore.

## TODO

- [x] revert temp prettyplease fix
- [x] fix eip712 test: `ethers_contract::it::contract::test_derive_eip712`

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Breaking changes

PR #2318, opened 2023-03-31
fix: use installed solc version

mattsse:
## Motivation
use the installed solc bin 

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


DaniPopes:
the path is wrong. that test is getting removed anyway in #2279

PR #2325, opened 2023-04-05
fix(etherscan)!: change the `GasOracle` fields to be `f64`

shekohex:
## Motivation

Closes  #2324

## Solution

This updates the `GasOracle`'s `{safe_gase_price,propose_gas_price,fast_gas_price}` from `u64` to `f64`

BREAKING CHANGE: This considered a public api breaking change, since `f64` and `u64` is not compatible with each other.

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Breaking changes

PR #2326, opened 2023-04-05
fix(middleware): no need to multiply again with `GWEI_TO_WEI_U256`

shekohex:
## Motivation

While working on #2325 I did not notice that `parse_units` will return the correct value and will be no need to multiply the result again with `GWEI_TO_WEI_U256` (`1_000_000_000`) extra zeros! which makes the gas calculations insanely incorrect!

## Solution

Removing the multiplication process solves that issue.

Edit: actually, there was a method called `from_gwei_f64` that we can use for that.

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Breaking changes


drewstone (approved):
LG

prestwich (approved):

PR #2327, opened 2023-04-05
feature: very permissive etherscan gwei deser

prestwich:
Extremely permissive deserializer for etherscan gas units

## Motivation

More complete fix to #2326 #2325 

## Solution

Introduce a deser that performs the following in order:
- checks for a JSON number, and scales it by WEI_PER_GWEI
- checks for a string, then tries to parse it as a float number of gwei

change the etherscan response object to have `U256` responses that are ALWAYS in `wei` rather than in f64 gwei

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Breaking changes

### DIFF
--- a/ethers-etherscan/src/gas.rs
+++ b/ethers-etherscan/src/gas.rs
@@ -1,25 +1,55 @@
 use crate::{Client, EtherscanError, Response, Result};
-use ethers_core::types::U256;
+use ethers_core::{types::U256, utils::parse_units};
 use serde::{de, Deserialize, Deserializer};
 use std::{collections::HashMap, str::FromStr};
+
+const WEI_PER_GWEI: u64 = 1_000_000_000;
 
 #[derive(Deserialize, Clone, Debug)]
 #[serde(rename_all = "PascalCase")]
 pub struct GasOracle {
-    #[serde(deserialize_with = "deserialize_number_from_string")]
-    pub safe_gas_price: f64,
-    #[serde(deserialize_with = "deserialize_number_from_string")]
-    pub propose_gas_price: f64,
-    #[serde(deserialize_with = "deserialize_number_from_string")]
-    pub fast_gas_price: f64,
+    /// Safe Gas Price in wei
+    #[serde(deserialize_with = "deser_gwei_amount")]
+    pub safe_gas_price: U256,
+    /// Propose Gas Price in wei
+    #[serde(deserialize_with = "deser_gwei_amount")]
+    pub propose_gas_price: U256,
+    /// Fast Gas Price in wei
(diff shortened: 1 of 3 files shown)


ethers-etherscan/src/gas.rs
```diff
@@ -1,27 +1,57 @@
...
+    pub suggested_base_fee: U256,
+    /// Gas Used Ratio
     #[serde(deserialize_with = "deserialize_f64_vec")]
     #[serde(rename = "gasUsedRatio")]
     pub gas_used_ratio: Vec<f64>,
 }
 
+// This function is used to deserialize a string or number into a U256 with an
+// amount of gwei. If the contents is a number, deserialize it. If the contents
+// is a string, attempt to deser as first a decimal f64 then a decimal U256.
```
shekohex, line 35:
```suggestion
// This function deserializes a string or number into a U256 with an amount of gwei. 
// If the content is a number, it is deserialized. If the content is a string, we attempt to deserialize it first // as a decimal f64 and then as a decimal U256.
```

ethers-etherscan/src/gas.rs
```diff
@@ -1,27 +1,57 @@
 use crate::{Client, EtherscanError, Response, Result};
-use ethers_core::types::U256;
+use ethers_core::{types::U256, utils::parse_units};
 use serde::{de, Deserialize, Deserializer};
 use std::{collections::HashMap, str::FromStr};
 
+const WEI_PER_GWEI: u64 = 1_000_000_000;
```
shekohex, line 6: maybe use `super::GWEI_TO_WEI` instead?
```suggestion
```

ethers-etherscan/src/gas.rs
```diff
@@ -1,27 +1,57 @@
...
+{
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum StringOrInt {
+        Number(u64),
+        String(String),
+    }
+
+    match StringOrInt::deserialize(deserializer)? {
+        StringOrInt::Number(i) => Ok(U256::from(i) * WEI_PER_GWEI),
```
shekohex, line 48:
```suggestion
        StringOrInt::Number(i) => Ok(U256::from(i) * super::GWEI_TO_WEI),
```

shekohex (changes requested):
Looks good!

ethers-etherscan/src/gas.rs
prestwich: doesn't exist in this package, that's in the middleware, not in the etherscan

shekohex: Ah I see, gotcha!

shekohex (approved):

PR #2330, opened 2023-04-06
chore: update deny config

DaniPopes:
## Motivation

- Remove outdated exceptions and `[[license.clarify]]`
- Make config more strict for wildcard and git dependencies
  - bad practice, git dependencies are not allowed when published so doesn't hurt adding an extra check
- Add `MPL-2.0` to license allow list since it generates a warning for being copy-left
  - Looked to other crates using the `webpki-roots` crate: <https://github.com/hyperium/tonic/blob/master/deny.toml>

The only remaining warnings are for duplicate crates

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/.github/workflows/deps.yml
+++ b/.github/workflows/deps.yml
@@ -8,27 +8,14 @@ on:
         branches: [master]
         paths: [Cargo.lock]
 
-env:
-    RUSTFLAGS: -D warnings
-    CARGO_TERM_COLOR: always
-
 concurrency: deps-${{ github.head_ref || github.run_id }}
 
 jobs:
     deny:
-        name: deny (${{ matrix.checks }})
+        name: deny
         runs-on: ubuntu-latest
-        strategy:
-            fail-fast: false
-            matrix:
-                checks:
-                    - advisories
-                    - bans licenses sources
-
-        continue-on-error: ${{ matrix.checks == 'advisories' }}
-
         steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/deps.yml
```diff
@@ -8,27 +8,14 @@ on:
...
-env:
-    RUSTFLAGS: -D warnings
-    CARGO_TERM_COLOR: always
-
 concurrency: deps-${{ github.head_ref || github.run_id }}
 
 jobs:
     deny:
-        name: deny (${{ matrix.checks }})
+        name: deny
```
mattsse, line 15: right, we def don't need a matrix here if we use `check all` which does both

mattsse (approved):

.github/workflows/deps.yml
DaniPopes: A thing I noticed about this is that if `check advisories` fails with an error, it will continue regardless, but then if the next matrix step succeeds the entire job succeeds, silencing the advisory..

DaniPopes: I initially took it from [reth](https://github.com/paradigmxyz/reth/blob/main/.github/workflows/deny.yml), should I update it there too?

mattsse: please do, ty!

PR #2331, opened 2023-04-06
feat(eip712): add raw_salt attribute

### LINKED ISSUE #2321: EIP712 `salt` is always keccak256 hashed
**Version**
main branch latest today

**Platform**
all

**Description**
- I've been working on a software project where we use both JS and Rust to generate valid EIP712 signatures and we're not using a contract to verify the signature but our own nodes (similar to how Farcaster does it).
- Inspired by them, I've been using the "salt" domain separator: https://github.com/farcasterxyz/hub-monorepo/blob/59920f9716838a19749f41c4f3c7f9ac979cac07/packages/utils/src/crypto/eip712.ts#L11
- But it seems that ethers-rs (in contrast to ethers-js) is always hashing the salt value using keccak256
  - JS: https://github.com/ethers-io/ethers.js/blob/v5.2/packages/hash/src.ts/typed-data.ts#L70
  - Rust: https://github.com/gakonst/ethers-rs/blob/64b7f1ef1ac71fefcacc44d8ff1ddfcb8e6b5417/ethers-core/src/types/transaction/eip712.rs#L383
  - Rust2: https://github.com/gakonst/ethers-rs/blob/404422e883b7a7fea34f4d8fbb3548cabb68f9c3/ethers-core/ethers-derive-eip712/src/lib.rs#L18
- And so now this isn't ideal for me. Because I'd rather just input a 32 bytes long sequence instead of having implicit hashing.
- So I've been thinking of just proposing to remove the line in the Rust implementation, but then that'd break the interface for existing users
- So what's an elegant solution for passing in 32 bytes that aren't later hashed?

DaniPopes:
## Motivation

Fixes #2321

Based on #2279

## Solution

- feat(eip712): add raw_salt attribute (9e923266d8e18c5b94444d10997711fad5a86ead)

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


DaniPopes:
Ready for review after #2279

prestwich (changes requested):
LGTM, needs rebase and conflict resolved

PR #2332, opened 2023-04-06
chore: dont retain errors

mattsse:
## Motivation
if we `compile_exact(file)` we should not filter errors because this obfuscates failed compile runs
discovered via:
https://github.com/foundry-rs/foundry/issues/4467

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


prestwich (commented):
The code change looks fine, but I don't have the context to review its impact on things. cc @gakonst

gakonst (approved):
Ah yeh the error may be in another file which our file imports. Good catch.

PR #2333, opened 2023-04-06
Add support for ABI encoded keccak256

crypdoughdoteth:
## Motivation

A utility to simplify code for developers hashing data to be used on-chain. Is equivalent to keccak25(abi.encode(...)) on-chain. 

## Solution

Abi encoding the arguments of a function before hashing them and calling .into() as a combinator on the [u8;32] we get back from the keccak256 function.  

## PR Checklist

-   [ +] Added Tests
-   [ +] Added Documentation
-   [ -] Breaking changes

### DIFF
--- a/ethers-core/src/utils/hash.rs
+++ b/ethers-core/src/utils/hash.rs
@@ -1,8 +1,8 @@
 //! Various utilities for manipulating Ethereum related data.
 
+use crate::abi::{self, ethabi::Bytes};
 use ethabi::ethereum_types::H256;
 use tiny_keccak::{Hasher, Keccak};
-
 /// Hash a message according to [EIP-191] (version `0x01`).
 ///
 /// The final message is a UTF-8 string, encoded as follows:
@@ -29,7 +29,6 @@ pub fn hash_message<T: AsRef<[u8]>>(message: T) -> H256 {
 /// Compute the Keccak-256 hash of input bytes.
 ///
 /// Note that strings are interpreted as UTF-8 bytes,
-// TODO: Add Solidity Keccak256 packing support
 pub fn keccak256<T: AsRef<[u8]>>(bytes: T) -> [u8; 32] {
     let mut output = [0u8; 32];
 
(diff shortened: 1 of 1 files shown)


ethers-core/src/utils/hash.rs
```diff
@@ -40,6 +40,13 @@ pub fn keccak256<T: AsRef<[u8]>>(bytes: T) -> [u8; 32] {
     output
 }
 
+/// matches keccak256(abi.encode(....)) pattern in solidity
+///
+/// accepts a borrowed array of type abi::Token
+fn solidity_keccak256(args: &[abi::Token]) -> Bytes {
```
mattsse, line 46: the naming might cause confusion, as a developer might now wonder what the difference is between the "keccak" function and this function.

maybe `keccak256_tokens`?

wdyt @prestwich

ethers-core/src/utils/hash.rs
```diff
@@ -40,6 +40,13 @@ pub fn keccak256<T: AsRef<[u8]>>(bytes: T) -> [u8; 32] {
     output
 }
 
+/// matches keccak256(abi.encode(....)) pattern in solidity
+///
```
mattsse, line 44: Please mention that this is a convenience function.

mattsse (changes requested):

ethers-core/src/utils/hash.rs
crypdoughdoteth: I agree that the name could confuse people. My motivation for naming the function with "solidity" prepended is to match the naming convention used in the Ethers.js package. Is it worthwhile to preserve the naming convention? 
image

prestwich: the ethers.js package uses this for `keccak256(abi.encodePacked(...))`, so this will result in a different output from ethers.js. I think `solidityKeccak256` is a poor name in the first place, as it doesn't communicate anything

is there a reason this is a util function and not a provided method on one of the tokenization traits?

prestwich: I'm not convinced it's a worthwhile inclusion. If it were, i'd probably name it `encoded_hashed` or similar to signifiy that there are 2 actions being performed

crypdoughdoteth: Should I add packing support (with the relevant helpers) + move to trait + rename? I think having both packed and non-packed options could be helpful: non-packed is used when there are multiple dynamic types because it prevents potential hash collisions (with the tradeoff of a little extra gas)

gakonst:
Would like to close as unneeded - this is easy enough that one can do themselves, and we can tackle packed stuff separately

PR #2335, opened 2023-04-07
fix(abigen): SolStruct derives edge case

DaniPopes:
## Motivation

From: <https://t.me/ethers_rs/27028>

A struct that contains an array of other structs doesn't add derives when it should. This is because the check [here](https://github.com/gakonst/ethers-rs/blob/1dd35458695e2b967d4c2ff278cc8d0eaadcc9fb/ethers-contract/ethers-contract-abigen/src/util.rs#L235-L239) is incorrect and shouldn't exist.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (approved):
nice find,
lgtm

gakonst:
LGTM - needs lint @DaniPopes

PR #2339, opened 2023-04-10
docs: update some doctest examples

DaniPopes:
## Motivation

Cleaned up doctests with `tokio::main` and `no_run`, and etherscan tests which all had redundant client initializations which are not the focus of the example

I don't believe running async tests in doctests is a great practice

Breaking change in `etherscan::Client::get_ether_balance_multi` because slice of references is not good

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


prestwich (approved):

prestwich:
We have another small breaking change in ethers-etherscan already from #2327

PR #2340, opened 2023-04-10
fix: poll stream until pending

mattsse:
## Motivation
streams only register the waker if they're pending
Returning "pending" when the underlying stream is still ready does not ensure that the rt will poll again. 

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-providers/src/rpc/pubsub.rs
+++ b/ethers-providers/src/rpc/pubsub.rs
@@ -94,16 +94,18 @@ where
             return Poll::Ready(next_element)
         }
 
-        let this = self.project();
-        match futures_util::ready!(this.rx.poll_next(ctx)) {
-            Some(item) => match serde_json::from_str(item.get()) {
-                Ok(res) => Poll::Ready(Some(res)),
-                Err(err) => {
-                    error!("failed to deserialize item {:?}", err);
-                    Poll::Pending
-                }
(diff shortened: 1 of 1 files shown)


ethers-providers/src/rpc/pubsub.rs
```diff
@@ -94,16 +94,18 @@ where
...
-            Some(item) => match serde_json::from_str(item.get()) {
-                Ok(res) => Poll::Ready(Some(res)),
-                Err(err) => {
-                    error!("failed to deserialize item {:?}", err);
-                    Poll::Pending
-                }
-            },
-            None => Poll::Ready(None),
+        let mut this = self.project();
+        loop {
```
prestwich, line 98: isn't this loop a no-op, as the return always ends it on the first loop?

prestwich: nvm i am misreading the code

mattsse: This continues the poll `rx` on `Err(err)` instead of returning Pending, and ensures that the underlying stream is completely drained or returns a valid element.

prestwich (approved):
CI failures are unrelated

PR #2293, opened 2023-03-21
fix: test features

DaniPopes:
## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

PR #2341, opened 2023-04-10
chore: minify test json files

DaniPopes:
## Motivation

A windows test broke in #2293 due to actual `\r\n` vs expected `\n`.

## Solution

Reproduce with:

```sh
minify () {
        for f
        do
                out="$(jq -rc . "$f" 2>&1)" 
                if [ $? -ne 0 ]
                then
                        echo "$out" >&2
                        continue
                fi
                echo "writing $f" >&2
                printf "%s\n" "$out" > "$f"
        done
}

minify **/*.json
git restore ethers-addressbook
```

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (commented):
supportive of minifying all json files

looks like this caused some issues

DaniPopes:
jq can't handle utf-8 apparently

PR #2343, opened 2023-04-12
docs: fix git-cliff tag filtering

DaniPopes:
## Motivation

The `git-cliff` config `ignore_tags` was ignoring *all* `^ethers-.*` tags, which includes the base `ethers-vX.Y.Z`.

## Solution

Exclude all `ethers-*` tags except for `ethers-vX.Y.Z`, and trim the name in the template

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@ which is configured [here](./cliff.toml).
 
 Please do not manually edit this file.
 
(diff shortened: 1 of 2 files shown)


gakonst:
good catch

PR #2350, opened 2023-04-14
fix: correctly handle Trezor One support

prusnak:
- Trezor version check works correctly only for Trezor Model T which uses version numbers 2.x.x
- this PR changes the logic so also Trezor One versions are properly handled

### DIFF
--- a/ethers-signers/src/trezor/app.rs
+++ b/ethers-signers/src/trezor/app.rs
@@ -35,7 +35,9 @@ pub struct TrezorEthereum {
     pub(crate) address: Address,
 }
 
-const FIRMWARE_MIN_VERSION: &str = ">=2.4.2";
+// we need firmware that supports EIP-1559 and EIP-712
+const FIRMWARE_1_MIN_VERSION: &str = ">=1.11.1";
+const FIRMWARE_2_MIN_VERSION: &str = ">=2.5.1";
 
 // https://docs.trezor.io/trezor-firmware/common/communication/sessions.html
 const SESSION_ID_LENGTH: usize = 32;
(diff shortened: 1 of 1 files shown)


ethers-signers/src/trezor/app.rs
```diff
@@ -73,12 +75,25 @@ impl TrezorEthereum {
...
+        let mut min_version;
 
-        // Enforce firmware version is greater than FIRMWARE_MIN_VERSION
+        match version.major {
+            1 => {
+                min_version = FIRMWARE_1_MIN_VERSION;
+            }
+            2 => {
+                min_version = FIRMWARE_2_MIN_VERSION;
+            }
```
DaniPopes, line 87:
```suggestion
        let min_version = match version.major {
            1 => FIRMWARE_1_MIN_VERSION,
            2 => FIRMWARE_2_MIN_VERSION,
```

ethers-signers/src/trezor/app.rs
```diff
@@ -73,12 +75,25 @@ impl TrezorEthereum {
     }
 
     fn check_version(version: String) -> Result<(), TrezorError> {
-        let req = semver::VersionReq::parse(FIRMWARE_MIN_VERSION)?;
         let version = semver::Version::parse(&version)?;
+        let mut min_version;
```
DaniPopes, line 79:
```suggestion
```

DaniPopes (approved):
lgtm, idiomatic nit

ethers-signers/src/trezor/app.rs
prusnak: addressed in dd108acb51141e98cfc734d7ebe2e2ed2ebbcd27

prestwich (approved):

Issue #2346, opened 2023-04-13
is_local_endpoint utility function should be working for every link-local


lucaprov1989:
**Version**
(https://github.com/gakonst/ethers-rs/commit/9a5b31a0a2302480080ffe34c9e1f4c957d0fd3d) (master)

**Platform**
MacOs 2.3 GHz 8-Core Intel Core i9

**Description**
Issue linked to foundry issue [here](https://github.com/foundry-rs/foundry/issues/4658)

The `is_local_endpoint` does a pattern match on the url as string missing all possible IPV4 link-local as `192.168.0.123`

I would love to take this as mentioned in the foundry issue.

Would probably change the approach using the IPV4 is_link_local function [here](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.is_link_local).

Any suggestion on how to deal with custom domains configured in etc/hosts is accepted.

gakonst:
Go for it :)

Evalir:
Closeable?

PR #2351, opened 2023-04-14
fix: ethers_providers::is_local_endpoint with rust matching pattern on URL::parse

lucaprov1989:
Related to #2346 and foundry issue [#4658](https://github.com/foundry-rs/foundry/issues/4658)

## Motivation

`ethers_providers::is_local_endpoint` function would only work with `localhost` and `127.0.0.1` strings.

## Solution

From substring matching to rust pattern matching based on `URL::parse` result.

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-providers/src/rpc/provider.rs
+++ b/ethers-providers/src/rpc/provider.rs
@@ -1476,7 +1476,6 @@ mod tests {
     use std::path::PathBuf;
 
     use super::*;
-
     use crate::Http;
     use ethers_core::{
         types::{
--- a/ethers-providers/src/rpc/provider.rs
+++ b/ethers-providers/src/rpc/provider.rs
@@ -1447,7 +1447,7 @@ impl ProviderExt for Provider<HttpProvider> {
 /// assert!(is_local_endpoint("http://169.254.0.0:8545"));
 /// assert!(is_local_endpoint("http://127.0.0.1:8545"));
 /// assert!(!is_local_endpoint("http://206.71.50.230:8545"));
-/// assert!(!is_local_endpoint("blabla"));
+/// assert!(!is_local_endpoint("havenofearlucishere"));
 /// ```
 #[inline]
 pub fn is_local_endpoint(endpoint: &str) -> bool {
--- a/ethers-providers/src/rpc/provider.rs
+++ b/ethers-providers/src/rpc/provider.rs
@@ -1447,6 +1447,8 @@ impl ProviderExt for Provider<HttpProvider> {
 /// assert!(is_local_endpoint("http://169.254.0.0:8545"));
 /// assert!(is_local_endpoint("http://127.0.0.1:8545"));
 /// assert!(!is_local_endpoint("http://206.71.50.230:8545"));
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm,

would appreciate a smol sanity test for this

lucaprov1989:
> lgtm,
> 
> would appreciate a smol sanity test for this

You want one more unit test? I thought doctest were enough?

mattsse (approved):
ah right, 

lgtm

ethers-providers/src/rpc/provider.rs
```diff
@@ -1443,13 +1444,25 @@ impl ProviderExt for Provider<HttpProvider> {
...
 /// ```
 #[inline]
-pub fn is_local_endpoint(url: &str) -> bool {
-    url.contains("127.0.0.1") || url.contains("localhost")
+pub fn is_local_endpoint(endpoint: &str) -> bool {
+    if let Ok(url) = Url::parse(endpoint) {
+        if let Some(host) = url.host() {
+            match host {
+                Host::Domain("localhost") => return true,
+                Host::Ipv4(ipv4) => return ipv4 == Ipv4Addr::LOCALHOST || ipv4.is_link_local(),
```
DaniPopes, line 1458: What about `ipv4.is_private()` and `ipv4.is_loopback()`?

ethers-providers/src/rpc/provider.rs
```diff
@@ -1443,13 +1444,25 @@ impl ProviderExt for Provider<HttpProvider> {
...
+                // skipping IPV6 check
+                _ => return false,
```
DaniPopes, line 1460: Any reason not to include the same for `Ipv6Addr`? (`ipv6.is_loopback()`)

mattsse: good point loopback is available for both variants

lucaprov1989: I did not support Ipv6 addresses since it looked like they were not being used where this function is being called.

If we need I can always add.

@mattsse should I?

ethers-providers/src/rpc/provider.rs
lucaprov1989: Added with [enabling also loopback and private addresses](https://github.com/gakonst/ethers-rs/pull/2351/commits/5935fc840849692a5fb7beba61cdda28833f1d0a)

ethers-providers/src/rpc/provider.rs
lucaprov1989: Added with [supporting also ipv6](https://github.com/gakonst/ethers-rs/pull/2351/commits/cad99e713d1aae97b57db6e9f910286537e0e053)

lucaprov1989:
@mattsse @DaniPopes all good?

mattsse (approved):
lgtm!

DaniPopes (approved):

prestwich (approved):

PR #2355, opened 2023-04-16
Add nameless decode to BaseContract

halo3mic:
## Motivation

[`BaseContract`](https://github.com/gakonst/ethers-rs/blob/a54eec73eacd2a17ff62b527d35c230a4795dae6/ethers-contract/src/base.rs#L18) provides various methods for decoding input data: `decode`, `decode_raw`, `decode_with_selector` and `decode_with_selector_raw`. 
These methods take a method name as an argument, which could be omitted as it is already provided through input-data.

## Solution

Add new methods that decode input-data without requiring method-name as a parameter - `decode_input_raw` and `decode_input`.

Existing decode methods (`decode`, `decode_raw`, `decode_with_selector` and `decode_with_selector_raw`) lose functionality with the introduction of nameless decode methods. They could be removed unless I missed a viable utility or should be left in for backward compatibility.

## PR Checklist

- [x] Added nameless decode methods to BaseContract
- [x] Add test
- [ ] Remove existing decode methods?

### DIFF
--- a/ethers-contract/src/base.rs
+++ b/ethers-contract/src/base.rs
@@ -157,6 +157,27 @@ impl BaseContract {
         decode_function_data(function, bytes, true)
     }
 
+    /// Decodes the provided ABI encoded input bytes
+    ///
+    /// Returns a [`Token`] vector, which lets you decode function arguments dynamically
+    /// without knowing the return type.
+    pub fn decode_input_raw<T: AsRef<[u8]>>(
+        &self,
+        bytes: T,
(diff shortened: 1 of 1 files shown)


prestwich:
needs formatting with `cargo +nightly fmt`

prestwich (changes requested):
concept ACK. See comment re rustfmt

prestwich (approved):
examples failure is unrelated

PR #2359, opened 2023-04-18
fix(core): fix decimal string IntOrHex parsing

Rjected:
## Motivation

Previously, we would parse decimal strings like `"100000"` as hex, because they did not match a raw `serde_json::Number` variant in `IntOrHex`. This caused problems when parsing genesis fields with `from_int_or_hex`, for example:
```json
            "alloc": {
                "0x3E951C9f69a06Bc3AD71fF7358DbC56bEd94b9F2": {
                  "balance": "1000000000000000000000000000"
                },
             }
``` 

This would parse `"1000000000000000000000000000"` instead as `0x1000000000000000000000000000`.

This did not fail `U256` parsing because its `FromStr` implementation accepts both `0x`-prefixed and non-`0x`-prefixed strings, and will deserialize both as hex.

## Solution

Check for `0x` prefixes in `from_int_or_hex` and `from_u64_or_hex` if the value matches a `String` instead of `serde_json::Number`. If there is a prefix, deserialize as hex. Otherwise, the value deserializes as a decimal string.

Tests are added for the failing `Genesis` alloc, and unit tests for the `from_int_or_hex` method itself are added.

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/utils/mod.rs
+++ b/ethers-core/src/utils/mod.rs
@@ -515,14 +515,20 @@ where
 {
     #[derive(Deserialize)]
     #[serde(untagged)]
-    enum IntOrHex {
+    enum IntOrString {
         Int(serde_json::Number),
-        Hex(String),
+        JsonString(String),
     }
 
-    match IntOrHex::deserialize(deserializer)? {
-        IntOrHex::Hex(s) => U256::from_str(s.as_str()).map_err(serde::de::Error::custom),
-        IntOrHex::Int(n) => U256::from_dec_str(&n.to_string()).map_err(serde::de::Error::custom),
+    match IntOrString::deserialize(deserializer)? {
+        IntOrString::JsonString(s) => {
+            if s.starts_with("0x") {
+                U256::from_str(s.as_str()).map_err(serde::de::Error::custom)
+            } else {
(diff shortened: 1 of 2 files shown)


ethers-core/src/utils/mod.rs
```diff
@@ -515,14 +515,20 @@ where
...
-    match IntOrHex::deserialize(deserializer)? {
-        IntOrHex::Hex(s) => U256::from_str(s.as_str()).map_err(serde::de::Error::custom),
-        IntOrHex::Int(n) => U256::from_dec_str(&n.to_string()).map_err(serde::de::Error::custom),
+    match IntOrString::deserialize(deserializer)? {
+        IntOrString::JsonString(s) => {
+            if s.starts_with("0x") {
+                U256::from_str(s.as_str()).map_err(serde::de::Error::custom)
+            } else {
+                U256::from_dec_str(&s).map_err(serde::de::Error::custom)
+            }
```
gakonst, line 529: cc @prestwich for ethers-next, you wouldn't believe how much time pinpointing this took us

gakonst (approved):
@mattsse

PR #2360, opened 2023-04-18
 fix(core): add from_x_or_hex to remaining genesis fields 

Rjected:
## Motivation

Some fields were missed for `Genesis` decimal string deserialization.

## Solution

Followup to #2359, adding remaining deserializers to `Genesis` fields.

This changes usages of `from_int_or_hex` to `deserialize_stringified_numeric`, and removes the `from_int_or_hex` methods.

Modifies `StringifiedNumeric` to support arbitrary precision.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (approved):
that the Uint types use decimal string if 0x is missing is a bot horrible

PR #2365, opened 2023-04-19
feat: add metis support

sakulstra:
## Motivation

I wanted to use erhers-rs for Metis network

## Solution

Add ing Metis in this pr

Following the example here: https://github.com/gakonst/ethers-rs/pull/2316/files


prestwich:
CI failures unrelated. Please rebase and resolve conflicts, then we can merge

sakulstra:
@prestwich rebased.

When looking into metis/blockscout also noticed that some foundry commands don't work with blockscout, although blockscout claims to be etherscan compatible: https://github.com/foundry-rs/foundry/issues/4776 - not quite sure where to fix this. As i understand the contract.rs is already more or less agnostic, but probably some adjustment is needed here: https://github.com/gakonst/ethers-rs/blob/master/ethers-etherscan/src/lib.rs#L78 as on blockscout there are no api keys?

PR #2368, opened 2023-04-21
feat(solc): color when formatting Error and OutputDiagnostics

DaniPopes:
## Motivation

Monotone color ugly and bad. More colors good.

Tried as much as possible to use the same format as [Solc's formatter](https://github.com/ethereum/solidity/blob/a297a687261a1c634551b1dac0e36d4573c19afe/liblangutil/SourceReferenceFormatter.cpp#L105).

Before:

image

After (left: `solc --error-codes`, right: this):

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (approved):
this is very nice to have!

all of the parser error branches use regular print, so we don't lose any information in case parsing the output failed.

lgtm

PR #2369, opened 2023-04-21
feat: add raw hash signing methods to ethers signers

Oighty:
## Motivation

Ethers currently exposes wallet methods that conform to standards for signing messages, transactions, and typed data. These are all for specific purposes which add formatted data to a payload or expect it to be in a specific format. However, there are some use cases where it would be useful to just sign a raw hash if it has been pre-computed elsewhere, e.g. by a smart contract or within a script. The generic private key wallet has an existing `sign_hash` function, but the other wallet types do not.

These methods can also be exposed downstream in foundry/cast to sign arbitrary data from the CLI or in foundry/forge to enable more options for signing data within scripts.

A specific use case would be enabling submission of a pre-computed multisig transaction hash using a signer from the CLI, which can allow better tooling for dealing with creation and submission of MS transactions for protocols in production.

## Solution

`sign_raw_hash` method added to the Signer trait and implemented in each of the Signer types.

## PR Checklist

-   [X] Added Tests
-   [ ] Added Documentation
-   [X] Breaking changes (none)


prestwich:
This behavior is forbidden by Ledger, and I believe Trezor. The `INS::SIGN` that you're using in the ledger mod is "sign transaction", and cannot be used to sign arbitrary data. See docs [here](https://github.com/LedgerHQ/app-ethereum/blob/develop/doc/ethapp.adoc#sign-eth-transaction)

In addition, it is generally considered an anti-feature to have a signer with no knowledge of what it is signing. the `sign_raw` in the local wallet is an unfortunate API artifact. Closing this as the spec is not desirable, and cannot be implemented as desired regardless

Oighty:
> 

That makes sense from an end user perspective. Preferred route is to just deal with `sign_typed_struct` downstream then? 

I originally came to this while trying to sign a payload via the CLI with cast. Currently, it only allows signing a message. Typed data could be provided, but the interface might be clunky.

prestwich:
yeah, preferred route for EIP-712 is to use [`Signer::sign_typed_data`](https://docs.rs/ethers/latest/ethers/signers/trait.Signer.html#tymethod.sign_typed_data).

it's hard to see how to sign typed data via the commandline and have a good UX. potentially could take a json file as input, and deserializes to the [`TypedData`](https://docs.rs/ethers/latest/ethers/types/transaction/eip712/struct.TypedData.html) (arguments to the MetaMask/geth `eth_signTypedData` request), the signs

PR #2372, opened 2023-04-22
feat: add fn parse_checksummed

izayl:
## Motivation

as ethers-rs provide `to_checksum` function, but there not provide a way to verify a checksum-like address.
so I add a utilities function to verify a checksum address is valid.

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/utils/mod.rs
+++ b/ethers-core/src/utils/mod.rs
@@ -421,6 +421,23 @@ pub fn to_checksum(addr: &Address, chain_id: Option<u8>) -> String {
     })
 }
 
+/// checks if the given address is a valid checksum address
+///
+/// Returns `true` if addr is a valid checksum address, `false` otherwise.
+pub fn verify_checksum(addr: &str, chain_id: Option<u8>) -> bool {
+    let addr = addr.strip_prefix("0x").unwrap_or(addr);
+    let address = addr.parse::<Address>();
+
+    // wrong address string
+    if address.is_err() {
+        return false
+    }
+
+    let checksum_addr = to_checksum(&address.unwrap(), chain_id);
+
+    return checksum_addr.strip_prefix("0x").unwrap_or(&checksum_addr) == addr
+}
+
 /// Returns a bytes32 string representation of text. If the length of text exceeds 32 bytes,
 /// an error is returned.
 pub fn format_bytes32_string(text: &str) -> Result<[u8; 32], ConversionError> {
@@ -889,6 +906,51 @@ mod tests {
         for (chain_id, addr, checksummed_addr) in addr_list {
             let addr = addr.parse::<Address>().unwrap();
             assert_eq!(to_checksum(&addr, chain_id), String::from(checksummed_addr));
+        }
+    }
+
+    #[test]
+    fn checksum_verify() {
+        let cases = vec![
+            // mainnet
+            // wrong case
+            (None, "0x27b1fdb04752bbc536007a920d24acb045561c26", true),
+            (None, "0x27B1fdb04752bbc536007a920d24acb045561c26", false),
+            // no checksummed
+            (None, "0x52908400098527886e0f7030069857d2e4169ee7", false),
+            // without 0x
+            (None, "0x42712D45473476b98452f434e72461577D686318", true),
+            (None, "42712D45473476b98452f434e72461577D686318", true),
+            // invalid address string
+            (None, "0x52908400098527886E0F7030069857D2E4169EE7", true),
+            (None, "0x52908400098527886E0F7030069857D2E4169EEX", false),
+            (None, "0x52908400098527886E0F7030069857D2E4169EE70", false),
+            // mistyped address
+            (None, "0x5aAeb6053F3E94C9b9A09f33669435E7Ef1BeAed", true),
+            (None, "0x5aAeb6053F3E94C9b9A09f33669435E7Ef1BeAe1", false),
+            // rsk mainnet
+            // wrong case
+            (Some(30), "0x27b1FdB04752BBc536007A920D24ACB045561c26", true),
+            (Some(30), "0x27b1FdB04752BBc536007A920D24ACB045561C26", false),
+            // without 0x
+            (Some(30), "0x3599689E6292B81B2D85451025146515070129Bb", true),
+            (Some(30), "3599689E6292B81B2D85451025146515070129Bb", true),
+            // invalid address string
+            (Some(30), "0x42712D45473476B98452f434E72461577d686318", true),
+            (Some(30), "0x42712D45473476B98452f434E72461577d686318Z", false),
+            // mistyped address
+            (Some(30), "0x52908400098527886E0F7030069857D2E4169ee7", true),
+            (Some(30), "0x52908400098527886E0F7030069857D2E4169ee9", false),
+        ]; // mainnet
+
+        for (chain_id, addr, expected) in cases {
+            assert_eq!(
+                verify_checksum(addr, chain_id),
+                expected,
+                "chain_id: {:?} addr: {:?}",
+                chain_id,
+                addr
+            );
         }
     }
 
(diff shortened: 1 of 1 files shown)


ethers-core/src/utils/mod.rs
```diff
@@ -421,6 +421,27 @@ pub fn to_checksum(addr: &Address, chain_id: Option<u8>) -> String {
...
+
+    // addr is checksummed address
+    if addr.chars().any(|c| c.is_uppercase()) {
+        let checksum_addr = to_checksum(&address.unwrap(), chain_id);
+
+        return checksum_addr.strip_prefix("0x").unwrap_or(&checksum_addr) == addr
+    }
+
+    // addr is not checksummed address
+    true
```
mattsse, line 442: this should be false, no?

ethers-core/src/utils/mod.rs
```diff
@@ -421,6 +421,27 @@ pub fn to_checksum(addr: &Address, chain_id: Option<u8>) -> String {
...
+    let addr = addr.strip_prefix("0x").unwrap_or(addr);
+    let address = addr.parse::<Address>();
+
+    // wrong address string
+    if address.is_err() {
+        return false
+    }
+
+    // addr is checksummed address
+    if addr.chars().any(|c| c.is_uppercase()) {
```
mattsse, line 435:
```suggestion
    if addr.chars().any(|c| c.is_ascii_uppercase()) {
```

ethers-core/src/utils/mod.rs
```diff
@@ -421,6 +421,27 @@ pub fn to_checksum(addr: &Address, chain_id: Option<u8>) -> String {
     })
 }
 
+/// checks if the given address is a valid checksum address
```
mattsse, line 424: needs info when this returns true or false

mattsse (changes requested):

ethers-core/src/utils/mod.rs
```diff
@@ -421,6 +421,27 @@ pub fn to_checksum(addr: &Address, chain_id: Option<u8>) -> String {
...
+pub fn verify_checksum(addr: &str, chain_id: Option<u8>) -> bool {
```
mattsse, line 425: don't love the name,
also not a fan of the `to_checksum` function, should have been named `to_checksum_address`
but, to keep it consistent, this is fine I guess.

izayl: maybe `check_address_checksum`?
ref: https://github.com/web3/web3.js/blob/aaf26c8806bc9fb60cf6dcb6658104963c6c7fc7/packages/web3-utils/src/Utils.js#L140

ethers-core/src/utils/mod.rs
izayl: maybe the function name cause this confuse.
if the address is checksummed address, it need to do validation.
but if the address is not checksummed, it should not to do validation, so I chose to return `true`

ethers-core/src/utils/mod.rs
```diff
@@ -421,6 +421,23 @@ pub fn to_checksum(addr: &Address, chain_id: Option<u8>) -> String {
...
+    let address = addr.parse::<Address>();
+
+    // wrong address string
+    if address.is_err() {
+        return false
+    }
+
+    let checksum_addr = to_checksum(&address.unwrap(), chain_id);
+
+    return checksum_addr.strip_prefix("0x").unwrap_or(&checksum_addr) == addr
```
mattsse, line 438: use 
```
if let Ok() {
let checksum_addr = to_checksum(&address.unwrap(), chain_id);

    return checksum_addr.strip_prefix("0x").unwrap_or(&checksum_addr) == addr

}
false
```

mattsse (changes requested):
smol style nit

ethers-core/src/utils/mod.rs
izayl: Done

ethers-core/src/utils/mod.rs
```diff
@@ -421,6 +421,20 @@ pub fn to_checksum(addr: &Address, chain_id: Option<u8>) -> String {
     })
 }
 
+/// checks if the given address is a valid checksum address
+///
+/// Returns `true` if addr is a valid checksum address, `false` otherwise.
+pub fn verify_checksum(addr: &str, chain_id: Option<u8>) -> bool {
```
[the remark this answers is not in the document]
izayl: like this?

```rs
/// Returns a `Result<Address>`: `Ok(Address)` if addr is checksummed, `Err()` otherwise
pub fn parse_checksummed(addr: &str, chain_id: Option<u8>) -> Result<Address>
```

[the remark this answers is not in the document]
prestwich: function signature is awesome 👌 

I'd write docs similar to this:
```/// Parses an [EIP-1191](https://eips.ethereum.org/EIPS/eip-1191) checksum address. Returns `Ok(address)` if the checksummed address is valid, `Err()` otherwise. If `chain_id` is `None`, falls back to [EIP-55](https://eips.ethereum.org/EIPS/eip-55) address checksum method```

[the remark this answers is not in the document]
Comment (izayl) [reply ethers-core/src/utils/mod.rs]: @prestwich refactored, pls check it

File: ethers-core/src/utils/mod.rs [hunk 7]
```diff
@@ -421,6 +423,24 @@ pub fn to_checksum(addr: &Address, chain_id: Option<u8>) -> String {
...
+
+    if let Ok(address) = addr.parse::<Address>() {
+        let checksum_addr = to_checksum(&address, chain_id);
+        if checksum_addr.strip_prefix("0x").unwrap_or(&checksum_addr) == addr {
+            return Ok(address)
+        } else {
+            return Err(ConversionError::InvalidAddressChecksum)
+        }
+    }
+    Err(ConversionError::TextTooLong)
```
prestwich, line 441: rather than returning text too long, I think we can do the following

1. add a hex error to the ConversionError enum
```
enum ConversionError { 
    ...,
    /// Hex error, usually from address parsing
    #[error(transparent)]
    FromHexError(#[from]hex::FromHexError)
}
```

2. change this function to use a `?`

```
        let addr: Address = addr.strip_prefix("0x").unwrap_or(addr).parse()?
        if checksum_addr.strip_prefix("0x").unwrap_or(&checksum_addr) == addr {
            return Ok(address)
        } else {
            return Err(ConversionError::InvalidAddressChecksum)
        }


```

prestwich:
changes are good. 1 more small ergonomics tweak 🙏

ethers-core/src/utils/mod.rs
izayl: the `FromHexError` is from `rustc_hex` crate, should I add a it?

izayl: https://github.com/paritytech/parity-common/blob/master/fixed-hash/src/hash.rs#L568-L593

prestwich: leave it to parity to make our lives more complicated 🤣 🙄

I think you should be able to make it work as follows:
1. reference the fully qualified associated type `<Address as std::str::FromStr>::Err` in the enum variant
2. instead of doing a `#[from]` in the error variant, use `map_err` to convert

```
// variant specifies that it takes "whatever error the `parse()` implementation says is the error"
FromHexError(<Address as std::str::FromStr>::Err)

// map err on the first line of the function
let addr: Address = addr.strip_prefix("0x").unwrap_or(addr).parse().map_err(ConversionError::FromHexError)?;
```

prestwich: sorry this is a pain. if it gets too much let me know and i'll do the tidying

izayl: @prestwich I've add this `FromHexError`

prestwich (approved):
Approving, pending CI run completion. Thanks for the contribution!

PR #2373, opened 2023-04-23
feat(ws): add methods taking in a tungstenite config

fulminmaxi:
## Motivation

Currently the websocket implementation does not allow to pass in a custom Tungestenite config. For example, its not possible to increase the default msg size over the Tungestenite default, which can lead to `Message Too Long` errors if fetching a lot of events with a lot of fields in them, one case where I've seen this failure in practice is when fetching around 8k events like this one:
```solidity
    event TradeRequested(
        address indexed tokenIn,
        address tokenOut,
        uint256 amountIn,
        uint256 minAmountOut,
        address recipient,
        uint256 indexed tradeIndex,
        address indexed trader
    );
```

## Solution

This PR adds 2 new methods to the `WsClient` and the `WsManager`:
- `connect_with_config` which is the same as `connect`, but allows taking in a Tungstenite config object
- `connect_with_config_and_reconnects` which is the same as `connect_with_reconnects`, but allows taking in a Tungstenite config object.

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-providers/src/rpc/transports/ws/manager.rs
+++ b/ethers-providers/src/rpc/transports/ws/manager.rs
@@ -1,7 +1,7 @@
 use super::{
     backend::{BackendDriver, WsBackend},
     ActiveSub, ConnectionDetails, InFlight, Instruction, Notification, PubSubItem, Response, SubId,
-    WsClient, WsClientError,
+    WebSocketConfig, WsClient, WsClientError,
 };
 use crate::JsonRpcError;
 use ethers_core::types::U256;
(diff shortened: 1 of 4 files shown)


ethers-providers/src/rpc/transports/ws/mod.rs
```diff
@@ -50,6 +50,34 @@ impl WsClient {
         Ok(this)
     }
 
+    /// Establishes a new websocket connection. This method allows specifying a custom websocket
+    /// configuration.
```
mattsse, line 54: can we link to tungsenite's connect functions here in the docs?

ethers-providers/src/rpc/transports/ws/mod.rs
```diff
@@ -50,6 +50,34 @@ impl WsClient {
...
+        let (man, this) = RequestManager::connect_with_config(conn.into(), config.into()).await?;
+        man.spawn();
+        Ok(this)
+    }
+
+    /// Establishes a new websocket connection with auto-reconnects. This method allows specifying a
+    /// custom websocket configuration.
+    pub async fn connect_with_reconnects_and_config(
+        conn: impl Into<ConnectionDetails>,
+        reconnects: usize,
```
mattsse, line 68: nit:
make last argument so arg 1+2 are identical to `connect_with_config`

mattsse (approved):
this mimics tungsenite's API and is an additional function so no breaking changes.

lgtm, pending nit

ethers-providers/src/rpc/transports/ws/mod.rs
fulminmaxi: good point, done in last commit, I've also changed the name so its `connect_with_config_and_reconnects` which will make it pop in IDE hints once you already decided you want to use a config

ethers-providers/src/rpc/transports/ws/mod.rs
fulminmaxi: done

prestwich (changes requested):
The config is discarded, so will not apply to successive reconnects. Need to modify the struct to store the Option<Config>, and the need to modify the `reconnect` function to use it if available

ethers-providers/src/rpc/transports/ws/manager.rs
```diff
@@ -234,6 +240,70 @@ impl RequestManager {
         ))
     }
 
+    #[cfg(not(target_arch = "wasm32"))]
```
prestwich, line 243: can we DRY these? this and `reconnect()` are largely duplicated code. take the part that needs to be conditionally compiled and move it into a smaller function instead of duplicating the entire function

fulminmaxi: should be better now

PR #2380, opened 2023-04-27
Fix `wast` field is missing

wtdcode:
## Motivation

The `wast` field maybe missing in solc compiler output.

JSON input:

```json
{
    "language": "Solidity",
    "sources": {
        "src/tfuzz.sol": {
            "content": "contract T{}"
        }
    },
    "settings": {
        "optimizer": {
            "enabled": false,
            "runs": 200
        },
        "outputSelection": {
            "*": {
                "": [
                    "*"
                ],
                "*": [
                    "*"
                ]
            }
        },
        "evmVersion": "london",
        "libraries": {}
    }
}
```

Output:

```bash
/tmp $ solc --standard-json < compile.json | jq . | grep -C1 wasm
        },
        "ewasm": {
          "wasm": ""
        },
/tmp $
```

This is causing failure to deserialize compilers json outputs while the compilation is successful.

## Solution

As proposed.

## PR Checklist

-   [ ] Added Tests

No need for tests.

-   [ ] Added Documentation

No need for further docs.

-   [ ] Breaking changes

This shouldn't introduce any breaking changes.

### DIFF
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -1779,6 +1779,7 @@ pub struct Creation {
 
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
 pub struct Ewasm {
+    #[serde(default)]
     pub wast: String,
     pub wasm: String,
 }
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -1779,8 +1779,8 @@ pub struct Creation {
 
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
 pub struct Ewasm {
-    #[serde(default)]
-    pub wast: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub wast: Option<String>,
     pub wasm: String,
 }
 


ethers-solc/src/artifacts/mod.rs
```diff
@@ -1779,6 +1779,7 @@ pub struct Creation {
 
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
 pub struct Ewasm {
+    #[serde(default)]
     pub wast: String,
```
mattsse, line 1783: this is no longer symmetrical.

```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub wast: Option<String>,
```

mattsse (changes requested):

ethers-solc/src/artifacts/mod.rs
wtdcode: Changing it to `Option<String>` will introduce breaking changes. Is that okay?

wtdcode: done

mattsse: is fine imo, almost certain that nobody's using this

mattsse (approved):

PR #1982, opened 2022-12-29
fix: onbjerg's typed ast integration

iFrostizz:
## Motivation

Following up on https://github.com/gakonst/ethers-rs/pull/1943 
This caused lots of issues, shown in this PR: #1966 
And modifying in order to integrate it in foundry without breaking compatibility: https://github.com/foundry-rs/foundry/pull/3989

## Solution

Recover some typings of the ast into the old ones, and providing an helper function to convert from old to typed ast.
This is still very much a WIP but open for thoughts, mainly for the old => typed ast function `to_typed()`.

PR #2381, opened 2023-04-27
Introduce the visitor implementation

wtdcode:
## Motivation

This is just some tedious merging work with #1982 which introduces the AST visitor features. I simply confirm the test works and format the code. I'm glad to work further on this PR.

cc @iFrostizz 

## Solution

As proposed.

## PR Checklist

-   [ x ] Added Tests
-   [ x ] Added Documentation
-   [ ] Breaking changes

I notice that there are actually a few break changes to AST structures (see `mod.rs`). However, considering that the parsed AST is not useful without a visitor implementation, I think previous users won't actually use too much of them. Even if some users used the structures, it's trivial to switch to new implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,6 +88,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
+name = "as-any"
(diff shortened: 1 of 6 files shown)


wtdcode:
Hey any review on this?

iFrostizz:
I think that before considering merging those changes, we should be 100% sure that it won't break Foundry. This is what happened in the days and is probably why it seems scary now to merge this.

PR #2390, opened 2023-04-30
feat: added Optimism deposited transaction support

merklefruit:
## Motivation

Optimism has an extra transaction type: the Deposited transaction (type 0x7E), with 3 new transaction fields as described [in the spec here](https://github.com/ethereum-optimism/optimism/blob/develop/specs/deposits.md#the-deposited-transaction-type). The RLP encoding order is also different.

## Solution

This PR introduces a new feature "optimism" much like the existing "celo" one, which enables the 3 new optional transaction fields.

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation


prestwich:
I'm supportive of adding first-class optimism objects. However, I am concerned that this change might be too large for our current release cycle. We're trying to put this version into maintenance so that we can focus on ethers@3. 

In addition, our opinion generally is that the Celo feature was a bad approach to network-specific behavior in the first place, and this ought to be solved by abstraction of the argument and return types in the Middleware trait

@gakonst do you think it is worth it to unblock optimism folks?

gakonst (approved):
I would like to add this because it also unblocks op-reth work for client side. I am open to merging this if CI passes

ethers-core/src/types/transaction/response.rs
```diff
@@ -65,6 +65,22 @@ pub struct Transaction {
     /// ECDSA signature s
     pub s: U256,
 
+    ///////////////// Optimism-specific transaction fields //////////////
+    /// The source-hash that uniquely identifies the origin of the deposit
+    #[cfg(feature = "optimism")]
+    #[serde(skip_serializing_if = "Option::is_none", rename = "sourceHash")]
```
prestwich, line 71: needs `#[serde(default)]`

ethers-core/src/types/transaction/response.rs
```diff
@@ -65,6 +65,22 @@ pub struct Transaction {
...
+    pub source_hash: Option<H256>,
+
+    /// The ETH value to mint on L2
+    #[cfg(feature = "optimism")]
+    #[serde(skip_serializing_if = "Option::is_none")]
```
prestwich, line 76: needs `#[serde(default)]`

ethers-core/src/types/transaction/response.rs
```diff
@@ -65,6 +65,22 @@ pub struct Transaction {
...
+    pub mint: Option<U256>,
+
+    /// True if the transaction does not interact with the L2 block gas pool
+    #[cfg(feature = "optimism")]
+    #[serde(skip_serializing_if = "Option::is_none", rename = "isSystemTx")]
```
prestwich, line 81: needs `#[serde(default)]`

ethers-core/src/types/transaction/response.rs
```diff
@@ -173,6 +191,20 @@ impl Transaction {
                 if let Some(chain_id) = self.chain_id {
                     rlp.append(&normalize_v(self.v.as_u64(), U64::from(chain_id.as_u64())));
                 }
+                rlp.append(&self.r);
+                rlp.append(&self.s);
+            }
+            // Optimism Deposited Transaction
+            #[cfg(feature = "optimism")]
+            Some(x) if x == U64::from(0x7E) => {
+                rlp.append(&self.source_hash.expect("sourceHash is required for deposited tx"));
```
prestwich, line 200: should `source_hash` be `<_>` instead of `Option<_>`

prestwich (changes requested):
Need to update CI to run with the optimsim feature enabled. 
Need to document that Optimism and Celo are mutually exclusive

ethers-core/src/types/transaction/response.rs
prestwich: oh i thought we were in a TypedTransaction context. I suppose this is fine here, but you should use `rlp_opt`, not a panic

you'll also need to add a TypedTransaction enum variant

README.md
```diff
@@ -58,6 +58,22 @@ You can get one [here](https://polygonscan.io/apis).
 There is abigen support for Avalanche and the Fuji test network. It is recommended that you set the `SNOWTRACE_API_KEY` environment variable.
 You can get one [here](https://snowtrace.io/apis).
 
+### Optimism support
+
+Optimism is supported via the `optimism` feature flag:
+
+```toml
+[dependencies]
+ethers = { version = "1.0.2", features = ["optimism"] }
```
DaniPopes, line 67:
```suggestion
ethers = { version = "2.0", features = ["optimism"] }
```

DaniPopes (changes requested):

ethers-core/src/types/transaction/response.rs
merklefruit: > you'll also need to add a TypedTransaction enum variant

Ok, @prestwich regarding this do you think we should reuse the `Eip1559TransactionRequest` type and extend it with the optimism feature, or instead create a new `OptimismDepositedTransactionRequest` type?

prestwich: if optimism supports both separately, do both

merklefruit:
Ok, I've split the addition of a new TypedTransaction variant to a secondary PR to make the diffs cleaner. 
Here: https://github.com/merklefruit/ethers-rs/pull/1

I've also added a couple tests for RLP encoding/decoding of this new tx type. Let me know what other kinds of tests I should eventually add!

gakonst (approved):

merklefruit:
@gakonst I believe the PR above (https://github.com/merklefruit/ethers-rs/pull/1) was not merged into this branch yet. I've closed it now, so I need to reopen a PR from this branch

PR #2392, opened 2023-05-01
feat: add revert data field to create result struct

michprev:
## Motivation

Until now, there was no way to set revert data for CREATE traces. For example, for sub-trace creating this contract

```solidity
contract Foo {
    constructor() {
        revert("aaa");
    }
}
```

it was not possible to store the revert data (`0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000036161610000000000000000000000000000000000000000000000000000000000` in this case) in the trace object. For non-CREATE traces, there is the `output` field that can be used both for storing the return value and the revert data, depending on the context.

## Solution

Added a new field named `revert_data` to the `CreateResult` struct. Both `revert_data` and `code` are now optional and should be mutually exclusive.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/types/trace/filter.rs
+++ b/ethers-core/src/types/trace/filter.rs
@@ -169,7 +169,11 @@ pub struct CreateResult {
     #[serde(rename = "gasUsed")]
     pub gas_used: U256,
     /// Code
-    pub code: Bytes,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub code: Option<Bytes>,
+    /// Revert data
+    #[serde(rename = "revertData", skip_serializing_if = "Option::is_none")]
+    pub revert_data: Option<Bytes>,
     /// Assigned address
     pub address: Address,
 }


mattsse (commented):
is the part of the original trace object

https://openethereum.github.io/JSONRPC-trace-module ?

michprev:
Looking at these lines https://github.com/openethereum/openethereum/blob/6c2d392d867b058ff867c4373e40850ca3f96969/crates/ethcore/src/trace/types/trace.rs#L298-L311 and https://github.com/openethereum/openethereum/blob/6c2d392d867b058ff867c4373e40850ca3f96969/crates/ethcore/src/trace/types/trace.rs#L36-L45 there is no such field. However, the current implementation in ethers-rs seems to differ anyway (compare https://github.com/openethereum/openethereum/blob/6c2d392d867b058ff867c4373e40850ca3f96969/crates/ethcore/src/trace/types/trace.rs#L298-L311 and https://github.com/gakonst/ethers-rs/blob/f7454f27666750449ca9005a4db032025b4c36ff/ethers-core/src/types/trace/filter.rs#L114-L125).

It seems like in the original OpenEthereum implementation it was not possible to retrieve the revert reason data (which is useful information) for both Create and Call traces, while in ethers-rs it is currently possible to retrieve the revert reason data only for Call traces.

To make this PR more closer to the original implementation, the added field could be named `output` instead of `revert_data`. However, if set, it would only contain revert data.

PR #2396, opened 2023-05-02
add mock response to test json rpc errors

mdtanrikulu:
## Motivation

The goal is to be able to test any json-rpc errors and consequences afterwards. I needed such modification while writing [test cases](https://github.com/ensdomains/ethers-ccip-read/blob/5c4610dcec7d0fdbcb1886a7cb1a43038159a0fc/src/middleware.rs#L500) for a middleware. 

## Solution

- Added new enum called `MockResponse`, instead of passing the `Value`, now we pass this enum as a response type. 
- MockProvider now checks the response types with `match` and treats accordingly.
- new fn called `push_response` helps pushing the JsonRpcError as the next response.

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-providers/src/rpc/transports/mock.rs
+++ b/ethers-providers/src/rpc/transports/mock.rs
@@ -160,7 +160,7 @@ impl From<MockError> for ProviderError {
 #[cfg(not(target_arch = "wasm32"))]
 mod tests {
     use super::*;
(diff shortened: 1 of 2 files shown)


DaniPopes (approved):
Thanks!

PR #2407, opened 2023-05-09
Added linea testnet

syahrul12345:
Add the linea testnet

### DIFF
--- a/ethers-core/src/types/chain.rs
+++ b/ethers-core/src/types/chain.rs
@@ -153,6 +153,8 @@ pub enum Chain {
 
     BaseGoerli = 84531,
 
+    LineaTestnet = 59140,
+
     #[strum(to_string = "zksync")]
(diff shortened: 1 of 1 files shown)


ethers-core/src/types/chain.rs
```diff
@@ -334,6 +337,7 @@ impl Chain {
             Boba |
             ZkSync |
             ZkSyncTestnet |
+            LineaTestnet |
```
mattsse, line 340: I don't think this is right. supports eip1559:

https://explorer.goerli.linea.build/tx/0x0bfb5ade28b967cd4ab7721d624cd8c580558a5bbd2f63ceee3dd1c6312680a9

mattsse (changes requested):

mattsse (approved):

prestwich (approved):
CI failure unrelated

Issue #2408, opened 2023-05-09
Remove eip712 feature


dbelv:
**Version**
Latest

**Description**
eip712 feature is not propagated from the ethers level toml to the submodules. Given it is marked as deprecated the feature flag should be removed and the functionality of the eip712 should been enabled by default. Discussed with @DaniPopes in telegram.

PR #2409, opened 2023-05-09
Remove the eip712 feature flag and have it enabled by default

dbelv:
## Motivation
Closes https://github.com/gakonst/ethers-rs/issues/2408

## Solution
Remove feature flag enabling by default.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-contract/Cargo.toml
+++ b/ethers-contract/Cargo.toml
@@ -58,3 +58,6 @@ legacy = []
 
 rustls = ["ethers-contract-abigen/rustls"]
 openssl = ["ethers-contract-abigen/openssl"]
+
+# Deprecated
+eip712 = []
--- a/ethers-contract/Cargo.toml
+++ b/ethers-contract/Cargo.toml
@@ -58,6 +58,3 @@ legacy = []
 
 rustls = ["ethers-contract-abigen/rustls"]
 openssl = ["ethers-contract-abigen/openssl"]
-
-# Deprecated
-eip712 = []
--- a/ethers-core/Cargo.toml
+++ b/ethers-core/Cargo.toml
@@ -70,6 +70,3 @@ rand.workspace = true
 celo = ["legacy"] # celo support extends the transaction format with extra fields
 legacy = []
 macros = ["syn", "cargo_metadata", "once_cell"]
-
-# Deprecated
-eip712 = []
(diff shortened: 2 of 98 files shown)


dbelv:
Muscle memory had me doing `cargo fmt` which is unfortunately why large amount of touched files within the same commit vs the actual implementation of the issue.

dbelv:
Sorry I have overloaded the PR for clippy fixes as well

ethers-contract/Cargo.toml
```diff
@@ -58,6 +58,3 @@ legacy = []
 
 rustls = ["ethers-contract-abigen/rustls"]
 openssl = ["ethers-contract-abigen/openssl"]
-
-# Deprecated
-eip712 = []
```
DaniPopes, line 63: Could you add them back here too?

ethers-core/Cargo.toml
```diff
@@ -70,6 +70,3 @@ rand.workspace = true
 celo = ["legacy"] # celo support extends the transaction format with extra fields
 legacy = []
 macros = ["syn", "cargo_metadata", "once_cell"]
-
-# Deprecated
-eip712 = []
```
DaniPopes, line 75: Same

DaniPopes (changes requested):

ethers-core/README.md
```diff
@@ -10,10 +10,6 @@ other utilities for interacting with the Ethereum ecosystem
 
 For more information, please refer to the [book](https://gakonst.com/ethers-rs).
 
-## Feature flags
-
--   `eip712`: Provides the `Eip712` trait and derive procedural macro for EIP-712 encoding of typed data.
```
prestwich, line 15: modify to say "does nothing"

ethers-core/src/types/signature.rs
```diff
@@ -69,7 +69,6 @@ impl fmt::Display for Signature {
     }
 }
 
-#[cfg(feature = "eip712")]
```
prestwich, line 72: combine with main impl block

prestwich (changes requested):
keep the flag in the Cargo.toml. Removing it will break compilation for library consumers

ethers-contract/Cargo.toml
dbelv: Ah right, yep.

DaniPopes (approved):
Thanks!

PR #2411, opened 2023-05-10
chore: make subscription examples runnable

mattsse:
add [example] sections to subscriptions module they can be executed


DaniPopes:
You can already run with `cargo r -p examples-subscriptions --example subscribe_blocks`, what does this do?

mattsse:
I see, my bad then

PR #2416, opened 2023-05-13
chore: make `abigen` offline by default, fix `ethers-solc` features

DaniPopes:
## Motivation

- `Multicall` and other stuff gated by `#[cfg(feature = "abigen")]` in `ethers-contract` are not accessible with `abigen-offline`
- `ethers-contract-abigen` is online by default in `ethers-contract` and `ethers`. If there are both `ethers-contract-abigen` as a build script, and `ethers{-contract}` with the `abigen` feature, then it will be built twice: once without `online` for the build script, and once with `online`.
- forward `ethers-solc` features from `ethers-etherscan`

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-contract/Cargo.toml
+++ b/ethers-contract/Cargo.toml
@@ -50,8 +50,8 @@ tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 [features]
 default = ["abigen"]
 
-abigen-offline = ["ethers-contract-abigen", "ethers-contract-derive"]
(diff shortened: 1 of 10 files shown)


mattsse (approved):
I think this is reasonable.
this should discourage http abigen links which aren't very stable anyway due flaky etherscan API

~~ci failing, likely feature snafu~~
it's new solc version check: https://github.com/gakonst/ethers-rs/pull/2417

PR #2419, opened 2023-05-15
fix: fix re-subscription on websocket reconnect

gbrew:
Add an in-flight request for each re-subscription request to ensure that the response from the server with the new internal subscription id is handled.

## Motivation

This change fixes issue 2418, which is an issue where websocket subscriptions can stop working after a websocket reconnect.

## Solution

The underlying issue is that the re-subscription code did not have an in-flight request for the resubscription requests. They were issued properly, but the response from the server containing the new server-side subscription ID was then dropped as a result. This change adds the appropriate in-flight requests so the new server-side ID is handled properly and updated.

## PR Checklist

-   [ x] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


prestwich (approved):

gbrew:
Thanks for the review @prestwich , and thanks for all the work on reconnects!

I had a quick look at the failing CI tests, and they all seem likely related to a solc version change ( ethers-solc compile::tests::test_detect_version ), and unrelated to my change. Should I worry about making those pass or just ignore them? And is there anything else I should be doing to get this merged?

prestwich:
> Should I worry about making those pass or just ignore them? And is there anything else I should be doing to get this merged?

just ignore those 👍  I'll be doing a pass later today to merge things and hopefully get those addressed as well

PR #2420, opened 2023-05-16
feat(solc): `0.8.20` support & set Shanghai as default EVM Version

Evalir:
## Motivation

Currently SVM is blocking ethers from being bumped so we can soon cut a Foundry 1.0 release. While we deal with transferring repos, I went ahead and forked the necessary repos and updated everything.

## Solution

Sets svm-rs to use the forked repo in the meanwhile, and implements all the necessary changes to use shanghai and 0.8.20.

cc @DaniPopes looping you in for if i missed anything—edits by maintainers should be enabled


DaniPopes:
Crates with git dependencies aren't allowed to be released on crates.io, and duplicate of #2414.

PR #2426, opened 2023-05-22
fix(`verify`): make verification result an option to handle blockscout's format

Evalir:
## Motivation

I've been investigating verification issues due to https://github.com/foundry-rs/foundry/issues/4909 — and it turns out that blockscout, unlike etherscan, likes to return `null` on the `result` field sometimes, instead of an empty string on `message` and the actual message on the `result` field, like etherscan. This in turn introduced a bug when detecting if a contract was already verified as it expects that the blockscout format is identical to etherscan's. 

To be a bit more specifc, this format is returned by blockscout when a contract is not verified, and this is _not_ what etherscan returns (see the `null` in result):

```
{
  "message": "Contract source code not verified",
  "result": null,
  "status": "0"
}
```

While making a blockscout verification client might be the "best" fix, this is a slight difference in formats and I opted for this solution instead. We can handle separating blockscout & etherscan verification later.

## Solution

Parses the `result` field from the API response as an Option instead of just a string, to handle cases where result is `null`.

### DIFF
--- a/ethers-etherscan/src/contract.rs
+++ b/ethers-etherscan/src/contract.rs
@@ -10,7 +10,6 @@ use ethers_core::{
 use semver::Version;
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, path::Path};
-use tracing::trace;
 
 #[cfg(feature = "ethers-solc")]
 use ethers_solc::{artifacts::Settings, EvmVersion, Project, ProjectBuilder, SolcConfig};
--- a/ethers-etherscan/src/contract.rs
+++ b/ethers-etherscan/src/contract.rs
@@ -9,8 +9,8 @@ use ethers_core::{
 };
 use semver::Version;
 use serde::{Deserialize, Serialize};
-use tracing::trace;
 use std::{collections::HashMap, path::Path};
+use tracing::trace;
 
 #[cfg(feature = "ethers-solc")]
 use ethers_solc::{artifacts::Settings, EvmVersion, Project, ProjectBuilder, SolcConfig};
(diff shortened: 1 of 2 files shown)


ethers-etherscan/src/contract.rs
```diff
@@ -331,17 +331,23 @@ impl Client {
         }
 
         let query = self.create_query("contract", "getabi", HashMap::from([("address", address)]));
-        let resp: Response<String> = self.get_json(&query).await?;
-        if resp.result.starts_with("Max rate limit reached") {
+        let resp: Response<Option<String>> = self.get_json(&query).await?;
+
+        let result = resp.result.unwrap_or("".to_string());
```
mattsse, line 336: null will now always be a serde error,

I'd rather have a new error variant for EmptyResult or something

mattsse (changes requested):
nice find!

ethers-etherscan/src/contract.rs
Evalir: done in https://github.com/gakonst/ethers-rs/pull/2426/commits/94ad922d9651ef3c1ff9cb5fc2acb0f5a43f5389 — we now bubble up the error with an `EmptyResult` variant

mattsse (approved):

ethers-etherscan/src/errors.rs
```diff
@@ -23,6 +23,8 @@ pub enum EtherscanError {
     Serde(#[from] serde_json::Error),
     #[error("Contract source code not verified: {0}")]
     ContractCodeNotVerified(Address),
+    #[error("Response result is unexpectedly empty")]
+    EmptyResult { status: String, message: String },
```
DaniPopes, line 27: Should the error message include `status` or `message`?

Evalir: Oh yeah, let's include both! the status should only be a number so it's short. The message is what actually has the error

Evalir: done on https://github.com/gakonst/ethers-rs/pull/2426/commits/fa5c243a40adbd4af7aa291cb90ea5f52594ae9b

DaniPopes (approved):

PR #2429, opened 2023-05-24
feat: add nonce helper function

dennohpeter:
Add a helper function to override nonce in contact `FunctionCall`

## Motivation

Currently, there is no way to override nonce from the contract `FunctionCall` struct. It would be useful to make it easier to override nonce too as it is with value, gas and gas price 

## Solution

I added a helper function to set the nonce, so it can be overridden before a transaction is submitted.

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-contract/src/call.rs
+++ b/ethers-contract/src/call.rs
@@ -256,6 +256,12 @@ where
         self.block = Some(block.into());
         self
     }
+
+    /// Sets the `nonce` field in the transaction to the provided value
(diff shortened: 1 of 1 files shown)


dennohpeter:
@DaniPopes

dennohpeter:
@mattsse

mattsse (approved):
lgtm, 

I believe this builder function was just missing

@DaniPopes

DaniPopes (approved):
Thanks!

Issue #2430, opened 2023-05-24
Add `Chain::supports_push_zero` function


mattsse:
similar to 

https://github.com/gakonst/ethers-rs/blob/748fd27096e713a3965c101ee4de4a46689bd66b/ethers-core/src/types/chain.rs#L313

add `supports_push_zero` helper function that should return true if the chain currently supports push0 opcode (with shanghai update)

DaniPopes:
Fixed in #2431

PR #2431, opened 2023-05-24
feat(core): Utility to detect Shanghai-enabled chains

Evalir:
## Motivation

We want to better handle EVM discrepancies between different chains (see https://github.com/foundry-rs/foundry/issues/748). One of these things would be a little utility on the `Chain` type to quickly detect if a chain is Shanghai-enabled or not.

## Solution

Adds the utility

## PR Checklist

-   [n/a] Added Tests
-   [n/a] Added Documentation
-   [n/a] Breaking changes

### DIFF
--- a/ethers-core/src/types/chain.rs
+++ b/ethers-core/src/types/chain.rs
@@ -360,6 +360,13 @@ impl Chain {
         }
     }
 
+    pub const fn is_shanghai(&self) -> bool {
+        match self {
+            Chain::Mainnet | Chain::Goerli | Chain::Sepolia => true,
+            _ => false,
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


ethers-core/src/types/chain.rs
```diff
@@ -360,6 +360,13 @@ impl Chain {
         }
     }
 
+    pub const fn is_shanghai(&self) -> bool {
```
mattsse, line 363: needs docs, and I think we settled on `supports_push0`

#2430

mattsse (changes requested):

DaniPopes (approved):

PR #2432, opened 2023-05-25
feat: allow for Authorization header override with raw parameter

kristian1108:
This PR creates a Raw Authorization type which allows the user to set the full Authorization header in a websocket request if the server they are connecting to does not conform to Bearer or Basic auth.

## PR Checklist

-   [trivial change ] Added Tests
-   [x ] Added Documentation
-   [ no] Breaking changes

### DIFF
--- a/ethers-providers/src/rpc/transports/common.rs
+++ b/ethers-providers/src/rpc/transports/common.rs
@@ -236,6 +236,8 @@ pub enum Authorization {
     Basic(String),
     /// Bearer Auth
     Bearer(String),
+    /// If you need to override the Authorization header value
+    Raw(String),
 }
 
 impl Authorization {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this is reasonable imo

DaniPopes (approved):
Thanks! Don't worry about the CI

weezy20:
I have an issue where I cannot get this enum variant to work as intended. My WSS endpoint expects a custom header and an api key, and if I provide it using `Authroization::raw`, I just get a `HTTP 403`. The reason this is strange is because on an equivalent js app using `web3-providers-ws`, we are able to instantiate the ws connection correctly using our custom headers : 

```js
const customWsProvider = new WebSocketProvider(WS_URL, {
  headers: {
    'KEY' : val
  },
})
```
in my experience the equivalent `ethers-rs` code is not totally equivalent: 
```rs
let api_key = // .. ;
let auth = Authorization::raw(format!("KEY {api_key}"));
let provider = Provider::<Ws>::connect_with_auth(url, auth).await?
```
Any help would be appreciated.

Issue #2356, opened 2023-04-16
`formatEther(x)` != `formatUnits(x,18)` 


halo3mic:
In ethers.js `formatEther(x) == formatUnits(x, 18)`, I noticed this is not the case with ethers.rs.  `format_ether` truncates the input number and returns U256, while`format_units` returns the input number in decimal representation as a string.

Is this intentional, or should it be like below?
```rust
pub fn format_ether<T: Into<ParseUnits>>(amount: T) -> Result<String, ConversionError> {
    format_units(amount, "ether")
}
```

culda:
+1
I can't work out how to print a decimal representation of ETH, like 0.1

culda:
I mean `format_units(x, 18)` works, but would expect it to be equivalent to `format_ether(x)`

AndoroidX:
Fixed with #2435 
@halo3mic you can close the issue now.

PR #2435, opened 2023-05-27
ethers::utils::format_ether should show decimal places

AndoroidX:
## Motivation
Well, I've just spent 30 minutes trying to understand why I see 0 ETH as the output of my code. Turns out, it was a problem in prase_ether function. Noticed an issue #2356 

## Solution
Use `parse_units(value, "ether").unwrap();`
We can safely unwrap here, since `parse_units` returns Err only when units >= 77.

## PR Checklist

-   [ x ] Added Tests
-   [ x ] Added Documentation
-   [ x ] Breaking changes

### DIFF
--- a/ethers-core/src/utils/mod.rs
+++ b/ethers-core/src/utils/mod.rs
@@ -141,8 +141,15 @@ construct_format_units_from! {
 /// in ether (instead of wei)
 ///
 /// Divides the input by 1e18
-pub fn format_ether<T: Into<U256>>(amount: T) -> U256 {
-    amount.into() / WEI_IN_ETHER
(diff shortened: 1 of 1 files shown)


DaniPopes:
This is great, thanks! It's a small enough breaking change that I'm OK with releasing this in a patch or minor version, wdyt @gakonst @prestwich?

prestwich (approved):

prestwich:
> 
This is great, thanks! It's a small enough breaking change that I'm OK with releasing this in a patch or minor version, wdyt @gakonst @prestwich?

agreed

PR #2406, opened 2023-05-08
feat: move `spoof` module to `ethers_core`, add stateoverrides to `GethDebugTracingCallOptions`

georgewhewell:
## Motivation
i want to use `stateoverrides` in `debug_traceCall`

## Solution
as described in #2298
- move `ethers_providers::types::geth::spoof` module to `ethers_core`
- add `state_overrides` to `GethDebugTracingCallOptions`

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

PR #2436, opened 2023-05-27
ci: add more feature tests and checks

DaniPopes:
## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


ethers-core/src/types/trace/geth.rs
DaniPopes, line 1: Not caught in CI previously (#2406)

ethers-providers/src/ext/admin.rs
DaniPopes, line 1: Some fields missing when deserializing the `EthInfo` struct, so a test was failing.

ethers-solc/Cargo.toml
DaniPopes, line 1: Cannot enable dev-dep features

.github/workflows/ci.yml
```diff
@@ -17,40 +17,32 @@ env:
...
                 flags:
                     - name: no default features
-                      flags: --workspace --no-default-features
+                      flags: --no-default-features
                     - name: default features
-                      flags: --workspace
+                      flags: ""
                     - name: all features
-                      flags: --workspace --all-features
+                      flags: --all-features
```
DaniPopes, line 32: `--workspace` is not needed, because the only crates that are not included are the example ones, which are tested in another job. This allows removing the windows hacks.

PR #2443, opened 2023-05-27
chore: remove Cargo.lock

DaniPopes:
## Motivation

See <https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries>

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (approved):
supportive,

this would also prevent issues where ethers CI passes because of locked deps but fails downstream.

PR #2449, opened 2023-05-31
feat(abigen): emit cargo build script directives

DaniPopes:
## Motivation

... and an `include_bytes!` when not in build scripts. Signals Cargo to rebuild if the source files change.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/lib.rs
+++ b/ethers-contract/ethers-contract-abigen/src/lib.rs
@@ -36,7 +36,7 @@ mod util;
 pub use ethers_core::types::Address;
 
 use contract::{Context, ExpandedContract};
-use eyre::{Context as _, Result};
+use eyre::Result;
 use proc_macro2::{Ident, TokenStream};
 use quote::ToTokens;
 use std::{collections::HashMap, fmt, fs, io, path::Path};
@@ -74,6 +74,11 @@ pub struct Abigen {
     /// Whether to format the generated bindings using [`prettyplease`].
     format: bool,
 
+    /// Whether to emit [cargo build script directives][ref].
+    ///
+    /// [ref]: https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script
+    emit_cargo_directives: bool,
+
     /// Manually specified contract method aliases.
     method_aliases: HashMap<String, String>,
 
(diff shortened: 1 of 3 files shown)


ethers-contract/ethers-contract-abigen/src/lib.rs
```diff
@@ -95,44 +115,31 @@ impl Abigen {
...
-        let name = file_name
-            .to_str()
-            .ok_or_else(|| eyre::eyre!("File name contains invalid UTF-8"))?
-            .split('.') // ignore everything after the first `.`
-            .next()
-            .unwrap(); // file_name is not empty as asserted by .file_name() already
-        let contents = fs::read_to_string(&path).wrap_err("Could not read file")?;
-
-        Self::new(name, contents)
+        let path = path.as_ref().to_str().ok_or_else(|| eyre::eyre!("invalid path"))?;
```
mattsse, line 139: nit: error should include the path

ethers-contract/ethers-contract-abigen/src/lib.rs
```diff
@@ -202,12 +209,33 @@ impl Abigen {
         self
     }
 
+    /// Specify whether to print [cargo build script directives][ref] if the source is a path. By
+    /// default, this is true only when executing inside of a build script.
+    ///
+    /// [ref]: https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script
```
mattsse, line 215: I need to adapt this href style

ethers-contract/ethers-contract-abigen/src/lib.rs
```diff
@@ -368,6 +396,11 @@ impl ContractBindings {
     }
 }
 
+/// Returns whether the current executable is a cargo build script.
+fn in_build_script() -> bool {
+    std::env::var("CARGO_MAKEFLAGS").is_ok()
```
mattsse, line 401: TIL

mattsse (approved):
nice, smol nit re error message otherwise lgtm

ethers-contract/ethers-contract-abigen/src/lib.rs
DaniPopes: honestly i just went over all of the env vars, and this one was the only one that is set only on build scripts, per docs. Not 100% sure

DaniPopes: `TARGET` should be fine as well.

Issue #2454, opened 2023-06-05
Support Enterprise Quorum


rodion-lim-partior:
**Is your feature request related to a problem? Please describe.**
There is a previous feature request on forge to support enterprise Quorum https://github.com/foundry-rs/foundry/issues/2933. This issue is raised on `ethers-rs` since it is pre-requisite before that can take place.

**Describe the solution you'd like**
There should be a conditional feature toggle to allow for quorum related fields. Quorum requires specific fields such as `privateFor` when sending the tx request via json rpc. When sending transactions, we should take in an optional field for privacy options according to the Quorum specifications.

**Describe alternatives you've considered**
-

**Additional context**
-

PR #2455, opened 2023-06-05
feat: support enterprise quorum (#2454)

rodion-lim-partior:
This PR is in relation to https://github.com/gakonst/ethers-rs/issues/2454. The purpose of the PR is to extend `ethers-rs` existing implementation to support private transactions to an enterprise quorum blockchain https://consensys.net/quorum/.

## Motivation

Quorum extends geth to allow for private transaction which are stored within Tessera. To support private transactions, there are additional privacy fields included before the submission of the json rpc payload.

## Solution

For a start, this PR extends the transaction object to have a privateFor field. It also extends abigen and contract factory to take in the privateFor parameter when calling `send`. In future, there are plans to extend privacy parameters to a privacy options object to allow for enhanced privacy mode within Quorum. Also, send raw transaction will be amended to first hit the Tessera API endpoint. 

## PR Checklist

-   [Y] Added Tests
-   [Y] Added Documentation
-   [N] Breaking changes


rodionlim:
@gakonst , any chance we can get this PR to be looked at?

PR #2460, opened 2023-06-06
fix: avoid payload moving

imotai:
## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [x] Breaking changes

### DIFF
--- a/ethers-core/src/types/signature.rs
+++ b/ethers-core/src/types/signature.rs
@@ -119,7 +119,7 @@ impl Signature {
     ///
     /// Recovery signature data uses 'Electrum' notation, this means the `v`
     /// value is expected to be either `27` or `28`.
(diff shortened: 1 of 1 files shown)


imotai:
> 

@prestwich  do you have some ideas about the trait itself should be implemented for &T?

mattsse (commented):
considering that this is only a convenience function for `Self::recover(T::encode_eip712)` and `fn encode_eip712(&self` takes self I think only taking a ref would be the more appropriate signature

mattsse:
> Maybe the trait itself should be implemented for `&T`? cc @prestwich

this would be possible with an extension trait, but perhaps not worth it. not sure if the user would have to import that trait as well.

prestwich:
I have no opinion on how this ought to be resolved. The PR as written is a sufficiently minor breaking API change that I would be comfortable with it as a patch release

DaniPopes (approved):

PR #2468, opened 2023-06-07
fix(`verification`): re-introduce etherscan typo and populate properly-spelled field as backup

Evalir:
## Motivation

So, it turns out that etherscan still has this `constructorArguements` typo and doesn't support the properly spelled version. However, some blockscout instances seem to be be able to parse `constructorArguments` and not `constructorArguements`, so as a workaround for people we're populating both fields and sending them, as the APIs do not discriminate on redundant json fields.

## Solution

Adds another field `blockscout_constructor_args`, that gets autopopulated with the same info that gets passed in from `constructors_args`. The former will get serialized to `constructorArguments`.

## PR Checklist

-   [n/a] Added Tests
-   [x] Added Documentation
-   [n/a] Breaking changes

cc @mattsse

### DIFF
--- a/ethers-etherscan/src/verify.rs
+++ b/ethers-etherscan/src/verify.rs
@@ -25,16 +25,18 @@ pub struct VerifyContract {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub runs: Option<String>,
     /// The constructor arguments for the contract, if any.
-    /// NOTE: This is obviously misspelled. The reason for this is that Etherscan has had
-    /// this misspelling on their API for quite a long time, and changing it would break
-    /// verification with arguments.
-    /// For instances (e.g some blockscout instances) that might support the proper field
-    /// and not the misspelling, the private field `blockscout_constructor_arguments` is
-    /// populated with the exact arguments passed to this field as well.
+    ///
+    /// NOTE: This is renamed as the misspelled `ethers-etherscan/src/verify.rs`. The reason for
+    /// this is that Etherscan has had this misspelling on their API for quite a long time, and
+    /// changing it would break verification with arguments.
+    ///
+    /// For instances (e.g. blockscout) that might support the proper spelling, the field
+    /// `blockscout_constructor_arguments` is populated with the exact arguments passed to this
+    /// field as well.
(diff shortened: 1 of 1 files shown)


ethers-etherscan/src/verify.rs
```diff
@@ -24,14 +24,19 @@ pub struct VerifyContract {
...
+    /// verification with arguments.
+    /// For instances (e.g some blockscout instances) that might support the proper field
+    /// and not the misspelling, the private field `blockscout_constructor_arguments` is
+    /// populated with the exact arguments passed to this field as well.
+    #[serde(rename = "constructorArguements", skip_serializing_if = "Option::is_none")]
     pub constructor_arguments: Option<String>,
+    /// Properly spelled constructor arguments. This is needed as some blockscout instances
+    /// can identify the correct spelling instead of the typo version above.
+    #[serde(rename = "constructorArguments", skip_serializing_if = "Option::is_none")]
+    blockscout_constructor_arguments: Option<String>,
```
DaniPopes, line 39: Use `serde(alias = "...")`

```suggestion
    /// The constructor arguments for the contract, if any.
    ///
    /// NOTE: This is also parsed as the misspelled `constructorArguements`. The reason for this is that Etherscan has had
    /// this misspelling on their API for quite a long time, and changing it would break
    /// verification with arguments. Other services have fixed this typo, so we support both.
    #[serde(rename = "constructorArguments", alias = "constructorArguements", skip_serializing_if = "Option::is_none")]
    pub constructor_arguments: Option<String>,
```

DaniPopes (changes requested):

ethers-etherscan/src/verify.rs
Evalir: Oh yeah, happy to turn this field public—this shouldn't be breaking.

However, we can't use `alias` as this only works for deserialization, not serialization, so we'd end up in the same place as before (not serializing etherscan's `constructorArguements` correctly)

DaniPopes: Makes sense

DaniPopes (approved):

PR #2470, opened 2023-06-13
fix: prevent panics decoding too short input

WestXu:
Got hit by a runtime panic on this. Simple fix.

### DIFF
--- a/ethers-contract/src/base.rs
+++ b/ethers-contract/src/base.rs
@@ -197,7 +197,11 @@ impl BaseContract {
     }
 
     fn get_fn_from_input(&self, input: &[u8]) -> Result<&Function, AbiError> {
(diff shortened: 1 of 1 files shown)


ethers-contract/src/base.rs
```diff
@@ -197,7 +197,11 @@ impl BaseContract {
     }
 
     fn get_fn_from_input(&self, input: &[u8]) -> Result<&Function, AbiError> {
-        let sig: [u8; 4] = input[0..4].try_into().map_err(|_e| AbiError::WrongSelector)?;
+        let sig: [u8; 4] = input
+            .get(0..4)
+            .ok_or(AbiError::WrongSelector)?
```
mattsse, line 202: this is for `missing or wrong selector, which makes sense here.

mattsse (approved):
lgtm

PR #2477, opened 2023-06-15
Add bundler to README.md

Vid201:
Adding bundler to the section "Projects using ethers-rs".

This PR can be ignored if you're not accepting projects to that list.


mattsse (approved):
lgtm, 

all test failures unrelated

DaniPopes (approved):

PR #2482, opened 2023-06-19
feat(abigen): do not re-parse ABI at runtime

DaniPopes:
## Motivation

Print the parsed ABI as tokens into a constructor function

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (approved):
very nice!

hope some of it is useful for new alloy parser

PR #2483, opened 2023-06-20
feat: feat: add BlockOverrides for debug_traceCall

chikko80:
Implements BlockOverrides for debug_TraceCall according to [Geth BlockOverrides Implementation](https://github.com/ethereum/go-ethereum/blob/6d2aeb43d516fbe2cafd9e65df7ccbd885f861d3/internal/ethapi/api.go#L944)

## PR Checklist

-   [ ] Added Tests
-   [x ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/types/trace/geth.rs
+++ b/ethers-core/src/types/trace/geth.rs
@@ -212,9 +212,9 @@ pub struct GethDebugTracingCallOptions {
     pub block_overrides: Option<BlockOverrides>,
 }
 
-/// Bindings for additional `debug_traceTransaction` options
+/// Additional BlockOverrides Options
 ///
-/// See <https://geth.ethereum.org/docs/rpc/ns-debug#debug_tracetransaction>
+/// See <https://geth.ethereum.org/docs/rpc/ns-debug#debug_tracecall>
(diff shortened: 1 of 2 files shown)


ethers-core/src/types/trace/geth.rs
```diff
@@ -209,7 +209,29 @@ pub struct GethDebugTracingCallOptions {
     pub tracing_options: GethDebugTracingOptions,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub state_overrides: Option<spoof::State>,
-    // TODO: Add blockoverrides options
+    pub block_overrides: Option<BlockOverrides>,
```
DaniPopes, line 212:
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub block_overrides: Option<BlockOverrides>,
```

DaniPopes (changes requested):
Need to skip serializing if `None`

DaniPopes (changes requested):
@chikko80 friendly ping :^) Can you implement the suggestion above and rebase to latest master?

PR #2485, opened 2023-06-27
feat: implement eip2718/2930/1559 for receipts

yangby-cryptape:
## Motivation

With current implementation of `TransactionReceipt` serialization, the `receipts_root` in latest blocks on Ethereum mainnet could NOT be verified.

That's because the part of receipts serialization in [`EIP-2718`] is not implemented.

## Solution

I use the method `pub fn rlp(&self) -> Bytes` to instead of the `rlp::Encodable` implementation for `TransactionReceipt`.

Then, `TransactionReceipt` will have the same API as `Transaction`.
https://github.com/gakonst/ethers-rs/blob/df165b84229cdc1c65e8522e0c1aeead3746d9a8/ethers-core/src/types/transaction/response.rs#L155-L159

:warning: Since I removed the `rlp::Encodable` implementation, it will be a breaking change.

:warning: I didn't find any documents about feature `optimism`, so I implemented that as the same as other typed receipts.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

[`EIP-2718`]: https://eips.ethereum.org/EIPS/eip-2718#receipts


ethers-core/src/types/transaction/response.rs
```diff
@@ -508,10 +498,59 @@ impl PartialOrd<Self> for TransactionReceipt {
...
+                encoded.into()
+            }
+            #[cfg(feature = "optimism")]
+            Some(x) if x == U64::from(0x7E) => {
+                encoded.extend_from_slice(&[0x7E]);
+                encoded.extend_from_slice(rlp_bytes.as_ref());
+                encoded.into()
+            }
+            _ => rlp_bytes,
+        }
```
DaniPopes, line 546: Might also be worth encoding the type always if `<= 0x7f`?

```suggestion
        match self.transaction_type.map(|x| x.as_u64()) {
            Some(0x1 | 0x2) => return [&[ty as u8][..], &rlp_bytes[..]].concat().into(),
            #[cfg(feature = "optimism")]
            Some(0x7E) => return [&[ty as u8][..], &rlp_bytes[..]].concat().into(),
            _ => {}
        }
        rlp_bytes.into()
```

DaniPopes (changes requested):

PR #2491, opened 2023-07-01
Etherscan could also have a plain source mapping

wtdcode:
## Motivation

Always finding funny things when playing with Etherscan API, try:

```bash
curl "https://api.etherscan.io/api?module=contract&action=getsourcecode&address=0x68b26dcf21180d2a8de5a303f8cc5b14c8d99c4c"
```

This gives a plain source code mapping in the `SourceCode` field instead of plain source code.

## Solution

As shown.

### DIFF
--- a/ethers-etherscan/src/contract.rs
+++ b/ethers-etherscan/src/contract.rs
@@ -48,6 +48,8 @@ pub enum SourceCodeMetadata {
         #[serde(default, skip_serializing_if = "Option::is_none")]
         settings: Option<serde_json::Value>,
     },
+    /// Contains just mapped source code.
+    Sources(HashMap<String, SourceCodeEntry>),
     /// Contains only the source code.
     SourceCode(String),
 }
@@ -57,7 +59,10 @@ impl SourceCodeMetadata {
         match self {
             Self::Metadata { sources, .. } => {
                 sources.values().map(|s| s.content.clone()).collect::<Vec<_>>().join("\n")
-            }
+            },
+            Self::Sources(sources) => {
+                sources.values().map(|s| s.content.clone()).collect::<Vec<_>>().join("\n")
+            },
             Self::SourceCode(s) => s.clone(),
         }
     }
@@ -65,6 +70,7 @@ impl SourceCodeMetadata {
     pub fn language(&self) -> Option<SourceCodeLanguage> {
         match self {
             Self::Metadata { language, .. } => language.clone(),
+            Self::Sources(_)=> None,
             Self::SourceCode(_) => None,
         }
     }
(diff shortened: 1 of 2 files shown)


DaniPopes:
`language` and `settings` in `SourceCodeMetadata::Metadata` are tagged `#[serde(default)]`, so they would be set to `None` if they are not present. Does that contract not parse already?

wtdcode:
> `language` and `settings` in `SourceCodeMetadata::Metadata` are tagged `#[serde(default)]`, so they would be set to `None` if they are not present. Does that contract not parse already?

If you try with the link above, you will find that it's missing the key fields, i.e. just __plain__ source code mapping like:

```json
{ "a.sol" : "..." }
```

DaniPopes:
Ah alright, that makes sense. Shouldn't it be just `Sources(HashMap<String, SourceCodeEntry>,)` then? Can you also add a test for this case with that contract?

wtdcode:
> Ah alright, that makes sense. Shouldn't it be just `Sources(HashMap<String, SourceCodeEntry>,)` then? Can you also add a test for this case with that contract?

Fixed already. Regarding the test case, I didn't see any existing test cases actually requesting the network (actually no existings test cases for SourceCodeMetadata). Is it proper to make unit tests rely on network service?

wtdcode:
> > Ah alright, that makes sense. Shouldn't it be just `Sources(HashMap<String, SourceCodeEntry>,)` then? Can you also add a test for this case with that contract?
> 
> Fixed already. Regarding the test case, I didn't see any existing test cases actually requesting the network (even no existing test cases for SourceCodeMetadata). Is it proper to make unit tests rely on network service?

DaniPopes:
You can add one at the bottom of [`ethers-etherscan/tests/it/contract.rs`](https://github.com/gakonst/ethers-rs/blob/5faea1bb17fbb3662809c6d40d5b407e5a45360c/ethers-etherscan/tests/it/contract.rs)

wtdcode:
> You can add one at the bottom of [`ethers-etherscan/tests/it/contract.rs`](https://github.com/gakonst/ethers-rs/blob/5faea1bb17fbb3662809c6d40d5b407e5a45360c/ethers-etherscan/tests/it/contract.rs)

The problem is not about the place for tests, but how to test it? Request etherscan?

DaniPopes:
Yes, see `can_fetch_contract_source_tree_for_multi_entry_contract` in that file, you can copy it and replace the assertions

wtdcode:
Done

DaniPopes (approved):
Thanks for spotting this! I've fixed the tests in #2492.

PR #2504, opened 2023-07-11
feat(ethers-solc): add with_via_ir_minimum_optimization

spockP:
## Motivation

related to https://github.com/foundry-rs/foundry/pull/5349: 

> More and more people (including myself) are using the magical via-ir to solve the "stack too deep" issue. However, this would cause problems in test coverage as all the optimizations are disabled in test coverage.
> 
> The solidity team suggests working around this issue by enabling "via-ir" with a https://github.com/ethereum/solidity/issues/12533#issuecomment-1013073350
> 
> And this workaround is also recommended in the [FAQ](https://github.com/sc-forks/solidity-coverage/blob/89882a211e9f69ed6b1afc99b2bb9d56b857e686/docs/faq.md?plain=1#L85) of solcoverage.
> 
> In this pr, a new "--ir-minimum" option is added to the coverage subcommand, which enables "via-ir" with a minimum amount of optimization.

And @mattsse  suggests adding a function to the `Settings` type  in `ethers-solc`. 

## Solution

Add a function called `with_via_ir_minimum_optimization` for enabling `viaIR` with a minimum amount of optimization. 

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -453,6 +453,34 @@ impl Settings {
         self.set_via_ir(true)
     }
 
+    /// Enable `viaIR` and use the minimum optimization settings
+    ///
+    /// This is useful in the following scenarios:
+    /// - When compiling for test coverage, this can resolve the "stack too deep" error while still
+    ///   giving a relatively accurate source mapping
+    /// - When compiling for test, this can reduce the compilation time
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):
Thanks!

Issue #2186, opened 2023-02-24
ContractError::Revert excludes message and error code


mattsse:
opening this to track

https://github.com/gakonst/ethers-rs/pull/2172 introduced `ContractError::Revert`,

the `from_middleware_err` functions only extracts the data:
https://github.com/gakonst/ethers-rs/blob/7e12f3ba363158a2d0e2598523e92ead7beb9386/ethers-contract/src/call.rs#L120-L120

but the message and error code can contain additional context besides "reverted".

Suggestion:

* add `RevertErr(JsonRpcError)` new type that is used by the `Revert`
* custom impl Display for `RevertErr` with attempt to decode revert data

@prestwich

PR #2515, opened 2023-07-19
ContractError: added ability to retrieve  revert message from bytes

TheBigBlase:
## Motivation
This tries to retrieve revert data message from a ContractError from a JsonRpcError, if the RPC sends it.
this somewhat solves #2186 and completes #2392 

## Solution
We use the solidity revert model: https://docs.soliditylang.org/en/v0.8.12/control-structures.html#revert and parse the returned bytes accordingly. 

I don't know if the rust compiler will make the bin call <str>.split_at multiple times, so i left the code in the most readable state possible

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation (inline rustdoc) 
-   [ ] Breaking changes

### DIFF
--- a/ethers-contract/src/call.rs
+++ b/ethers-contract/src/call.rs
@@ -104,6 +104,39 @@ impl<M: Middleware> ContractError<M> {
         }
     }
 
+    /// extract reason msg from a revert error
+    ///
+    /// ## Returns
+    /// `None` if we couldnt decode
+    /// `Some(string)` if we could get the revert message. The string has no null bytes
+    ///
+    /// ## NOTE
+    ///
+    /// A lot of ethers RPC do not enable this by default. Make sure yours has it enabled.
+    /// for more details see https://docs.soliditylang.org/en/v0.8.12/control-structures.html#revert
+    pub fn get_revert_msg(&self) -> Option<String> {
+        self.as_revert().and_then(|data: &Bytes| {
+            let bytes = data.to_vec();
+            let mut bytes = bytes.as_slice();
+            let data_string: &[u8];
+
+            (_, bytes) = bytes.split_at(4); // first 4 bytes for function selector
+            (_, bytes) = bytes.split_at(32); // next 32 for data offset
+            (_, bytes) = bytes.split_at(32); // next 32 bytes for str len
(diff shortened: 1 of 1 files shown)


ethers-contract/src/call.rs
```diff
@@ -104,6 +104,39 @@ impl<M: Middleware> ContractError<M> {
...
+    pub fn get_revert_msg(&self) -> Option<String> {
+        self.as_revert().and_then(|data: &Bytes| {
+            let bytes = data.to_vec();
+            let mut bytes = bytes.as_slice();
+            let data_string: &[u8];
+
+            (_, bytes) = bytes.split_at(4); // first 4 bytes for function selector
+            (_, bytes) = bytes.split_at(32); // next 32 for data offset
+            (_, bytes) = bytes.split_at(32); // next 32 bytes for str len
+            (data_string, _) = bytes.split_at(32); // remaining 32 for string data
```
DaniPopes, line 126: You can just slice the `Bytes` with `data[100..]`, and you also have to add a length check to avoid panicking

ethers-contract/src/call.rs
```diff
@@ -104,6 +104,39 @@ impl<M: Middleware> ContractError<M> {
...
+    ///
+    /// A lot of ethers RPC do not enable this by default. Make sure yours has it enabled.
+    /// for more details see https://docs.soliditylang.org/en/v0.8.12/control-structures.html#revert
+    pub fn get_revert_msg(&self) -> Option<String> {
+        self.as_revert().and_then(|data: &Bytes| {
+            let bytes = data.to_vec();
+            let mut bytes = bytes.as_slice();
+            let data_string: &[u8];
+
+            (_, bytes) = bytes.split_at(4); // first 4 bytes for function selector
```
DaniPopes, line 123: Might be worth returning `None` if the selector is not `08c379a0` (`keccak256("Error(string)")`). You can achieve this by doing `bytes[..4] != [0x08, 0xc3, 0x79, 0xa0]`

DaniPopes (changes requested):

TheBigBlase:
I'm not sure that i did it in the most idiomatic way, i mean i used `return` in the middle of a function. 
Please let me know ; and thank you for taking some time looking into this

PR #2523, opened 2023-07-24
chore: support latest svm

Evalir:
## Motivation

Due to https://github.com/alloy-rs/svm-rs/commit/9b0fa3a751312c153d9bb785b48a25d9bf679e40#diff-4ac4c674e2696c0c39e91b7ba5af3f65b6ffb1a0fa9b90d3f9930a14254ecb98L126 some changes are needed to point to the correct var here, as `SVM_HOME` does not exist anymore

## Solution

Use `SVM_DATA_DIR`

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-solc/src/compile/mod.rs
+++ b/ethers-solc/src/compile/mod.rs
@@ -353,7 +353,7 @@ impl Solc {
         let _lock = take_solc_installer_lock();
 
         // load the local / remote versions
-        let versions = utils::installed_versions(svm::SVM_HOME.as_path()).unwrap_or_default();
+        let versions = utils::installed_versions(svm::SVM_DATA_DIR.as_path()).unwrap_or_default();
 
         let local_versions = Self::find_matching_installation(&versions, sol_version);
         let remote_versions = Self::find_matching_installation(&RELEASES.1, sol_version);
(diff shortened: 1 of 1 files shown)


mattsse (approved):
can also bump svm in cargo.lock file

Evalir:
we don't commit the lockfile ser

ethers-solc/src/compile/mod.rs
```diff
@@ -218,10 +218,21 @@ impl Solc {
...
-        home::home_dir().map(|dir| dir.join(".svm"))
+        match home::home_dir().map(|dir| dir.join(".svm")) {
+            Some(dir) => {
+                if !dir.exists() {
+                    dirs::data_dir().map(|dir| dir.join("svm"))
+                } else {
+                    Some(dir)
+                }
+            }
+            None => dirs::data_dir().map(|dir| dir.join("svm")),
```
mattsse, line 234: Isn't this also just SVM_DATA_DIR now?

can we use this instead?

mattsse (changes requested):

ethers-solc/src/compile/mod.rs
Evalir: Ah, the problem with using this directly is that we'd lock the crate to use svm always instead of conditionally (it's currently a feature)

mattsse:
yeah unrelated geth snafu

@DaniPopes mind merging this wen you find the time?

mattsse (approved):

DaniPopes (approved):

PR #2527, opened 2023-07-27
generic Provider allowing relaxed json rpc handling

aathan:
## Motivation

Sometimes you have no choice but to interact with json rpc endpoints that don't exactly adhere to the standards. Innocuous departures from the standard should be tolerable.

## Solution

I'm new to Rust, so I do not claim this implementation is ideal. From what I could tell the only way to achieve this is to modify the deserializer, and in order to do this without additional runtime state variables, I implemented this via a const bool generic type parameter on Response and Provider. A pull request I am submitting to https://github.com/foundry-rs/foundry will reference this pull request.

Because I export a type name Http which is equivalent to the old type, I believe this is not a breaking change; but that depends on what level of abstraction your source is at.

## PR Checklist

-   [x] Added Tests -- the existing tests cover the implementation except in the relaxed/loose case which I tested manually
-   [x] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-providers/src/rpc/transports/common.rs
+++ b/ethers-providers/src/rpc/transports/common.rs
@@ -96,7 +96,7 @@ impl<'a, T> Request<'a, T> {
 
 /// A JSON-RPC response
 #[derive(Debug)]
-pub enum Response<'a> {
+pub enum Response<'a, const STRICT:bool > {
     Success { id: u64, result: &'a RawValue },
     Error { id: u64, error: JsonRpcError },
     Notification { method: &'a str, params: Params<'a> },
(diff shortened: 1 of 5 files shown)


mattsse (commented):
I'm supportive of this feature, because this really comes up a lot in foundry, unfortunately, especially with new testnets.

having an option to allow malformed json rpc would be useful.

defer to @DaniPopes re changes/how to proceed

aathan:
Since there's interest to accept this PR I'm putting in the time to do a couple of minor cleanups. E.g., the name of the typevar is now RELAXED and LooseHttp is RelaxedHttp.

Note that the real thrust of this change is to allow these types to be generics so that behavior can be parameterized in general. To do this RELAXED would be a more complex type, but the bones are there.

Evalir (commented):
would love to get this in cc @DaniPopes 

@aathan anything else needs cleanup here?=

Evalir:
👋 checking in here

ethers-providers/src/rpc/transports/common.rs
```diff
@@ -109,120 +109,122 @@ pub struct Params<'a> {
     pub result: &'a RawValue,
 }
 
-// FIXME: ideally, this could be auto-derived as an untagged enum, but due to
-// https://github.com/serde-rs/serde/issues/1183 this currently fails
-impl<'de: 'a, 'a> Deserialize<'de> for Response<'a> {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+struct ResponseVisitor<'b,const RELAXED:bool>(&'b ());
```
DaniPopes, line 112: This should use `PhantomData`

```suggestion
struct ResponseVisitor<'b,const RELAXED:bool>(PhantomData<&'b ()>);
```

DaniPopes (changes requested):

DaniPopes (changes requested):
Needs `cargo +nightly fmt`.

This is a pretty big breaking change, I think this is better implemented with a run time check that is opt-in at construction of the provider, like a `relaxed(bool)` method.

aathan:
May I ask what would break exactly? It's been a while since I wrote/submitted this but as I recall (a) the ResponseVisitor didn't seem like something with external dependents (b) a pub type ResponseVisitor = GenResponseVisitor<false>  could be done if it did have external dependencies and (c) granted, the performance implications may be fairly small, but it seemed wasteful to do a check at runtime for something that is selected at compile time -- but maybe it'd be optimized out anyway? Also granted, the actual check probably only happens in the rare instance of a malformed response that contains the method. I can try to find the time to make the change if you feel strongly about it.

aathan:
`cargo +nightly fmt` done and pushed.

aathan:
Ok I brought this PR up to date, hopefully the changes to ethers-rs can be approved and then this can be merged.

This is an imperfect PR that does not address some bigger problems I see in the codebase. Specifically, that remote addresses (urls, ips, whatever) are stored basically bare, and then, providers are instantiated from them.

When those new providers are created, there's no attempt to mimic the connectivity features relevant to successfully communicating with that far end. Timeouts, relaxed/strict rpc handling, ssl certs, whatever.

Every place that stores a "fork_url" w/ or w/o separate connectivity parameters should probably instead store a "RemoteConnInfo" struct that either contains or references enough info from which to build connectivity providers. I did notice there are both ForkUrl and RpcUrl types floating around.

PR #2529, opened 2023-07-28
feat: add engine api-compatible bearer token generation

Rjected:
## Motivation

I was looking for a way to easily make engine API calls using ethers-rs / `cast`, and realized that there is no JWT signing / encoding, the token would have to be generated elsewhere, before using `Authorization::Bearer`.

## Solution

This adds code used by lighthouse to communicate with an EL's engine API:
https://github.com/sigp/lighthouse/blob/dfcb3363c757671eb19d5f8e519b4b94ac74677a/beacon_node/execution_layer/src/engine_api/auth.rs#L132

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-providers/src/rpc/transports/common.rs
+++ b/ethers-providers/src/rpc/transports/common.rs
@@ -5,6 +5,7 @@ use ethers_core::{
     abi::AbiDecode,
     types::{Bytes, U256},
 };
+use jsonwebtoken::{encode, errors::Error, get_current_timestamp, Algorithm, EncodingKey, Header};
 use serde::{
     de::{self, MapAccess, Unexpected, Visitor},
     Deserialize, Serialize,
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm

ethers-providers/src/rpc/transports/common.rs
```diff
@@ -270,6 +271,101 @@ impl fmt::Display for Authorization {
...
+    /// Decode the given string from hex (no 0x prefix), and attempt to create a key from it.
+    pub fn from_hex(hex: &str) -> Result<Self, String> {
+        let bytes = hex::decode(hex).map_err(|e| format!("Invalid hex: {}", e))?;
+        Self::from_slice(&bytes)
+    }
+
+    /// Returns a reference to the underlying byte array.
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0
+    }
```
DaniPopes, line 307: Needs a way to get the underlying bytes back

```suggestion
    /// Returns a reference to the underlying byte array.
    pub fn as_bytes(&self) -> &[u8; JWT_SECRET_LENGTH] {
        &self.0
    }
    
    pub fn into_bytes(self) ...
```

DaniPopes (changes requested):

ethers-providers/src/rpc/transports/common.rs
Rjected: makes sense, added

DaniPopes (approved):

PR #2530, opened 2023-07-31
Don't overwrite remappings if provided by Settings

wtdcode:
## Motivation

remappings should be simply overwritten if there is any, which is especially the case when downloading standard json from etherscan.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -159,7 +159,11 @@ impl CompilerInput {
         if self.is_yul() {
             warn!("omitting remappings supplied for the yul sources");
         } else {
-            self.settings.remappings = remappings;
+            if self.settings.remappings.is_empty() {
+                self.settings.remappings = remappings;
+            } else {
+                warn!("remappings in settings conflicts with path configurations");
+            }
         }
         self
     }


ethers-solc/src/artifacts/mod.rs
```diff
@@ -159,7 +159,11 @@ impl CompilerInput {
         if self.is_yul() {
             warn!("omitting remappings supplied for the yul sources");
         } else {
-            self.settings.remappings = remappings;
+            if self.settings.remappings.is_empty() {
+                self.settings.remappings = remappings;
+            } else {
+                warn!("remappings in settings conflicts with path configurations");
+            }
```
mattsse, line 166: I disagree here, 
I'd expect that a function called `with_remappings` would always set the remappings

wtdcode: That's the point where I'm open to discussion.

Another way is to modify here: https://github.com/gakonst/ethers-rs/blob/594627dc1c3b490ba8f513f8f5e23d11448cbcf8/ethers-solc/src/compile/project.rs#L546 by checking `Settings` there. Are you fine with that?

DaniPopes:
@wtdcode The assumption is that the function always sets the remappings. Yul doesn't have remappings (I think?), and in practice it's almost always false. What would this fix?

wtdcode:
> @wtdcode The assumption is that the function always sets the remappings. Yul doesn't have remappings (I think?), and in practice it's almost always false. What would this fix?

https://github.com/gakonst/ethers-rs/blob/28bb3c02a585c52fb56ee5b3c8a359602597187c/ethers-solc/src/compile/project.rs#L543

When compiling a project, it simply overwrites all existing remappings (in many cases, empty remappings), which overwrites the remappings inherited from the settings.

PR #2535, opened 2023-08-03
fix(`solc`): do not filter remappings with `cwd`

Evalir:
## Motivation

This is an upstream fix for a few foundry compatibility problems we had due to remapping contexts being introduced.

Filtering with `cwd` at first makes sense—but it filters out contexts that we _do_ wanna keep when different dependencies use both the same library and it's a js-style setup.

## Solution

Do not use `cwd` and revert to old way of resolving remappings.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-solc/src/config.rs
+++ b/ethers-solc/src/config.rs
@@ -328,25 +328,21 @@ impl ProjectPathsConfig {
     pub fn resolve_library_import(&self, import: &Path) -> Option<PathBuf> {
         // if the import path starts with the name of the remapping then we get the resolved path by
         // removing the name and adding the remainder to the path of the remapping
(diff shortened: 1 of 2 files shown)


mattsse (approved):
as discussed, lgtm

pinging @DaniPopes for merge 🙏

or should we hold off until we confirmed foundry fix @Evalir ?

Evalir:
yep i'll test a bit more @mattsse — but this should work

maybe we can loop @onbjerg if he's not too busy so we have more eyes? this is super easy to get wrong

Evalir:
closing—found easier fix and this is not entirely correct

Issue #2532, opened 2023-08-02
Support wasm32 environments that are not related to browsers


oblique:
### Why?

I'm working in a project that ethers will be used within [Internet Computer](https://internetcomputer.org/docs/current/developer-docs/) environment, which is using WebAssembly.

### The issue

What I'm going to describe is an issue in various projects in Rust ecosystem and ethers happens to have it too. At the end I will propose a solution (which I will implement), however it is not going to be backward compatible in the case of `wasm32` targets.

`wasm32` is an architecture but it does not describe the environment that is going to be used. Most projects assume that the environment is always a browser. Within browser environment it is valid to use `wasm-bindgen`, `js-sys`,  and `web-sys`, but this is not valid in any other environments. To be more correct `wasm-bindgen` and `js-sys` can be used only in environments with JavaScript and `web-sys` in browser environments.

`getrandom` solved this issue by requiring from the user to [enable the `js` feature flag manually](https://docs.rs/getrandom/latest/getrandom/#webassembly-support). Because of this a user needs to add the feature in the top-level crate otherwise they will get a [compilation error](https://github.com/rust-random/getrandom/blob/2e483d68aaa57168a84489349d6473b492e05478/src/lib.rs#L283-L288).

### Proposal

In my opinion something similar needs to be done in ethers too. I'm proposing the following:

* Introduce two new feature gates: `web-sys` and `ic` (for Internet Computer).
* Both feature gates must be disabled by default.
* In case of `wasm32-unknown-unknown` only one of the flags must be enabled. If none of them or both of them are enabled then we trigger `compile_error!`.
* All `wasm32` dependencies must be optional and feature flags will enable what is needed.
* `reqwest` and `futures-timer` crates have the same issue so we need to do the following:
	* In case of `wasm32` we have them as optional and we enable them with `web-sys` flag.
	* In case of `not(wasm32)` we have them as required.
	* In case of `wasm32` and `ic` flag:
		* `reqwest` will be replaced with `ic-cdk`.
		* `futures-timer` will be replaced with `ic-cdk-timers`.
		* `wasm-bindgen-futures` will be replaced with `ic-cdk`.
* Internet Computer does not support WebSockets, so implementation will not be provided if `ic` is enabled.

### Other notes

* Should I create a crate for this compatibility layer? If yes, do you have any name in mind?
* `web-sys` can also be named as `browser`. The `js` is not a good option since `web-sys` is used by `reqwest`.
[...]

PR #2536, opened 2023-08-05
feat(ethers-contract): Make ethers-providers optional

oblique:
## Motivation

I'm working in a very restricted WASM environment (see #2532), which I can not use `ethers-providers` crate. However I would like to use `ethers_contract::EthCall` derive functions.

## Solution

With this PR, I introduced the `providers` flag in `ethers-contract`, `ethers-contract-abigen`, and `ethers-contract-derive`. Anything related to `ethers-providers` is now enabled with the flag.

`providers` flag was added in the default features of `ethers-contract`.

If a user adds `ethers` or `ethers-middleware` crates, then `providers` is always enabled, even if user disabled the default flags. This is because they have `ethers-providers` as a dependency and there is no reason to make the flag optional.

If a user adds `ethers-contract` and disables default features, then they need to add `providers` feature to get the same functionality as before. - Is this edge case a major breaking change?

If you have any other ideas how this can be better, let me know.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


ethers-contract/src/multicall/middleware.rs
oblique, line 1: Content of `ethers-contract/src/multicall/mod.rs` was moved here

ethers-contract/src/event_core.rs
oblique, line 1: Content from `ethers-contract/src/event.rs` was moved here. Nothing else was changed.

oblique:
The tests that are failing are not related to my changes and they failing before the changes on my machine.

oblique:
I found that foundry was responsible for the test failures. The tests are passing with version `nightly-e05b9c75b4501d5880764948b61db787f3dd7fe0` and they fail with version `nightly-2d87c0c2fcc47088feecc72721c46d8e07e3c220`

oblique:
With the latest foundry nightly the tests passed, so this PR is ready. Any comments are welcome.

ethers-contract/src/lib.rs
```diff
@@ -3,24 +3,25 @@
 #![warn(missing_docs)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
-#[path = "contract.rs"]
-mod _contract;
-pub use _contract::{Contract, ContractInstance};
+macro_rules! if_providers {
```
zvolin, line 6: I'd prefer having `pub(crate) use` for this macro and importing/calling by full path on call site

oblique: I want to avoid importing of this specific macro

ethers-contract/Cargo.toml
```diff
@@ -23,9 +23,8 @@ rustdoc-args = ["--cfg", "docsrs"]
 all-features = true
 
 [dependencies]
-ethers-providers.workspace = true
+ethers-providers = { workspace = true, optional = true }
 ethers-core.workspace = true
-ethers-signers.workspace = true
```
DaniPopes, line 28: Is this not needed anymore?

ethers-contract/ethers-contract-abigen/Cargo.toml
```diff
@@ -50,6 +50,7 @@ tokio = { workspace = true, optional = true }
 url = { workspace = true, optional = true }
 
 [features]
+providers = []
```
DaniPopes, line 53: This should be enabled by default, likewise for `ethers-contract-derive`

DaniPopes (changes requested):
This is great! Just a few things

ethers-contract/Cargo.toml
oblique: No, it is only used in `tests/it/abigen.rs`

ethers-contract/ethers-contract-abigen/Cargo.toml
oblique: Done

DaniPopes (approved):

PR #2538, opened 2023-08-07
feat: impl PubsubClient for  MockProvider

dereksione:
## Motivation

The current implementation of mock provider doesn't support any of the eth JSON RPC endpoints that establish subscriptions. Supporting this would allow for unit tests and smoke tests for implementations that depend on these features.

## Solution

This implementation extends the idea of the mock provider being a sink of pre-rendered responses piped in before beginning the testing process. There is a simple public facing function `setup_subscription` that takes all the previously pushed responses and converts them into a stream that can be subscribed to using any of the eth JSONRPC requests that create such subscriptions. 

Internally, the implementation is a little convoluted because the `PubsubClient` takes an immutable reference to `self`. The requirements of the task is that we need the mutable sender of a stream that we can push things to (and this cannot be inside the PubsubClient trait impl). The solution implemented here is uses an intermediary stream that is initialized before the `subscribe` function is called, and it forwards all of its data to the stream that is returned by the `subscribe` function in the `PubsubClient` trait.

## PR Checklist

-   [x] Added Tests
-   [x] Added easy to use public facing API


ethers-providers/src/rpc/provider.rs
```diff
@@ -827,7 +827,7 @@ impl<P: JsonRpcClient> Middleware for Provider<P> {
                         };
                         let data = self.call(&tx.into(), None).await?;
                         if decode_bytes::<Address>(ParamType::Address, data) != owner {
-                            return Err(ProviderError::CustomError("Incorrect owner.".to_string()))
+                            return Err(ProviderError::CustomError("Incorrect owner.".to_string()));
```
DaniPopes, line 830: Please format code with nightly (`cargo +nightly fmt`)

ethers-providers/src/rpc/transports/mock.rs
```diff
@@ -72,6 +83,31 @@ impl JsonRpcClient for MockProvider {
     }
 }
 
+impl PubsubClient for MockProvider {
```
DaniPopes, line 86: I believe these implementations will need to be gated with `not(target_arch = "wasm32")`, see CI

DaniPopes (changes requested):

DaniPopes (changes requested):
Also needs a rebase and clippy fix.

PR #2549, opened 2023-08-11
perf: buffer writes when serializing json

DaniPopes:
## Solution

Default is 8KiB, bumped to 256KiB for solc cache files since it seems more adequate

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (approved):
nice

PR #2550, opened 2023-08-11
perf(solc): memoize `is_dirty` results

DaniPopes:
## Motivation

Hot code path that contributes to ~30% of the time spent preparing for a compiler run in forge

cc @mattsse 

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (approved):
cool!
thinks looks correct to me, even if the cache tests pass I'd feel a bit more confident if we can try this in foundry first via a patch PR first before merging

DaniPopes:
foundry-rs/foundry#5607

mattsse:
let's do it!

PR #2553, opened 2023-08-12
fix: skip checksums on windows for now

mattsse:
ref https://github.com/foundry-rs/foundry/issues/5601

https://github.com/alloy-rs/svm-rs/issues/95

> on windows solc downloads are zip files, the checksums are for the zip files
> however, we extract the exe and are hashing this file in ethers-solc

this disables the checksum until we can properly do this, this will make it easier for windows users to use forge without redownloading all files over and over again


Evalir (approved):
oof good catch, this slows down windows users a ton

DaniPopes (approved):

Issue #2500, opened 2023-07-07
Receipt RLP is Inaccurate


MonkeyKing-1:
**Version**
2.0.7

**Platform**
Darwin Macbook-Air 22.4.0 Darwin Kernel Version 22.4.0: Mon Mar  6 21:01:02 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T8112 arm64

**Description**
rlp_bytes() for TransactionReceipts is faulty (In particular, it lacks the prefix and also omits the intermediate root for pre-Byzantium blocks).

```
impl rlp::Encodable for TransactionReceipt {
    fn rlp_append(&self, s: &mut RlpStream) {
        s.begin_list(4);
        rlp_opt(s, &self.status);
        s.append(&self.cumulative_gas_used);
        s.append(&self.logs_bloom);
        s.append_list(&self.logs);
    }
}
```

Above is the code found in ethers-core-2.0.7/src/types/transaction/response.rs.

jonathanpwang:
bump on this

PR #2571, opened 2023-08-26
fix: accurate receipt rlp

MonkeyKing-1:
## Motivation
Fixes [#2500 ](url)

## Solution
New rlp() function for receipts that mirrors the rlp() function for transactions. Gives the rlp encoding of the receipt that is consistent with the receipt root for all blocks.

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [x] Breaking changes

### DIFF
--- a/ethers-core/src/types/transaction/response.rs
+++ b/ethers-core/src/types/transaction/response.rs
@@ -479,7 +479,11 @@ pub struct TransactionReceipt {
 impl rlp::Encodable for TransactionReceipt {
     fn rlp_append(&self, s: &mut RlpStream) {
         s.begin_list(4);
-        rlp_opt(s, &self.status);
+        if let Some(post_state) = self.root {
+            s.append(&post_state);
+        } else {
+            s.append(&self.status.expect("No post-state or status in receipt"));
+        }
         s.append(&self.cumulative_gas_used);
         s.append(&self.logs_bloom);
         s.append_list(&self.logs);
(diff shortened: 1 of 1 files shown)


ethers-core/src/types/transaction/response.rs
```diff
@@ -508,10 +498,44 @@ impl PartialOrd<Self> for TransactionReceipt {
     }
 }
 
+impl TransactionReceipt {
+    pub fn rlp(&self) -> Bytes {
```
gakonst, line 502: We still want to implement the trait right? Functionality seems fine. Would the test case below fail w/o this PR?

MonkeyKing-1: I think I didn't implement the trait because Transaction seemed to not implement the trait (I'll check again). Regardless, it's pretty easy to implement the trait. The new test case would fail w/o the PR because the old rlp encoding did not include the transaction type.

gakonst: Got it! Helpful context. Please add the trait implementation, and if you see other structs that are missing it, please also go for these as well. Thanks for taking this on!!!

MonkeyKing-1: The traits should be implemented now for both Transaction and TransactionReceipt. I didn't implement Decodable for TransactionReceipt because I feel like I don't have enough familiarity with the documentation of all the structs and functions involved.

PR #2572, opened 2023-08-27
PubSubItem deserialize

AmazingGame:
…ng in errors.

json example:
"{\"jsonrpc\":\"2.0\",\"id\":268,\"result\":null,\"error\":{\"code\":-32000,\"message\":\"tracing failed: fee cap less than block base fee: address 0x4a4BA8a759e7217a1ff38749Aa6E73564C1D8a01, gasFeeCap: 11470668621 baseFee: 11484931392\"}}"

If the json response contains both result and errors, it's better to return errors instead of "response must be a success/error or notification object"

Also, after returning "de::Error::custom", in

```
  pub async fn handle_text(&mut self, t: String) -> Result<(), WsClientError> {
         trace!(text = t, "Received message");
         match serde_json::from_str(&t) {
             Ok(item) => {
                 trace!(%item, "Deserialized message");
                 let res = self.handler.unbounded_send(item);
                 if res.is_err() {
                     return Err(WsClientError::DeadChannel)
                 }
             }
             Err(e) => {
                 error!(e = %e, "Failed to deserialize message");
             }
         }
         Ok(())
     }

```
An error should be returned instead of just printing the error, which would cause the request to wait indefinitely.

## PR Checklist

-   [x ] Added Tests
-   [x ] Added Documentation
-   [x ] Breaking changes

### DIFF
--- a/ethers-providers/src/rpc/transports/ws/types.rs
+++ b/ethers-providers/src/rpc/transports/ws/types.rs
@@ -67,7 +67,7 @@ impl<'de> Deserialize<'de> for PubSubItem {
                     match key {
                         "jsonrpc" => {
                             if jsonrpc {
-                                return Err(de::Error::duplicate_field("jsonrpc"))
+                                return Err(de::Error::duplicate_field("jsonrpc"));
                             }
 
                             let value = map.next_value()?;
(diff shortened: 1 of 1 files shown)


ethers-providers/src/rpc/transports/ws/types.rs
```diff
@@ -67,54 +67,54 @@ impl<'de> Deserialize<'de> for PubSubItem {
                     match key {
                         "jsonrpc" => {
                             if jsonrpc {
-                                return Err(de::Error::duplicate_field("jsonrpc"))
+                                return Err(de::Error::duplicate_field("jsonrpc"));
```
DaniPopes, line 70: Please format with `cargo +nightly fmt` (and a recent nightly version)

DaniPopes (changes requested):

PR #2582, opened 2023-09-02
fix(solc): empty 'Solc error: ' message

DaniPopes:
## Motivation

When the Solc executable fails and the stderr is empty we throw `Solc error: ` which is not very helpful

cc @mattsse 

## Solution

Make it print `Solc exited with {output.status}\n{output.stderr || output.stdout || "<empty output>"}`

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (approved):
lgtm

PR #2600, opened 2023-09-15
chore(solc): slightly improve resolution error msg

DaniPopes:


ethers-solc/src/resolver/mod.rs
```diff
@@ -903,32 +905,27 @@ impl Node {
...
-                v.parse().map_err(|err| SourceVersionError::InvalidVersion(v.to_string(), err))?;
-
-            if !all_versions.iter().any(|v| req.matches(v.as_ref())) {
-                return if offline {
-                    Err(SourceVersionError::NoMatchingVersionOffline(req))
-                } else {
-                    Err(SourceVersionError::NoMatchingVersion(req))
-                }
-            }
+    ) -> std::result::Result<(), SourceVersionError> {
```
mattsse, line 913: hmm, why this change?

DaniPopes: Complicates for no reason, it's a version validation

mattsse (approved):
right the None wasn't even used

PR #2602, opened 2023-09-15
fix(oracle): update helper function to use correct conversion maths

0xfourzerofour:
When converting between gwei and U256 it was rounding to the nearest whole number based on the decimal. For example if you added 2.49, the result would be 20000000 which is not idea

## Solution

Multiplied the f64 before the rounding occurs

## PR Checklist

-   [ yes] Added Tests
-   [no ] Added Documentation
-   [no ] Breaking changes


0xfourzerofour:
@FrankieIsLost

mattsse (approved):
lgtm

PR #2606, opened 2023-09-19
feat: add `dependencies` to `MultiBindingsInner`

### LINKED ISSUE #2599: feat: derive `Serialize` and `Deserialize` for `Event`, `{}Filter`, `Call`, and `Errors` in contract bindings
When working with contract bindings, it would be nice to be able to use `serde` to manage these structs/enums. For instance, if I was doing something like this:
```rust
let mut stream = event.stream().await.unwrap();
                while let Some(Ok(event)) = stream.next().await {
                    println!("Event: {:?}", serde_json::to_string(&event));
                }
```
it would be great to be able to do this serialization so I can, for example, output this all to a CSV nicely.

For instance, we can just add `Serialize` in the output for the bindings such as here:
```rust
fn expand_events_enum(&self) -> TokenStream {
        let variants = self
            .abi
            .events
            .values()
            .flatten()
            .map(|e| {
                event_struct_name(&e.name, self.event_aliases.get(&e.abi_signature()).cloned())
            })
            .collect::<Vec<_>>();

        let enum_name = self.expand_event_enum_name();

        let mut derives = self.expand_extra_derives();
        let params =
            self.abi.events.values().flatten().flat_map(|err| &err.inputs).map(|param| &param.kind);
        util::derive_builtin_traits(params, &mut derives, false, true);

        let ethers_core = ethers_core_crate();
        let ethers_contract = ethers_contract_crate();

        quote! {
            #[doc = "Container type for all of the contract's events"]
            #[derive(Clone, #ethers_contract::EthAbiType, #derives)]
            pub enum #enum_name {
                #( #variants(#variants), )*
            }
            \\ ~ snip
```
we can just replace `#[derive(Clone, #ethers_contract::EthAbiType, #derives)]` with `#[derive(Clone, Serialize, Deserialize, #ethers_contract::EthAbiType, #derives)]`. Note that this is located [here](https://github.com/gakonst/ethers-rs/blob/c702e6cdece63f3c47f3476c626024c11f4b8cec/ethers-contract/ethers-contract-abigen/src/contract/events.rs#L54-L77). 

**Additional context**
Though I just showed the potential solution for `Event`s, this could be applied to `Call`s, `Error`s, and `{}Filter` objects as well.

Autoparallel:
## Motivation
This seeks to close #2599 and should link to [Foundry PR #5836](https://github.com/foundry-rs/foundry/pull/5836). The intent is to be able to allow for `serde::{Deserialize, Serialize}` to be derived on contract bindings either via module or crate. 

PR here has the intent of providing support for an additional dependency in the `Cargo.toml` generated when bindings are output as a crate.

## Solution
This solution here seems a bit simplistic and forcefully assumes the end user will want a specific version of `serde` as an included dependency. It could be good, instead, to:
- Allow for dependencies to be chosen if there are extra derives added to the crate.
- Also forcefully add the `serde` derives to the contract bindings at this level.
- Check for most recent `serde` version or the version used within `ethers-rs` directly.

Let me know if we wish to make any of the above changes or otherwise!

## PR Checklist

-   [ ] Added Tests
-   [x] Added Documentation
-   [x] Breaking changes

### DIFF
--- a/ethers-contract/ethers-contract-abigen/src/multi.rs
+++ b/ethers-contract/ethers-contract-abigen/src/multi.rs
@@ -144,7 +144,11 @@ impl MultiAbigen {
     /// Build the contract bindings and prepare for writing
     pub fn build(self) -> Result<MultiBindings> {
         let format = self.abigens.iter().any(|gen| gen.format);
-        Ok(MultiBindings { expansion: MultiExpansion::from_abigen(self.abigens)?.expand(), format })
+        Ok(MultiBindings {
+            expansion: MultiExpansion::from_abigen(self.abigens)?.expand(),
+            format,
+            dependencies: vec![],
+        })
     }
 }
 
@@ -299,7 +303,12 @@ impl MultiExpansionResult {
     }
 
     /// Converts this result into [`MultiBindingsInner`]
-    fn into_bindings(mut self, single_file: bool, format: bool) -> MultiBindingsInner {
+    fn into_bindings(
+        mut self,
+        single_file: bool,
+        format: bool,
+        dependencies: Vec<String>,
+    ) -> MultiBindingsInner {
         self.set_shared_import_path(single_file);
         let Self { contracts, shared_types, root, .. } = self;
         let bindings = contracts
@@ -333,12 +342,7 @@ impl MultiExpansionResult {
             None
         };
 
-        MultiBindingsInner {
-            root,
-            bindings,
-            shared_types,
-            dependencies: vec![String::from(r#"serde = "1.0.188""#)],
-        }
+        MultiBindingsInner { root, bindings, shared_types, dependencies }
     }
 }
 
@@ -371,6 +375,7 @@ impl MultiExpansionResult {
 pub struct MultiBindings {
     expansion: MultiExpansionResult,
     format: bool,
+    dependencies: Vec<String>,
 }
 
 impl MultiBindings {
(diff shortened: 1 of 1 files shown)


ethers-contract/ethers-contract-abigen/src/multi.rs
```diff
@@ -333,7 +333,12 @@ impl MultiExpansionResult {
             None
         };
 
-        MultiBindingsInner { root, bindings, shared_types }
+        MultiBindingsInner {
+            root,
+            bindings,
+            shared_types,
+            dependencies: vec![String::from(r#"serde = "1.0.188""#)],
```
mattsse, line 340: hmm I don't think we serde by default here

what we need is a setter function just like `MultiBindings::format`

https://github.com/gakonst/ethers-rs/blob/a65abb57b2ee63470a9cf7c4fc98faeaaee8db53/ethers-contract/ethers-contract-abigen/src/multi.rs#L395-L402

so that this can be set manually:

`MultiBindings::dependencies(impl IntoIter<impl Into<String>) -> Self`

mattsse (changes requested):
almost there :)

Autoparallel:
@mattsse you think this is good a enough approach?

I'll fix linter issues. Any tests or anything you need me to add here? Doesn't seem breaking!

mattsse:
just want this addressed, then this is good to merge

https://github.com/gakonst/ethers-rs/pull/2606#discussion_r1329910564

Autoparallel:
Ah okay I will get to it. My jet lagged brain missed your comment! 😅

Autoparallel:
Alright, I think that does it. I appreciate the help and I see how that externalizes it now and it seems to be working with a slight change in Foundry now too.

Autoparallel:
@mattsse just checking in on this, I just ran the formatter and hoping this passes now.

mattsse (approved):
lgtm

gakonst (approved):
really sorry for the delay - good improvement, thank you

PR #2607, opened 2023-09-19
fix(solc): use Solc::version_req to parse version req

mattsse:
use custom Solc::version_req to parse version req because semver Version req differs from solidity's format


DaniPopes (approved):

Evalir (approved):
oof nice catch

PR #2617, opened 2023-09-28
remove arbitrary precision feature

### LINKED ISSUE #2206: serde_json arbitrary_precision flag causes conflicts
**Version**
current master branch, this is NOT an issue with 1.0.2.

**Description**
In the following [commit](https://github.com/gakonst/ethers-rs/commit/4b685621ed3a4b827276a13107f30457a19379db) there was an update to use the `serde_json` `arbitrary_precision` feature. This feature has been known to cause problems with the `serde_json` macros as documented [`here`](https://github.com/serde-rs/json/issues/505), with the currently tracking issue [here](https://github.com/serde-rs/serde/issues/1183). What will happen is that anyone that uses the `ethers` package will see problems when attempting to use the `tag` `untagged` and `flatten` macros. The open issues mentioned show multiple examples of this. Below is an example that I have.

I tried this code:
[dependencies]
serde_json = { version = "1.0.64", default-features = false, features = ["arbitrary_precision"] }

```rust
    #[test]
    fn temp() {
        struct M {
            value: f32,
        }

        impl<'de> Deserialize<'de> for M {
            fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
            where
                D: serde::Deserializer<'de>,
            {
                #[derive(Debug, Deserialize)]
                #[serde(untagged)]
                enum TempM {
                    Single(f32),
                    Sum { one: f32, two: f32 },
                }

                match TempM::deserialize(deserializer)? {
                    TempM::Single(value) => Ok(M { value }),
                    TempM::Sum { one, two } => Ok(M { value: one + two }),
                }
            }
        }

        let input = r#"10.1"#;
        let result: M = serde_json::from_str(input).unwrap();
    }
```

I expect in the example above for the `result` to resolve to `M { value: 10.1 }`

Instead i get the following error:
```panicked at 'called `Result::unwrap()` on an `Err` value: Error("data did not match any variant of untagged enum TempM", line: 0, column: 0)'```

When disabling the `arbitrary_precision` flag this correctly resolves.

mattsse:
closes #2206

this feature is kinda horrible and we should not force anyone into using it.

this is also only useful when deserializing json large numbers but since u256 are supposed to be hex it's fine to not support deserializing arbitrary numbers


mattsse:
fixing clippy rn

DaniPopes:
off-spec alert!

Rjected:
> @Rjected should we remove the totaldifficulty from the object?

We probably can, IIRC this is not required for spawning geth with a normal genesis, and we don't use ethers for hive tests / hive genesis

mattsse:
@Rjected turns out this can be fixed by manually converting the captured float, let's do that on reth as well

PR #2619, opened 2023-09-30
Recognize localdev.me as "local" domain

joroshiba:
## Motivation

When running an application in k8s locally it is common to use `localdev.me` domain for ingress, as documented in [ingress-nginx docs](https://kubernetes.github.io/ingress-nginx/deploy/#local-testing). This is a domain owned by Amazon which DNS points to `127.0.01`, can be considered stable. Ethers-rs, however, doesn't recognize this as a local chain. 

My primary motivation here is for polling in foundry to have local polling behavior running against a local k8s cluster.

## Solution

Update the`is_local_endpoint` rules to recognize `localdev.me` as local chain.

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation (I don't think this is needed)
-   [x] Breaking changes (none)

### DIFF
--- a/ethers-providers/src/rpc/provider.rs
+++ b/ethers-providers/src/rpc/provider.rs
@@ -1501,6 +1501,7 @@ impl ProviderExt for Provider<HttpProvider> {
 /// ```
 /// use ethers_providers::is_local_endpoint;
 /// assert!(is_local_endpoint("http://localhost:8545"));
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I think that's reasonable

DaniPopes (approved):

PR #2622, opened 2023-10-03
Re-enable the transaction.other field for optimism feature

mdehoog:
## Motivation

I'm in the process of adding op-stack deposit tx support to anvil. This requires enabling the `optimism` feature.

Anvil uses the `other` field to pass back additional information: https://github.com/foundry-rs/foundry/blob/49007938138ae26379e7a19bf3b2ec2ba6822017/crates/anvil/src/eth/api.rs#L1801-L1812

## Solution

Remove the `optimism` feature opt-out from the `other` field. This also makes it consistent with the `Block` struct: https://github.com/gakonst/ethers-rs/blob/70e5f022b1bd94c26dbc85d57b5342e5e23ba8c2/ethers-core/src/types/block.rs#L109-L112
and the `TransactionReceipt` struct:
https://github.com/gakonst/ethers-rs/blob/70e5f022b1bd94c26dbc85d57b5342e5e23ba8c2/ethers-core/src/types/transaction/response.rs#L531-L534

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mdehoog:
My understanding is that Anvil uses the `other` field to pass back additional information: https://github.com/foundry-rs/foundry/blob/49007938138ae26379e7a19bf3b2ec2ba6822017/crates/anvil/src/eth/api.rs#L1801-L1812

If we enable the `optimism` feature for the Anvil build, we lose this field.

mdehoog:
for example, in my anvil branch that adds op deposit tx support, i've had to disable usage of the other field: https://github.com/foundry-rs/foundry/compare/master...mdehoog:foundry:op-deposit#diff-a76a82dc700c0415fcd07ec10cde2ad56234e41af0ea3c07f8d1a8383b8b0913L1803-R1812

ethers-core/src/types/transaction/response.rs
```diff
@@ -132,7 +132,7 @@ pub struct Transaction {
     pub chain_id: Option<U256>,
 
     /// Captures unknown fields such as additional fields used by L2s
-    #[cfg(not(any(feature = "celo", feature = "optimism")))]
+    #[cfg(not(any(feature = "celo")))]
     #[serde(flatten)]
     pub other: crate::types::OtherFields,
```
mattsse, line 137: okay, I think this is reasonable to always enable this because this is also useful for other chains that include additional fields

mattsse (approved):

mattsse (approved):
as discussed, this is reasonable to capture anything else, regardless of the optimism feature

PR #2626, opened 2023-10-06
fix(ethers_solc): hardcoded import remapping fix

plotchy:
## Motivation

When using ethers_solc to compile contracts, remappings have an issue with the hardcoded specific fix for an openzeppelin contracts issue. 

I'm not very familiar with the issue described in the doc comments, but it is erroring in situations where it shouldn't be applicable. 

```rust
// we handle the edge case where the path of a remapping ends with "contracts"
// (`<name>/=.../contracts`) and the stripped import also starts with
// `contracts`
if let Ok(adjusted_import) = stripped_import.strip_prefix("contracts/") {
    if r.path.ends_with("contracts/") && !lib_path.exists() {
        return Path::new(&r.path).join(adjusted_import)
    }
}
```

### Problematic setup

Example import in sol file:
```
import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
```

Example remapping:
```
@openzeppelin/=lib/openzeppelin-contracts/
```

in this case, the path of the remapping ends in "contracts/" (lib/openzeppelin-contracts/) and the stripped import also starts with "contracts" (contracts/utils/math/Math.sol).

this makes the resolver look at:
lib/openzeppelin-contracts/utils/math/Math.sol
when it should instead look at:
lib/openzeppelin-contracts/contracts/utils/math/Math.sol

## Solution

Stripping based on `.ends_with("contracts/")` isn't precise. Better suited for issue regarding the comment is `.ends_with("/contracts/")`.

In this instance, the remapping correctly would not get flagged, as it is does not end with "/contracts/" (lib/openzeppelin-contracts/).

### DIFF
--- a/ethers-solc/src/config.rs
+++ b/ethers-solc/src/config.rs
@@ -320,8 +320,8 @@ impl ProjectPathsConfig {
     ///
     /// There is no strict rule behind this, but because [`crate::remappings::Remapping::find_many`]
     /// returns `'@openzeppelin/=node_modules/@openzeppelin/contracts/'` we should handle the
-    /// case if the remapping path ends with `contracts` and the import path starts with
-    /// `<remapping name>/contracts`. Otherwise we can end up with a resolved path that has a
+    /// case if the remapping path ends with `/contracts/` and the import path starts with
+    /// `<remapping name>/contracts/`. Otherwise we can end up with a resolved path that has a
     /// duplicate `contracts` segment:
     /// `@openzeppelin/contracts/contracts/token/ERC20/IERC20.sol` we check for this edge case
     /// here so that both styles work out of the box.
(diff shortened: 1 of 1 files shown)


ethers-solc/src/config.rs
```diff
@@ -348,7 +348,7 @@ impl ProjectPathsConfig {
                     // (`<name>/=.../contracts`) and the stripped import also starts with
                     // `contracts`
                     if let Ok(adjusted_import) = stripped_import.strip_prefix("contracts/") {
-                        if r.path.ends_with("contracts/") && !lib_path.exists() {
+                        if r.path.ends_with("/contracts/") && !lib_path.exists() {
```
mattsse, line 351: let's add a note here with the openzeppelin-contracts as example

I need as much context as possible for all these rules

mattsse (changes requested):
nice find!

smol doc nit

ethers-solc/src/config.rs
plotchy: Ok! Since it applies to more than oz, I wanted to keep it general.
we want to prevent matching on directories that end in things like `-contracts`, `_contracts`, `.contracts`

mattsse (approved):
thank you!

paging @DaniPopes

DaniPopes (approved):

Issue #1239, opened 2022-05-09
Generate exact constructor bindings


mattsse:
**Is your feature request related to a problem? Please describe.**
* currently abigen loosely implements a `deploy(args: impl Tokenziable)` function 
**Describe the solution you'd like**
Generate exact, typed arguments 
**Describe alternatives you've considered**

**Additional context**

PR #2631, opened 2023-10-10
FIX: Generate structs for constructor arguments

aakoshh:
## Motivation

I noticed that the `Abigen` did not generate a `struct` for the constructor parameters. This was a struct that did not appear anywhere else in the interface, so I thought it was because ethers only generates bindings for interacting with an already deployed contract, especially because there was an open issue seemingly asking for this feature: https://github.com/gakonst/ethers-rs/issues/1239 and also because the generated `deploy` method took a generic input parameter.

Upon closer inspection it looked like it _should_ generate structs for the constructor inputs, however this did not happen because the constructor (at least in our case) didn't have a `name`. This would normally be the name of the `function`, but since a `constructor` doesn't need any, it's empty.

## Solution

Use the name `"constructor"` if the `item.name` is empty. 

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (approved):
cool, tysm

PR #2632, opened 2023-10-10
conversion of a public key, regardless of compression, to an address

nanaknihal:
This commit exposes a method to convert public keys to addresses. It is tested against the public key examples on https://docs.ethers.org/v5/api/utils/address/#utils-computeAddress. It does not convert private keys to address because `Wallet` already does that.

## Motivation
Often, it is helpful to compute the address from the public key, especially in threshold signature schemes, but as far is I can tell there is no convenient function to do so. _ethers.js_ exposes a `computeAddress`, but I could not find an _ethers-rs_ function to do the same. Those unfamiliar with elliptic curves typically would rather avoid dealing with the question of how to hash a public key, as they would likely have to study point encoding to do it.

## Solution
This provides a simple function to convert a public key to an Ethereum address, regardless of compression.

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/utils/mod.rs
+++ b/ethers-core/src/utils/mod.rs
@@ -606,7 +606,7 @@ pub(crate) fn unused_port() -> u16 {
     local_addr.port()
 }
 
-
+/// Converts an public key, in compressed or uncompressed form to an Ethereum address
 pub fn pubkey_to_addr<T: AsRef<[u8]>>(pubkey: T) -> Result<[u8; 20], elliptic_curve::Error> {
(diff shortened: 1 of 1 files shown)


mattsse:
defer to @prestwich here

prestwich (changes requested):
method is good, thanks for adding a test :)

nits:
- move the function impl up around L390, it should be just before or just after `secret_key_to_address`
- refactor `secret_key_to_address` to use this method 
- rename to `public_key_to_address` to match the other method

:)

prestwich:
closing this as the commits are now included in main. Thank you for your contribution!

PR #2637, opened 2023-10-16
Fix ContractRevert::decode_with_selector

aakoshh:
## Motivation

I tried decoding an error returned by one of our contracts by calling `ContractRevert::decode_with_selector` using the `enum` that `Abigen` generates for the errors that the contract can revert with, but it kept turning into `None`. During debugging I noticed that while the selector is correctly accepted, it is not made available to the actual decode functions which try each possible error type, so they all fail trying to decode empty data.

## Solution

Unless the error data prefix is the one corresponding to the `String` type, pass the data to the generated `AbiDecode::decode` as-is, so that it can be inspected by the independent decoders of the types that make up the `enum`.

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


ethers-contract/src/error.rs
```diff
@@ -34,7 +34,15 @@ pub trait ContractRevert: AbiDecode + AbiEncode + Send + Sync {
...
-        <Self as AbiDecode>::decode(&data[4..]).ok()
+
+        if selector == String::selector() {
+            <Self as AbiDecode>::decode(&data[4..]).ok()
+        } else {
+            // Try with and without a prefix, just in case.
+            <Self as AbiDecode>::decode(data)
+                .or_else(|_| <Self as AbiDecode>::decode(&data[4..]))
+                .ok()
+        }
```
mattsse, line 45: I see, for the string case, we expect that `Self` is `String`,
if not we try without prefix and with

mattsse (approved):
lgtm

ethers-contract/src/error.rs
aakoshh: Yes, the `String` decoder doesn't expect the prefix. I thought everything else did, but some example tests failed without the fallback approach. I don't know why, the tests just timed out waiting for Anvil to start, twice - could be unrelated.

PR #2640, opened 2023-10-17
feat: improve gnosis chain configuration

luzzif:
## Motivation

I just wanted to update Gnosis Chain's config in the lib as it was pretty outdated (Gnosis Chain underwent the Shapella upgrade recently and with it comes `push0` support for example).

## Solution

- Rename xDai to Gnosis in chain enum
- Remove mention of xDai in serde/strum
- Add average blocktime for Gnosis and Chiado
- Mark Gnosis and Chiado as supporting EIP-1559 and `push0`
- Use Gnosisscan as Gnosis' block explorer instead of BlockScout

### DIFF
--- a/ethers-core/src/types/chain.rs
+++ b/ethers-core/src/types/chain.rs
@@ -92,9 +92,9 @@ pub enum Chain {
 
     Metis = 1088,
 
-    #[strum(to_string = "xdai", serialize = "gnosis", serialize = "gnosis-chain")]
-    #[serde(alias = "xdai", alias = "gnosis", alias = "gnosis_chain")]
-    XDai = 100,
+    #[strum(to_string = "gnosis", serialize = "gnosis-chain")]
+    #[serde(alias = "gnosis", alias = "gnosis_chain")]
+    Gnosis = 100,
 
     Polygon = 137,
     #[strum(to_string = "mumbai", serialize = "polygon-mumbai")]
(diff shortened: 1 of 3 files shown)


ethers-core/src/types/chain.rs
```diff
@@ -92,9 +92,9 @@ pub enum Chain {
 
     Metis = 1088,
 
-    #[strum(to_string = "xdai", serialize = "gnosis", serialize = "gnosis-chain")]
-    #[serde(alias = "xdai", alias = "gnosis", alias = "gnosis_chain")]
-    XDai = 100,
+    #[strum(to_string = "gnosis", serialize = "gnosis-chain")]
+    #[serde(alias = "gnosis", alias = "gnosis_chain")]
+    Gnosis = 100,
```
DaniPopes, line 97: Can you keep `"xdai"` as an alias on both strum and serde to avoid a breaking change?

DaniPopes (changes requested):

ethers-core/src/types/chain.rs
luzzif: Oh sorry, absolutely. Addressed in ab0071ba4f63d1db346fde773be0b14e25c32028

DaniPopes (approved):

mattsse (approved):

PR #2659, opened 2023-11-01
use https://api.scrollscan.com/ instead

0xCalibur:
https://scrollscan.com/api

results in

```
{"status":"0","message":"NOTOK","result":"Invalid API URL endpoint, use https://api.scrollscan.com"}
```

I presume this is why scroll verify is now failing


mattsse (approved):
thanks
will bump on foundry

DaniPopes (approved):

PR #2662, opened 2023-11-03
feat: Add fields to debug_trace_call

sakotn:
## Motivation

Some fields are missing, if `enable_return_data: Some(true)` at `.debug_trace_call` - return data is not deserialized

## Solution

Added this fields

## PR Checklist

-   [ ] Added Tests

### DIFF
--- a/ethers-core/src/types/trace/geth.rs
+++ b/ethers-core/src/types/trace/geth.rs
@@ -30,6 +30,7 @@ pub struct DefaultFrame {
 
 // https://github.com/ethereum/go-ethereum/blob/e91cdb49beb4b2a3872b5f2548bf2d6559e4f561/eth/tracers/logger/logger.go#L65
 #[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+[serde(rename_all = "camelCase")]
 pub struct StructLog {
     pub depth: u64,
     #[serde(default, skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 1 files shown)


ethers-core/src/types/trace/geth.rs
```diff
@@ -48,6 +48,10 @@ pub struct StructLog {
     pub stack: Option<Vec<U256>>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub storage: Option<BTreeMap<H256, H256>>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub mem_size: Option<i32>,
```
mattsse, line 52:
```suggestion
    pub mem_size: Option<u64>,
```

ethers-core/src/types/trace/geth.rs
```diff
@@ -48,6 +48,10 @@ pub struct StructLog {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub return_data: Option<String>,
```
mattsse, line 54: these need renames because struct is not marked with `camelCase`

mattsse (changes requested):
smol nits

ethers-core/src/types/trace/geth.rs
sakotn: I thought about it too, but I didn't understand in case of absence of camel_case the rest of the structure is serialized

mattsse (approved):

PR #2668, opened 2023-11-07
Apply WebSocketConfig to initial WS provider connection

svenski123:
WsClient::connect_with_config() et al. ignores the passed in WebSocketConfig object on the initial connection.

On subsequent reconnections the WebSocketConfig object (as saved by the RequestManager) is correctly applied to the backend.

This is used to among other things increase the maximum allowable frame size (i.e. >16MB) or message size.

### DIFF
--- a/ethers-solc/src/compile/mod.rs
+++ b/ethers-solc/src/compile/mod.rs
@@ -874,7 +874,7 @@ mod tests {
             (">=0.4.0 <0.5.0", "0.4.26"),
             // latest - this has to be updated every time a new version is released.
             // Requires the SVM version list to be updated as well.
-            (">=0.5.0", "0.8.22"),
+            (">=0.5.0", "0.8.23"),
         ] {
             let source = source(pragma);
             let res = Solc::detect_version(&source).unwrap();


svenski123:
Rebased/force pushed on top of 0543e1cc which addressed the failed clippy and fmt checks in other parts of the code unrelated to this PR.

svenski123:
I bumped the solc version from v0.8.22 to v0.8.23 in the test case in ethers-solc/src/compile/mod.rs.
This should take care of the failed tests.

ethers-providers/src/rpc/transports/ws/manager.rs
```diff
@@ -291,7 +310,7 @@ impl RequestManager {
         reconnects: usize,
     ) -> Result<(Self, WsClient), WsClientError> {
         let (backend, (instructions_tx, instructions_rx), channel_map) =
-            Self::connect_internal(conn.clone()).await?;
+            Self::connect_with_config_internal(conn.clone(), config.clone()).await?;
```
mattsse, line 313: ah I see, we didn't use the config for the connect

mattsse (approved):
lgtm, ty

PR #2671, opened 2023-11-08
chore: make `compiled_artifacts` public in `ProjectCompileOutput` struct

dutterbutter:
## Motivation

This pull request proposes a change to the `ProjectCompileOutput` struct by making the `compiled_artifacts` field public. The modification is required for the usage of extending `foundry` to work with non-solc compilers like [`zksolc`](https://github.com/matter-labs/era-compiler-solidity/tree/main) for zkSync.

**Context:**

Currently, the `compiled_artifacts` field is marked as `pub(crate)`, making it accessible only within the same crate. This encapsulation poses a limitation when the struct is used as part of a larger build process where access to `compiled_artifacts` is needed.

**Rationale:**

In effort to extend `foundry` support to zkSync ecosystem, direct access is needed to the `compiled_artifacts` field. You can see where we make use of `compiled_artifacts` [here](https://github.com/matter-labs/foundry-zksync/blob/bedd7221cd5af506c01080eb794e72391fdd97c7/crates/common/src/zk_compile.rs#L351).

- **Enhanced Flexibility:** Making `compiled_artifacts` public allows for more flexible post-compilation processes, including custom artifact handling, storage, and modifications tailored to specific use cases.

**Functionality Impact:**

The proposed change will have no impact on existing functionality. It merely extends the accessibility of the `compiled_artifacts` field, enabling external crates to utilize the compiled bytecode. This change is backward-compatible and does not alter the behavior of any existing code that depends on the `ProjectCompileOutput` struct.

## Solution

**Solution:**

```
pub compiled_artifacts: Artifacts<T::Artifact>,
```

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-solc/src/compile/output/mod.rs
+++ b/ethers-solc/src/compile/output/mod.rs
@@ -28,7 +28,7 @@ pub struct ProjectCompileOutput<T: ArtifactOutput = ConfigurableArtifacts> {
     /// contains the aggregated `CompilerOutput`
     pub(crate) compiler_output: AggregatedCompilerOutput,
     /// all artifact files from `output` that were freshly compiled and written
-    pub(crate) compiled_artifacts: Artifacts<T::Artifact>,
+    pub compiled_artifacts: Artifacts<T::Artifact>,
     /// All artifacts that were read from cache
     pub(crate) cached_artifacts: Artifacts<T::Artifact>,
     /// errors that should be omitted


mattsse (changes requested):
henlo @dutterbutter :)

we're about to deprecate this and moved this over to https://github.com/foundry-rs/compilers same code, just no longer dependent on ethers, instead it uses alloy deps now for json abi etc.

foundry is also using this already

so please open the PR there, and a getter function instead of `pub` :)

PR #2679, opened 2023-11-20
fix(core): normalize signature S

DaniPopes:
## Motivation

https://github.com/bluealloy/revm/pull/870

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/types/signature.rs
+++ b/ethers-core/src/types/signature.rs
@@ -127,8 +127,8 @@ impl Signature {
 
     /// Retrieves the recovery signature.
     fn as_signature(&self) -> Result<(RecoverableSignature, RecoveryId), SignatureError> {
-        let recovery_id = self.recovery_id()?;
-        let signature = {
+        let mut recovery_id = self.recovery_id()?;
+        let mut signature = {
             let mut r_bytes = [0u8; 32];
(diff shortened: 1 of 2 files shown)


ethers-core/src/types/signature.rs
```diff
@@ -324,6 +332,16 @@ mod tests {
         assert_eq!(tx.recover_from().unwrap(), expected);
     }
 
+    #[test]
+    fn can_recover_tx_sender_not_normalized() {
```
DaniPopes, line 336: This is from the Anvil test removed in https://github.com/foundry-rs/foundry/pull/6327 which I'll add back after this PR is merged

DaniPopes: https://github.com/foundry-rs/foundry/pull/6367

mattsse (approved):
lgtm,
this should close https://github.com/foundry-rs/foundry/issues/6072

DaniPopes:
> lgtm, this should close [foundry-rs/foundry#6072](https://github.com/foundry-rs/foundry/issues/6072)

Before this patch it gets rejected as per k256 0.13.2. With this patch it succeeds again.

PR #2398, opened 2023-05-03
feat: implement CCIP Read as eth_call middleware

yorhodes:
Implement [CCIP Read spec](https://eips.ethereum.org/EIPS/eip-3668) as optional provider middleware.

## Motivation

Fixes https://github.com/gakonst/ethers-rs/issues/2143

## Solution

See [ethers-js implementation](https://github.com/ethers-io/ethers.js/blob/f2fb8b5ede48f8e7a3af76f71cc01c22753d2ca5/src.ts/providers/abstract-provider.ts#L757)

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

PR #2684, opened 2023-11-22
Add skeleton of multicall middleware

yorhodes:
## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-middleware/src/multicall.rs
+++ b/ethers-middleware/src/multicall.rs
@@ -1,0 +1,124 @@
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use ethers_contract::{multicall::Multicall, BaseContract, ContractCall, FunctionCall};
+use ethers_core::{
+    abi::{Bytes, Tokenizable},
+    types::{transaction::eip2718::TypedTransaction, BlockId},
+};
+use ethers_providers::{Middleware, MiddlewareError};
+use instant::Duration;
+use thiserror::Error;
+
+use tokio::sync::mpsc;
+use tokio::sync::oneshot;
+
+type MulticallTx<M: Middleware> = (ContractCall<M, Bytes>, oneshot::Sender<Result<Bytes, MulticallError<M>>>);
+
+/// Middleware used for transparently leveraging multicall functionality
+pub struct MulticallMiddleware<M: Middleware> {
+    inner: Arc<M>,
+    contract: BaseContract,
+    multicall: Multicall<M>,
+    rx: mpsc::UnboundedReceiver<MulticallTx<M>>,
+    tx: mpsc::UnboundedSender<MulticallTx<M>>,
+    checkpoint: instant::Instant,
+    frequency: Duration
+}
+
+impl<M> MulticallMiddleware<M>
+where
+    M: Middleware,
+{
+    /// Instantiates the nonce manager with a 0 nonce. The `address` should be the
+    /// address which you'll be sending transactions from
+    pub async fn new(inner: M, contract: BaseContract, frequency: Duration) -> Result<Self, MulticallError<M>> {
+        // TODO: support custom multicall address
+        let multicall = Multicall::new(inner, None).await?;
+
+        let (tx, rx) = mpsc::unbounded_channel();
+
+        let timestamp = instant::now();
+
+        Ok(Self { inner: Arc::new(inner), multicall, contract, tx, rx, checkpoint: timestamp, frequency })
+    }
+
+    pub async fn run(&mut self) {
+        while let Some((call, callback)) = self.rx.recv().await {
+            self.multicall.add_call(call, false);
+
+            let timestamp = instant::now();
+            if timestamp.duration_since(self.checkpoint) > self.frequency {
+                self.checkpoint = timestamp;
+
+                let results = self.multicall.call_raw().await?;
+                self.multicall.clear_calls();
+
+                callback.send(results.pop());
+            }
+        }
+    }
+
+    fn call_from_tx<D: Tokenizable>(&self, tx: &TypedTransaction, block: Option<BlockId>) -> Option<ContractCall<M, D>> {
+        if let Some(data) = tx.data() {
+            if let Ok(function) = self.contract.get_fn_from_input(data) {
+                return Some(FunctionCall::new(*tx, *function, self.inner, block));
+            }
+        }
+
+        None
+    }
+}
+
+#[derive(Error, Debug)]
+/// Thrown when an error happens at the Multicall middleware
+pub enum MulticallError<M: Middleware> {
+    /// Thrown when the internal middleware errors
+    #[error("{0}")]
+    MiddlewareError(M::Error),
+}
+
+impl<M: Middleware> MiddlewareError for MulticallError<M> {
+    type Inner = M::Error;
+
+    fn from_err(src: M::Error) -> Self {
+        MulticallError::MiddlewareError(src)
+    }
+
+    fn as_inner(&self) -> Option<&Self::Inner> {
+        match self {
+            MulticallError::MiddlewareError(e) => Some(e),
+        }
+    }
+}
+
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+impl<M> Middleware for MulticallMiddleware<M>
+where
+    M: Middleware,
+{
+    type Error = MulticallError<M>;
+    type Provider = M::Provider;
+    type Inner = M;
+
+    fn inner(&self) -> &M {
+        &self.inner
+    }
+
+    async fn call(
+        &self,
+        tx: &TypedTransaction,
+        block: Option<BlockId>,
+    ) -> Result<Bytes, Self::Error> {
+        if let Some(call) = self.call_from_tx(tx, block) {
+            let (tx, rx) = oneshot::channel();
+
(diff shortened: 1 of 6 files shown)


yorhodes:
hopefully more in scope than my last middleware contribution https://github.com/gakonst/ethers-rs/pull/2398

ethers-middleware/src/multicall.rs
```diff
@@ -0,0 +1,386 @@
...
+        }
+    }
+}
+
+impl<M> MulticallMiddleware<M>
+where
+    M: Middleware,
+{
+    /// Instantiates the multicall middleware to recognize the given `match_abis`
+    pub fn new(
```
mattsse, line 150: need #Panics section

ethers-middleware/src/multicall.rs
```diff
@@ -0,0 +1,386 @@
...
+        inner: M,
+        match_abis: Vec<Abi>,
+        max_batch_size: usize,
+        multicall_address: Option<Address>,
+    ) -> (Self, MulticallProcessor<M>) {
+        if max_batch_size < 2 {
+            panic!("batches must be at least 2 calls to justify the overhead of multicall");
+        }
+
+        let (tx, rx) = mpsc::channel(max_batch_size);
```
mattsse, line 160: should this channel be unbounded?

ethers-providers/src/rpc/transports/mock.rs
```diff
@@ -5,14 +5,15 @@ use serde_json::Value;
...
+    ops::Deref,
     sync::{Arc, Mutex},
 };
 use thiserror::Error;
 
 /// Helper type that can be used to pass through the `params` value.
 /// This is necessary because the wrapper provider is supposed to skip the `params` if it's of
 /// size 0, see `crate::transports::common::Request`
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
```
mattsse, line 16:
```suggestion
#[derive(Debug, PartialEq, Eq)]
```

ethers-middleware/src/multicall.rs
```diff
@@ -0,0 +1,386 @@
...
+/// Middleware used for transparently leveraging multicall functionality
+pub struct MulticallMiddleware<M: Middleware> {
+    inner: Arc<M>,
+    contracts: Vec<BaseContract>,
+    tx: mpsc::Sender<MulticallRequest<M>>,
+}
+
+#[derive(Error, Debug)]
+/// Thrown when an error happens at the Multicall middleware
+pub enum MulticallMiddlewareError<M: Middleware> {
```
mattsse, line 50: move below impl block

ethers-middleware/src/multicall.rs
```diff
@@ -0,0 +1,386 @@
...
+    multicall_address: Option<Address>,
+    max_batch_size: usize,
+    rx: mpsc::Receiver<MulticallRequest<M>>,
+}
+
+#[derive(Debug, Clone)]
+/// Middleware used for transparently leveraging multicall functionality
+pub struct MulticallMiddleware<M: Middleware> {
+    inner: Arc<M>,
+    contracts: Vec<BaseContract>,
```
mattsse, line 44: this is immutable?
can Arc then?

ethers-middleware/src/multicall.rs
```diff
@@ -0,0 +1,386 @@
...
+use ethers_providers::{Middleware, MiddlewareError};
+use thiserror::Error;
+
+use tokio::sync::{mpsc, oneshot};
+
+type MulticallResult<M> = Result<Token, Arc<MulticallError<M>>>;
+type MulticallRequest<M> = (ContractCall<M, Token>, oneshot::Sender<MulticallResult<M>>);
+
+#[derive(Debug)]
+/// Processor for multicall middleware requests
```
mattsse, line 32: swap doc and derive order

ethers-middleware/src/multicall.rs
```diff
@@ -0,0 +1,386 @@
...
+}
+
+impl<M> MulticallProcessor<M>
+where
+    M: Middleware,
+{
+    /// Should be run in a separate task to process requests
+    pub async fn run(mut self) {
+        let mut multicall: Multicall<M> =
+            Multicall::new(self.inner, self.multicall_address).await.unwrap();
```
mattsse, line 85: this should return an error, so we don't panic (unwrap)

ethers-middleware/src/multicall.rs
```diff
@@ -0,0 +1,386 @@
...
+                    // For both errors (Disconnected and Empty), the correct action
+                    // is to process the items.  If the error was Disconnected, on
+                    // the next iteration rx.recv().await will be None and we'll
+                    // break from the outer loop anyway.
+                    Err(_) => break,
+                }
+            }
+
+            let (calls, callbacks): (Vec<_>, Vec<_>) = requests.into_iter().unzip();
+
```
mattsse, line 110: do we need any special handling if there's only one request?

mattsse (changes requested):
sorry for the late review.

I think this looks pretty neat and I can see how this is useful

a few small requests.


design wise, I'm not sure we need a new service for this if we instead put the `MultiCall` handling into a shared state via some async Mutex, we could probably make it work like that, but I guess a service task also works fine

I think one perf issue here is that executing multicalls is now sequential and no longer concurrent which could be an issue

ethers-middleware/src/multicall.rs
yorhodes: we don't *need* special handling but its a minor optimization?

ethers-middleware/src/multicall.rs
yorhodes: I was thinking that maybe when new contracts are instantiated we might append here?

ethers-providers/src/rpc/transports/mock.rs
yorhodes: I think I tried this

ethers-middleware/src/multicall.rs
yorhodes: I don't have a great intuition for this but I figured bounded was safer for memory growth?

yorhodes:
> design wise, I'm not sure we need a new service for this if we instead put the `MultiCall` handling into a shared state via some async Mutex, we could probably make it work like that, but I guess a service task also works fine

@mattsse I agree ideally users dont need to think about another service
I'm not actually sure how to achieve this because I don't have great knowledge of the available async primitives but if you give me some guidance I can try refactoring

> I think one perf issue here is that executing multicalls is now sequential and no longer concurrent which could be an issue

I'm not sure how we can avoid this and achieve the transparent functionality? I assume anyone who wants to use the middleware would like to eliminate the concurrency on the RPC

yorhodes:
for posterity, @mattsse and I discussed offline and decided to punt on some `parking_lot::Mutex` approach for a followup PR

ethers-middleware/src/multicall.rs
```diff
@@ -0,0 +1,392 @@
...
+use ethers_core::{
+    abi::{encode, Abi, AbiDecode, AbiEncode, Token, Tokenizable},
+    types::{
+        transaction::eip2718::TypedTransaction, Address, BlockId, Bytes, NameOrAddress,
+        TransactionRequest,
+    },
+};
+use ethers_providers::{Middleware, MiddlewareError};
+use thiserror::Error;
+
```
mattsse, line 25:
```suggestion
```

ethers-middleware/src/multicall.rs
```diff
@@ -0,0 +1,392 @@
+#![cfg(not(target_arch = "wasm32"))]
+
+use std::{ops::Deref, sync::Arc};
+
```
mattsse, line 4:
```suggestion
```

mattsse (approved):

BlinkyStitt:
This looks much more ergonomic. Very neat idea.

I don't like that you have to register all the abi ahead of time though. I don't always know the abis at start but I'd still like to have the calls batched still. 

And how does calling for blocks other than latest work? I see some possibly related comments about system transactions but need to read more of the code to tell what those are.

yorhodes:
> This looks much more ergonomic. Very neat idea.

Thanks! Seems related to your work on https://github.com/llamanodes/web3-proxy

> I don't like that you have to register all the abi ahead of time though. I don't always know the abis at start but I'd still like to have the calls batched still. Also, having to register and check the abis like this has a potential for hash collisions.

Knowing the ABIs is necessary to leverage multicall because you need to decode and slice the return data. Collisions are not handled but a more robust implementation could register ABIs for specific `to` addresses, rather than globally. This registration could be automated whenever a contract instance is connected to a provider with this middleware.

> And how does calling for blocks other than latest work? I see some possibly related comments about system transactions but need to read more of the code to tell what those are.

This isnt supported but you could implement this by having separate queues per block number

BlinkyStitt:
I did some experimenting today combining some of these ideas with some non-transparent multicall code that I wrote years ago. So far it only handles call and not the other helpers, but it doesn't need abis and it supports setting the block.

setting the block i think is rather crucial. Otherwise the transparent nature of the middleware could lead to very surprising results. 

I also want to automatically add state overrides for the contract if it's used on a block that predates the contract being deployed.

BlinkyStitt:
Here's a gist of what I did: https://gist.github.com/BlinkyStitt/837ffdbee2892c8857d192a5f4604b9c

PR #2685, opened 2023-11-23
fix: missing `AbiEncode` implem for tuples of length 16

makcandrov:
fixes:
- missing `AbiEncode` implem for tuples of length 16
- unused `num` variable in `impl_abi_codec_tuple` & `impl_abi_type_tuple` macros


mattsse (approved):
reasonable

PR #2693, opened 2023-12-01
fix: source code downloading and update api url on metis network

brotherlymite:
## Motivation

Currently when we run `cast etherscan-source --chain 1088 0x6c3bD2D5885D9e8a9B48c403EEb7A25195b7CCc1` to download source code on metis network, it throws an error `data did not match any variant of untagged enum ResponseData` which is because the response from the api `https://andromeda-explorer.metis.io/api` is not 100% compatible with the etherscan, so when it tries to deserialize it [here](https://github.com/gakonst/ethers-rs/blob/master/ethers-etherscan/src/lib.rs#L206) during the sanity check the error is thrown.

## Solution

Metis has a new updated explorer and api which is more compatible with etherscan. The api response for downloading the source code on the new api is in-line with the etherscan one and should fix the issue. 
Also as per the Metis team, the old api url and explorer is soon gonna be deprecated so I think it is a good idea to update it.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/types/chain.rs
+++ b/ethers-core/src/types/chain.rs
@@ -516,7 +516,7 @@ impl Chain {
             }
 
             Metis => {
-                ("https://andromeda-explorer.metis.io/api", "https://andromeda-explorer.metis.io/")
+                ("https://api.routescan.io/v2/network/mainnet/evm/1088/etherscan/api", "https://explorer.metis.io/")
             }
 
             Chiado => {


ethers-core/src/types/chain.rs
```diff
@@ -516,7 +516,7 @@ impl Chain {
             }
 
             Metis => {
-                ("https://andromeda-explorer.metis.io/api", "https://andromeda-explorer.metis.io/")
+                ("https://api.routescan.io/v2/network/mainnet/evm/1088/etherscan/api", "https://explorer.metis.io/")
```
mattsse, line 519: this url looks oddly long,
is this documented somewhere?

mattsse:
do you have a sample response that fails deserialization?

brotherlymite:
> do you have a sample response that fails deserialization?

You can query this [api](https://andromeda-explorer.metis.io/api?module=contract&action=getsourcecode&address=0x6c3bD2D5885D9e8a9B48c403EEb7A25195b7CCc1&apikey=4NU4PT84EWZ5FJCT6DHY4UWDEXGJ5CZY8K) to get the response

PR #2694, opened 2023-12-05
Avoid panic on transaction deserialization

emostov:
## Motivation

I encountered an edge case where `TypedTransaction::decode` panick'ed from a malformed transaction.

## Solution

Instead of using `rlp_data.into::<U64>()` (which panics on failure), use `U64::decode` method, which will error on failure.

## PR Checklist

-   [ ] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


ethers-core/src/types/transaction/eip2718.rs
```diff
@@ -417,7 +417,7 @@ impl TypedTransaction {
 impl Decodable for TypedTransaction {
     fn decode(rlp: &rlp::Rlp) -> Result<Self, rlp::DecoderError> {
         let tx_type: Option<U64> = match rlp.is_data() {
-            true => Some(rlp.data()?.into()),
+            true => Some(U64::decode(rlp)?),
```
mattsse, line 420: hmm, unclear what this even did before.

this entire function looks a bit weird.

this should not rlp decode here I believe, because this is the type id of the envelope?

unclear what `rlp.data()?.into()` even did, but I think this here should just use the single byte.

idk what `is_data` is

ethers-core/src/types/transaction/eip2718.rs
```diff
@@ -909,4 +909,12 @@ mod tests {
             assert_eq!(tx0, tx1);
         }
     }
+
+    #[test]
+    fn malformed_transaction_does_not_panic() {
+        let tx = hex::decode("b9011e02f9011a0180850ba43b7400850df8475864830493e0947a250d5630b4cf539739df2c5dacb4c659f2488d88016345785d8a0000b8e4b6f9de95000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000000000005bd929b84c3dac16515c6b6c95f70fdae9c05c0c00000000000000000000000000000000000000000000000000000000650267140000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2000000000000000000000000dac17f958d2ee523a2206206994597c13d831ec7c0808080").unwrap();
```
mattsse, line 915: where's this coming from?

mattsse (changes requested):

ethers-core/src/types/transaction/eip2718.rs
emostov: This was given to me - apparently it was spit out from go-ethereum. But I also have encountered the same issue while fuzzing inputs to this function. Regardless of the exact payload, I don't think think we want it to panic. If you run the test on the [commit prior ]([9ad0740](https://github.com/gakonst/ethers-rs/pull/2694/commits/9ad074096f83c1f70251c1a1eadf8c3c281f1150))to the fix being introduced, it will panic.

ethers-core/src/types/transaction/eip2718.rs
DaniPopes: `rlp.data()` returns `Result<&[u8]>`, and apparently Parity Uint types [implement `From<&[u8]>`](https://docs.rs/uint/latest/src/uint/uint.rs.html#1363-1368) [by panicking](https://docs.rs/uint/latest/src/uint/uint.rs.html#1239-1253) ^^

emostov: I refactored a bit to explicitly check the first byte as a the tx type - see [e7f8369](https://github.com/gakonst/ethers-rs/pull/2694/commits/e7f836907980e9de4cf4a4d83cecdd34dd381258).

It is a bit strange that the initial code was trying to coerce into a `U64`, when EIP-2718 specifies a single u8 as the tx type

emostov: I also noticed [`decode_signed` has similar logic ](https://github.com/gakonst/ethers-rs/blob/f0e5b194f09c533feb10d1a686ddb9e5946ec107/ethers-core/src/types/transaction/eip2718.rs#L381-L413) and saw that you made a similar refactor to it in [pr 1733](https://github.com/gakonst/ethers-rs/pull/1733/files)

emostov:
Pushed a fix that should get CI to pass (or at least get farther along)

emostov:
Looks like CI is good other then just a flaky example

PR #2697, opened 2023-12-07
FIX: Accept 'input' or 'data', (but not both), in transaction request

aakoshh:
## Motivation

https://github.com/ethereum/go-ethereum/pull/28078 changed the `go-ethereum` client to send the transaction payload in an `input` field instead of `data`, which is how it's [defined](https://github.com/ethereum/execution-apis/blob/431cf72fd3403d946ca3e3afc36b973fc87e0e89/src/schemas/transaction.yaml#L33-L35) in the API. Apparently most server implementations accept both fields, preferring `input` where both are present at the same time.

The `Eip1559TransactionRequest` has a `data` field, and thus it could not be used to deserialize such requests. 

## Solution

The PR adds a special deserializer to the `Eip1559TransactionRequest::data` field so that it looks for `input` first, `data` second.

## PR Checklist

-   [x] Added Tests
-   [x] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-core/src/types/transaction/eip1559.rs
+++ b/ethers-core/src/types/transaction/eip1559.rs
@@ -8,7 +8,7 @@ use thiserror::Error;
 /// EIP-1559 transactions have 9 fields
 const NUM_TX_FIELDS: usize = 9;
 
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::{Deserialize, Serialize};
 
 /// An error involving an EIP1559 transaction request.
 #[derive(Debug, Error)]
@@ -42,12 +42,7 @@ pub struct Eip1559TransactionRequest {
 
     /// The compiled code of a contract OR the first 4 bytes of the hash of the
     /// invoked method signature and encoded parameters. For details see Ethereum Contract ABI
-    #[serde(
-        skip_serializing_if = "Option::is_none",
-        deserialize_with = "deserialize_input_or_data",
-        serialize_with = "serialize_input_or_data",
-        flatten
-    )]
+    #[serde(skip_serializing_if = "Option::is_none", alias = "input")]
     pub data: Option<Bytes>,
 
     /// Transaction nonce (None for next available nonce)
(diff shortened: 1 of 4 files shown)


ethers-core/src/types/transaction/eip1559.rs
```diff
@@ -42,7 +42,12 @@ pub struct Eip1559TransactionRequest {
 
     /// The compiled code of a contract OR the first 4 bytes of the hash of the
     /// invoked method signature and encoded parameters. For details see Ethereum Contract ABI
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        deserialize_with = "deserialize_input_or_data",
+        serialize_with = "serialize_input_or_data",
+        flatten
```
mattsse, line 49: I think this is just `alias = "input"`?

mattsse (changes requested):

ethers-core/src/types/transaction/eip1559.rs
aakoshh: That's what I tried to do as well, however it fails on the example where both fields are present:
```
---- types::transaction::eip1559::tests::test_tx_with_input_or_data stdout ----
thread 'types::transaction::eip1559::tests::test_tx_with_input_or_data' panicked at 'failed to parse request { "gas": "0x186a0",
                    "maxFeePerGas": "0x77359400",
                    "maxPriorityFeePerGas": "0x77359400",
                    "nonce": "0x2",
                    "to": "0x96216849c49358B10257cb55b28eA603c874b05E",
                    "value": "0x5af3107a4000",
                    "chainId": "0x539",
                    "accessList": [],
                     "data":"0x02", "input":"0x01" 
                }: duplicate field `data` at line 9 column 43', ethers-core/src/types/transaction/eip1559.rs:340:37
```

I'm not sure how important that is, whether there are actual clients that are so defensive that they send both fields, not knowing which one the server might expect; I only got it as a hint from someone who debugged this problem that certain implementation have a _preference_ for `input`, which presumes that there is a choice to made.

mattsse: sigh

ethers-core/src/types/transaction/eip1559.rs
```diff
@@ -283,3 +288,63 @@ impl From<&Transaction> for Eip1559TransactionRequest {
...
+}
+
+fn serialize_input_or_data<S: Serializer>(data: &Option<Bytes>, s: S) -> Result<S::Ok, S::Error> {
+    #[derive(Serialize)]
+    struct InputOrData<'a> {
+        data: &'a Option<Bytes>,
+    }
+
+    InputOrData { data }.serialize(s)
+}
```
mattsse, line 314: this we don't need, right?

mattsse: or is this to make flatten work?

ethers-core/src/types/transaction/eip1559.rs
aakoshh: Just saw in https://github.com/web3/web3.js/issues/1340 that Parity rejects the request if both `input` and `data` is present, so maybe it's okay to require one or the other with a simple `alias`.

ethers-core/src/types/transaction/eip1559.rs
aakoshh: Yeah, I thought so too, but unfortunately the `flatten` trick which makes both `input` and `data` available for `deserialize_input_or_data` means that during serialization it also wants to flatten the results, which it can't because it's a `String`. This returns a `{"data": "0x..."}` object which can be flattened :disappointed:

ethers-core/src/types/transaction/eip1559.rs
aakoshh: Duplication used to be an issue with web3 but got fixed apparently: https://github.com/web3/web3.js/issues/6301

aakoshh: Switched to `alias`, and added it to the legacy transaction type as well, just in case.

ethers-core/src/types/transaction/eip1559.rs
aakoshh: Removed it to keep it simple.

PR #2699, opened 2023-12-16
fix: StorageProof key should be U256

blckngm:
## Motivation

According to API sepcs, storage proof keys may not be full 32-byte long:

https://github.com/ethereum/execution-apis/blob/a0d03086564ab1838b462befbc083f873dcf0c0f/src/schemas/state.yaml#L50C29-L50C29

https://github.com/ethereum/execution-apis/blob/a0d03086564ab1838b462befbc083f873dcf0c0f/tests/eth_getProof/get-account-proof-with-storage.io#L2

## Solution

Change the key to `U256`.

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes
  This might be a breaking change. Fixing this without a breaking change will be cumbersome, by e.g. manually implementing `Deserialize`.


mattsse:
@Rjected is this fixed in geth?

blckngm:
We should still be compatible with geth since `U256` can deserialize keys with or without leading zeros.

Related issue:

https://github.com/ethereum/go-ethereum/issues/27306

Rjected:
> We should still be compatible with geth since U256 can deserialize keys with or without leading zeros.

yes, so deserialization should still work. serializing as a U256 should also work

DaniPopes (approved):

PR #2723, opened 2024-02-01
Add polygon amoy to chain definitions

0xfourzerofour:
https://polygon.technology/blog/introducing-the-amoy-testnet-for-polygon-pos

Polygon recently gave more information about the polygon amoy chain that will be the new standard testnet while Mumbai is to be deprecated as it will stop working once goerli is deprecated.

- Add the basic chain information to the existing Chain enum


mattsse (commented):
thanks, could you please move this over to https://github.com/alloy-rs/chains

DaniPopes (approved):
Thanks!

PR #2724, opened 2024-02-01
Update strum

richardpringle:
## Motivation

Simply (and hopefully) fixing the failures from the @dependabot PR -> #2718

## Solution

Use the new name

## PR Checklist

It's unfortunate that this is a breaking change, but I guess any dependency update is a breaking change...

-   [ ] Added Tests
-   [ ] Added Documentation
-   [x] Breaking changes


DaniPopes (approved):
Thanks! It's not a breaking change, just a deprecation of the old name. The test failures were unrelated.

richardpringle:
> Thanks! It's not a breaking change, just a deprecation of the old name. The test failures were unrelated.

Isn't updating `strum` always a breaking change? In any case, thanks for accepting the PR!

PR #2736, opened 2024-02-12
chore: clippy

DaniPopes:
Unused stuff.


ethers-providers/src/rpc/pubsub.rs
```diff
@@ -14,18 +11,6 @@ use std::{
 };
 use tracing::error;
 
-/// A transport implementation supporting pub sub subscriptions.
-pub trait PubsubClient: JsonRpcClient {
```
DaniPopes, line 18: This was mistakenly duplicated and in a glob `::*` re-export so it was never caught.

mattsse (approved):

PR #2746, opened 2024-02-21
feat: Add support for cancun evm version

rimrakhimov:
This PR adds a new "Cancun" evm version into the enum of supported evm versions.

## Motivation

Starting from v0.8.24 a new "Cancun" network upgrade has been added into solidity compilers. The corresponding `ethers-solc::EvmVersion` has not been updated though, which results in inability for clients to use "Cancun" evm version when compiling the contracts using default `ethers-solc::CompilerInput`.

"Shanghai" remains a default evm version as v0.8.24 release does not yet make "Cancun" its default target

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes

### DIFF
--- a/ethers-solc/src/artifacts/mod.rs
+++ b/ethers-solc/src/artifacts/mod.rs
@@ -782,6 +782,7 @@ pub enum EvmVersion {
     Paris,
     #[default]
     Shanghai,
+    Cancun,
 }
 
 impl EvmVersion {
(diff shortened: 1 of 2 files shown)


rimrakhimov:
I've found the "Cancun" implementation in `foundry-compilers` crate and moved to it from `ethers-solc`. Will leave the PR open in case if you will still find it useful, but is not required for me anymore

Issue #2753, opened 2024-03-03
Failed to deserialize the empty key of storage proof in `EIP1186ProofResponse` (for Mainnet)


silathdiir:
**Version**
ethers v2.0.11

**Platform**
MacOS

**Description**

Download the `EIP1186ProofResponse` data to deserialize:
```
curl https://eth.llamarpc.com \
  -X POST \
  -H "Content-Type: application/json" \
V  --data '{"method":"eth_getProof","params":["0x105dD0eF26b92a3698FD5AaaF688577B9Cafd970",["0"],"latest"],"id":1,"jsonrpc":"2.0"}' > tmp_proof_empty_key.json

# Unnecessary to format the JSON.
cat tmp_proof_empty_key.json | jq > proof_empty_key.json
cp proof_empty_key.json ethers-core/testdata/proof_empty_key.json
```

Edit `ethers-core/testdata/proof_empty_key.json` to delete JSON RPC header:
```
{
-   "jsonrpc": "2.0",
-   "id": 1,
-   "result": {
  "accountProof": [
...
-  }
}
```

Add this test cases to `ethers-core/src/types/proof.rs`:
```
#[test]
fn can_deserialize_proof_empty_key() {
    serde_json::from_str::<EIP1186ProofResponse>(include_str!(
        "../../testdata/proof_empty_key.json"
    ))
    .unwrap();
}
```

And run the test:
```
cargo test -- can_deserialize_proof_empty_key
```

PR #2754, opened 2024-03-03
fix: make the empty key of storage proof in `EIP1186ProofResponse` could be deserialized

silathdiir:
Close https://github.com/gakonst/ethers-rs/issues/2753

## Motivation

Make the empty key (as `0x`) in storage proof could be deserialized.

## Solution

Add annotation `deserialize_stringified_numeric` to the field `key` of `StorageProof`.

## PR Checklist

-   [x] Added Tests
-   [ ] Added Documentation
-   [ ] Breaking changes


mattsse (approved):
makes sense, although empty key is weird.

I guess this would also be the case here

https://github.com/alloy-rs/alloy/blob/main/crates/rpc-types/src/eth/account.rs

PR #2810, opened 2024-09-06
chore: mark mantle and mantle sepolia chains as supporting eip 1559

luzzif:
## Motivation

Mantle and Mantle Sepolia support EIP 1559. As proof, we can see that by doing:

```
// for mantle
curl -X POST --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["latest", false],"id":1}' https://rpc.mantle.xyz

// for mantle sepolia
curl -X POST --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["latest", false],"id":1}' https://rpc.sepolia.mantle.xyz
```

the returned block object contains the `baseFeePerGas` attribute.

## Solution

I just updated the `is_legacy` function by moving `Mantle` and `MantleSepolia` to the `false` arm.

P.S.: I have a Rust program posting transactions to Mantle Sepolia and all of them are currently failing at the `fill_transaction` stage. Is there any workaround I can use while this is being reviewed and released?

### DIFF
--- a/ethers-core/src/types/chain.rs
+++ b/ethers-core/src/types/chain.rs
@@ -378,8 +378,6 @@ impl Chain {
             Boba |
             ZkSync |
             ZkSyncTestnet |
-            Mantle |
-            MantleTestnet |
             PolygonZkEvm |
             PolygonZkEvmTestnet |
             Metis |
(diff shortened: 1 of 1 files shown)


mattsse:
can you please apply the changes to https://github.com/alloy-rs/chains instead

luzzif:
I can do that too, but in my project I'm still using Ethers. Has Ethers been completely deprecated in favor of alloy?

